PR #9, opened 2020-11-18
Byte buffer

kneekey23:
*Description of changes:* Add ByteBuffer extensions with streaming transforms.


Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -30,6 +33,11 @@ public class ByteBuffer {
         currentEndianness = endianness
         return self
     }
+    
+    public func putBytes(_ value: UInt8) {
```
justinboswell, line 37: Should this be a byte array? put(UInt8) seems to have this functionality covered already?

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -259,3 +267,52 @@ extension ByteBuffer: AwsStream {
         return !self.status.is_end_of_stream
     }
 }
+
+extension ByteBuffer {
+    public convenience init(data: Data) {
+        self.init(size: data.count)
+        for byte in data {
```
justinboswell, line 274: Is there not an API to byte swap the whole buffer? There are HUGE accelerations for that if you can do it across whole streams vs byte by byte.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
kneekey23: I just needed another function that put a byte that didnt return self. want me to call it just `putByte` maybe?

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
kneekey23: see new code, maybe it's faster?

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
justinboswell: yeah, let's do `putByte`, `putBytes` is a LIE

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -271,15 +270,14 @@ extension ByteBuffer: AwsStream {
 extension ByteBuffer {
     public convenience init(data: Data) {
         self.init(size: data.count)
-        for byte in data {
-            array.append(byte.byteSwapped)
-        }
+        let byteArray: [UInt8] = data.map { $0 }
```
justinboswell, line 273: Is there a byte swap still happening here magically? should this be `$0.byteSwapped`?

justinboswell (approved):
Question-n-ship

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
kneekey23: idk, I stole this code from someone online. let me test. idk how this magic is occurring really, it was purely theft on my part.

kneekey23: turns out you didnt need the byte swapped to get the right type, I just thought you did. the magic works and should be the default.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -38,32 +37,16 @@ public class ByteBuffer {
...
     }
 
     public func put(_ value: UInt8) -> ByteBuffer {
         array.append(value)
-        pointer.advanced(by: currentIndex).initialize(to: value)
         return self
     }
 
-    public func put(_ value: Data) -> ByteBuffer {
+    public func put(_ value: Data) {
```
justinboswell, line 47: So just out of curiosity, why does only this overload not return self?

justinboswell (approved):

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
kneekey23: idk why I made any of them return self to be honest. I now want to just undo that on all of them. I can't remember if there was a good reason to do so. but this works so I figure ill just confirm this doesnt have any issues moving forward and I can do a separate PR to fix all the others.

PR #13, opened 2020-12-02
Adding Future Extension to track completion of a Future array

kneekey23:
*Description of changes:* Added single extension function on a group of futures

### DIFF
--- a/Source/AwsCommonRuntimeKit/crt/Future.swift
+++ b/Source/AwsCommonRuntimeKit/crt/Future.swift
@@ -69,7 +69,7 @@ public final class Future<Value> {
     }
 }
 
-extension Future {
+public extension Future {
     func chained<T>(closure: @escaping (FutureResult) -> Future<T>) -> Future<T> {
         // We'll start by constructing a "wrapper" promise that will be
         // returned from this method:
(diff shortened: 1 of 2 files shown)


Source/AwsCommonRuntimeKit/crt/Future.swift
```diff
@@ -91,4 +91,36 @@ extension Future {
...
+                }
+            case .failure(let error):
+                future.fail(error)
+            }
+        }
+
+        for(index, future) in futures.enumerated() {
+            if let result = future._value {
+                processResult(index, result)
+
```
justinboswell, line 121: weird newline

Source/AwsCommonRuntimeKit/crt/Future.swift
```diff
@@ -91,4 +91,36 @@ extension Future {
 
         return promise
     }
+
+    static func whenAllComplete(_ futures: [Future<Value>]) -> Future<Void> {
```
justinboswell, line 95: Should this be `Future<Value>` instead of `Future<Void>`?

justinboswell: Seems like you'd want the value if the bundled futures produce one.

kneekey23: ya I thought about returning each value in a completion as they come in, but I wrote this for a specific use case. I'll make it more generic :) good point

justinboswell (approved):
Based on discussion in slack, going to ignore the value return for now.

Source/AwsCommonRuntimeKit/crt/Future.swift
kneekey23: leaving off for now and add documentation. will return to this if required

PR #14, opened 2021-01-04
Signing

kneekey23:
*Description of changes:* Changes needed in order to use signer in client runtime

### DIFF
--- a/Source/AwsCommonRuntimeKit/io/Stream.swift
+++ b/Source/AwsCommonRuntimeKit/io/Stream.swift
@@ -11,12 +11,15 @@ private var vtable = aws_input_stream_vtable(seek: doSeek,
                                              destroy: doDestroy)
 //swiftlint:disable trailing_whitespace
 public class AwsInputStream {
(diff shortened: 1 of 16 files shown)


Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -11,12 +11,15 @@ private var vtable = aws_input_stream_vtable(seek: doSeek,
                                              destroy: doDestroy)
 //swiftlint:disable trailing_whitespace
 public class AwsInputStream {
+    //to be used in conversion from http request to sdk request and back during signing
+    public var byteBuffer: ByteBuffer
```
kneekey23, line 15: not sure if this is the right more here. would love to discuss as I have no issue converting an `SdkHttpRequest` to an `HttpRequest` for CRT to send it but trying to convert it back from `HttpRequest` to `SdkHttpRequest` is difficult.

justinboswell (approved):

PR #27, opened 2021-06-28
chore: remove unsafe cflags

wooj2:
This is required in order for `smithy-swift` to pull in this package as tagged dependency, because swift package manager will not allow the use of `unsafeFlags` in a tagged dependency.


kneekey23 (commented):
Are we able to either comment them out or put them in when there's an env variable so when we debug we aren't debugging blind?

wooj2 (commented):
Closing in favor of:
https://github.com/awslabs/aws-crt-swift/pull/28

PR #30, opened 2021-07-13
fix: add codable conformance to byte buffer plus extension

kneekey23:
*Description of changes:* This PR adds needed extensions to use `ByteBuffer` in streaming in the sdk

### DIFF
--- a/Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
+++ b/Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
@@ -177,6 +177,10 @@ public class ByteBuffer {
         return currentEndianness == .little ? Double(bitPattern: result.littleEndian)
             : Double(bitPattern: result.bigEndian)
     }
+    
+    public func toByteArray() -> [UInt8] {
+        return array
+    }
 
     public enum Endianness {
         case little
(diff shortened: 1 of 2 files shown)


Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -178,6 +191,10 @@ public class ByteBuffer {
             : Double(bitPattern: result.bigEndian)
     }
 
+    public func toByteArray() -> [UInt8] {
+        return array
+    }
+
```
wooj2, line 197: per convo: if we don't need this, lets remove it

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -49,6 +49,19 @@ public class ByteBuffer {
         return self
     }
 
+    public func put(_ value: ByteBuffer) {
+        array.append(contentsOf: value.array)
+    }
+
+    public func put(_ value: ByteBuffer, offset: UInt = 0, maxBytes: UInt?) {
```
wooj2, line 56: * if we default offset to 0, then it should come after variables that do not have defaults
* investigate combining both put functions to single function.

justinboswell: +1, these are the same function if `offset` and `maxBytes` have defaults

justinboswell (approved):

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
kneekey23: merging into one function

justinboswell (approved):

PR #34, opened 2021-07-30
fix: move s2n stuff in same package manifest

kneekey23:
*Description of changes:* This PR moves the S2N and LibCrypto modules into the same Package.swift manifest as the rest of the modules for proper tagging and release. Additionally it removes all build warnings.


Package.swift
```diff
@@ -1,36 +1,63 @@
...
 )
 
-let excludesFromAll = ["bin", "tests", "cmake", "codebuild", "CONTRIBUTING.md",
-                       "LICENSE", "format-check.sh", "NOTICE", "builder.json",
-                       "sanitizer-blacklist.txt", "CMakeLists.txt", "README.md",
-                       "CODE_OF_CONDUCT.md", "build-deps.sh"]
+var calDependencies: [Target.Dependency] = ["AwsCCommon"]
+var ioDependencies: [Target.Dependency] = ["AwsCCommon", "AwsCCal"]
+
+#if os(Linux)
```
JonathanHenson, line 20: is freebsd or netbsd supported for spm yet?

JonathanHenson (approved):

Package.swift
kneekey23: not that I am aware of, os options here: https://docs.swift.org/swift-book/ReferenceManual/Statements.html#:~:text=Valid%20arguments-,os(),-macOS%2C%20iOS%2C%20watchOS

PR #38, opened 2021-08-11
fix: refactor pointer creation to protect against improper init

kneekey23:
*Description of changes:* This PR refactors the way pointers are created and initialized in Swift to protect against cases where the pointer is created but uninitialized later causing a seg fault when the c code calls back into the Swift code.

### DIFF
--- a/Source/AwsCommonRuntimeKit/Utilities.swift
+++ b/Source/AwsCommonRuntimeKit/Utilities.swift
@@ -58,11 +58,18 @@ extension Bool {
     }
 }
 
-func fromPointer<T, P: PointerConformance>(ptr: T?) -> P? {
-   if let ptr = ptr {
-        let pointer = UnsafeMutablePointer<T>.allocate(capacity: 1)
-        pointer.initialize(to: ptr)
-       return P(OpaquePointer(pointer))
+// Ensure that any UnsafeXXXPointer is ALWAYS initialized to either nil or a value in a single call. Prevents the
+// case where you create an UnsafeMutableWhatever and do not call `initialize()` on it, resulting in a non-null but
+// also invalid pointer
+func fromPointer<T, P: PointerConformance>(ptr: T) -> P {
+    let pointer = UnsafeMutablePointer<T>.allocate(capacity: 1)
+    pointer.initialize(to: ptr)
+    return P(OpaquePointer(pointer))
+}
+
+func fromOptionalPointer<T, P: PointerConformance>(ptr: T?) -> P? {
+    if let ptr = ptr {
+        return fromPointer(ptr: ptr)
     }
     return nil
 }
--- a/Source/AwsCommonRuntimeKit/Utilities.swift
+++ b/Source/AwsCommonRuntimeKit/Utilities.swift
@@ -74,6 +74,21 @@ func fromOptionalPointer<T, P: PointerConformance>(ptr: T?) -> P? {
     return nil
 }
 
+func allocate<T>(_ capacity: Int = 1) -> UnsafeMutablePointer<T> {
+    let ptr = UnsafeMutablePointer<T>.allocate(capacity: capacity)
+    zeroStruct(ptr)
+    return ptr
+}
+
+func toPointerArray<T, P: PointerConformance>(_ array: [T]) -> P {
+    let pointers = UnsafeMutablePointer<T>.allocate(capacity: array.count)
+
+    for index in 0...array.count {
+        pointers.advanced(by: index).initialize(to: array[index])
+    }
+    return P(OpaquePointer(pointers))
+}
+
 protocol PointerConformance {
     init(_ pointer: OpaquePointer)
 }
--- a/Source/AwsCommonRuntimeKit/Utilities.swift
+++ b/Source/AwsCommonRuntimeKit/Utilities.swift
@@ -58,24 +58,43 @@ extension Bool {
     }
 }
 
-func getMutablePointer<T>(memoryType: T?) -> UnsafeMutablePointer<T>? {
-    if let memoryType = memoryType {
+func fromPointer<T, P: PointerConformance>(ptr: T?) -> P? {
+   if let ptr = ptr {
         let pointer = UnsafeMutablePointer<T>.allocate(capacity: 1)
-        pointer.initialize(to: memoryType)
-        return pointer
+        pointer.initialize(to: ptr)
+       return P(pointer: OpaquePointer(pointer))
     }
-
     return nil
 }
 
-func getMutableRawPointer<T>(memoryType: T?) -> UnsafeMutableRawPointer? {
-    return UnsafeMutableRawPointer(getMutablePointer(memoryType: memoryType))
+protocol PointerConformance {
+
+    init(pointer: OpaquePointer)
 }
 
-func getUnsafePointer<T>(memoryType: T?) -> UnsafePointer<T>? {
-    return UnsafePointer(getMutablePointer(memoryType: memoryType))
+extension UnsafeMutablePointer: PointerConformance {
+    init(pointer: OpaquePointer) {
+        self = UnsafeMutablePointer(pointer)
+    }
 }
 
-func getUnsafeRawPointer<T>(memoryType: T?) -> UnsafeRawPointer? {
-    return UnsafeRawPointer(getMutableRawPointer(memoryType: memoryType))
+extension UnsafeMutableRawPointer: PointerConformance {
+    init(pointer: OpaquePointer) {
+        self = UnsafeMutableRawPointer(pointer)
+    }
 }
+
+extension UnsafePointer: PointerConformance {
+    init(pointer: OpaquePointer) {
+        self = UnsafePointer(pointer)
+    }
+}
+
+extension UnsafeRawPointer: PointerConformance {
+    init(pointer: OpaquePointer) {
+        self = UnsafeRawPointer(pointer)
+    }
+}
+
+
+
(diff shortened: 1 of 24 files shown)


Source/AwsCommonRuntimeKit/Utilities.swift
```diff
@@ -57,3 +57,25 @@ extension Bool {
         return self ? 1 : 0
     }
 }
+
+func getMutablePointer<T>(memoryType: T?) -> UnsafeMutablePointer<T>? {
```
justinboswell, line 61: Is it possible to write this as:
```swift
func fromPointer<T, P>(ptr: T?) -> P<T>? {
   if let ptr = ptr {
        let pointer = UnsafeMutablePointer<T>.allocate(capacity: 1)
        pointer.initialize(to: ptr)
        return P<T>(pointer)
    }
    return nil
}
```

kneekey23: tried something similar to this but let me try again

Source/AwsCommonRuntimeKit/Utilities.swift
```diff
@@ -57,3 +57,27 @@ extension Bool {
         return self ? 1 : 0
     }
 }
+
+func fromPointer<T, P: PointerConformance>(ptr: T?) -> P? {
+   if let ptr = ptr {
+        let pointer = UnsafeMutablePointer<T>.allocate(capacity: 1)
+        pointer.initialize(to: ptr)
+       return P(OpaquePointer(pointer))
```
justinboswell, line 65: whitespace is weird

Source/AwsCommonRuntimeKit/Utilities.swift
```diff
@@ -57,3 +57,27 @@ extension Bool {
         return self ? 1 : 0
     }
 }
+
```
justinboswell, line 60:
```swift
// Ensure that any UnsafeXXXPointer is ALWAYS initialized to either nil or a value in a single call. Prevents the
// case where you create an UnsafeMutableWhatever and do not call `initialize()` on it, resulting in a non-null but // also invalid pointer
```

Source/AwsCommonRuntimeKit/auth/credentials/CRTCredentialsProvider.swift
```diff
@@ -68,24 +65,19 @@ class WrappedCRTCredentialsProvider {
...
+        let pointer: UnsafeMutablePointer<CRTCredentialsProviderShutdownOptions>? = fromOptionalPointer(ptr: shutDownOptions)
+        let shutDownOptionsC = aws_credentials_provider_shutdown_options(shutdown_callback: { userData in
+            guard let userData = userData else {
+                return
+            }
+            let pointer = userData.assumingMemoryBound(to: CRTCredentialsProviderShutdownOptions.self)
+            defer {pointer.deinitializeAndDeallocate()}
+            pointer.pointee.shutDownCallback()
+
+        }, shutdown_user_data: pointer)
```
justinboswell, line 78: Feels like this section needs a format/lint

Source/AwsCommonRuntimeKit/auth/credentials/CRTCredentialsProvider.swift
```diff
@@ -48,14 +48,11 @@ class WrappedCRTCredentialsProvider {
 
         })
         let shutDownOptions = Self.setUpShutDownOptions(shutDownOptions: shutDownOptions)
-        let intPointer = UnsafeMutablePointer<Int>.allocate(capacity: 1)
-        intPointer.pointee = 1
-        let atomicVar = aws_atomic_var(value: UnsafeMutableRawPointer(intPointer))
+        let intPointer: UnsafeMutableRawPointer = fromPointer(ptr: 1)
```
justinboswell, line 51: I hadn't noticed this code before. Does this imply that Swift doesn't have atomics?!

Source/AwsCommonRuntimeKit/io/EventLoopGroup.swift
```diff
@@ -12,24 +12,18 @@ public final class EventLoopGroup {
...
-            shutDownPtr.initialize(to: shutDownOptions)
-            let options = aws_shutdown_callback_options(shutdown_callback_fn: { (userData) in
-                guard let userdata = userData else {
-                    return
-                }
-                let pointer = userdata.assumingMemoryBound(to: ShutDownCallbackOptions.self)
-                defer { pointer.deinitializeAndDeallocate() }
-                pointer.pointee.shutDownCallback(pointer.pointee.semaphore)
+        let shutDownPtr: UnsafeMutablePointer<ShutDownCallbackOptions>? = fromOptionalPointer(ptr: shutDownOptions)
+        let options = aws_shutdown_callback_options(shutdown_callback_fn: { (userData) in
```
justinboswell, line 16: I've just realized that we've re-used this exact pattern at least 3 times, we should have a function that produces ShutdownCallbackOptions from a closure.

Source/AwsCommonRuntimeKit/io/TlsConnectionOptions.swift
```diff
@@ -9,10 +9,8 @@ public final class TlsConnectionOptions {
 
 	init(_ context: TlsContext, allocator: Allocator) {
 		self.allocator = allocator
-        let connectionOptionsPtr = UnsafeMutablePointer<aws_tls_connection_options>.allocate(capacity: 1)
-        //initialize pointer to empty connection options struct
-        zeroStruct(connectionOptionsPtr)
-        self.rawValue = connectionOptionsPtr
+
+        self.rawValue = allocatePointer()
```
justinboswell, line 13: Formatting?

Source/AwsCommonRuntimeKit/io/retryer/CRTAWSRetryToken.swift
```diff
@@ -7,17 +7,14 @@ public final class CRTAWSRetryToken {
     var rawValue: UnsafeMutablePointer<aws_retry_token>
 
     public init(retryStrategy: CRTRetryStrategy, allocator: Allocator = defaultAllocator) {
-        let intPointer = UnsafeMutablePointer<Int>.allocate(capacity: 1)
-        intPointer.pointee = 1
-        let atomicVar = aws_atomic_var(value: UnsafeMutableRawPointer(intPointer))
-        let retryStrategyPtr = UnsafeMutablePointer<CRTRetryStrategy>.allocate(capacity: 1)
-        retryStrategyPtr.initialize(to: retryStrategy)
+        let intPointer: UnsafeMutableRawPointer = fromPointer(ptr: 1)
```
justinboswell, line 10: Here's that atomic thing again. If Swift doesn't have atomics, we should make a wrapper in this library for a generic atomic<T> where T = integral type or pointer

Source/AwsCommonRuntimeKit/mqtt/MqttConnection.swift
```diff
@@ -462,7 +442,8 @@ public class MqttConnection {
         }, opCallbackPtr)
 
         defer {
-            pointers.deinitializeAndDeallocate()
+            topicPointer.deinitializeAndDeallocate()
```
justinboswell, line 445: defer isn't necessary here, topic and payload have already been used

Beginning to feel like we could benefit a ScopedPtr<P> that we can leave on the stack and calls deinitializeAndDeallocate() from its deinit or something

Source/AwsCommonRuntimeKit/auth/credentials/CRTCredentialsProvider.swift
kneekey23: formatted and linted

Source/AwsCommonRuntimeKit/io/EventLoopGroup.swift
kneekey23: made this!

Source/AwsCommonRuntimeKit/auth/credentials/CRTCredentialsProvider.swift
kneekey23: updated this to use our own Atomic that creates the c type

Source/AwsCommonRuntimeKit/mqtt/MqttConnection.swift
kneekey23: interesting idea, so a pointer as a class that deinitializes itself from its own deinit?

justinboswell: yes, essentially, it exists on the stack, and when it goes out of scope, it dies. Think of it as an automatically managed temporary

Source/AwsCommonRuntimeKit/crt/Atomic.swift
```diff
@@ -0,0 +1,13 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+import AwsCCommon
+
+struct Atomic<I> {
```
justinboswell, line 5: Hilarious. We don't even access the atomics in Swift. This is something I'll fix later when we move to the FFI library, this is unnecessarily part of the API in use here. Not Swift's problem.

justinboswell (approved):

Source/AwsCommonRuntimeKit/crt/Atomic.swift
kneekey23: yep its the c atomic type I need access to so I can pass it but I agree you can abstract that away!

justinboswell (approved):

PR #41, opened 2021-08-23
fix: customcredprovider

kneekey23:
*Description of changes:* This PR fixes the custom credentials provider pathway for a customer. It was completely broken before.

### DIFF
--- a/Source/AwsCommonRuntimeKit/auth/credentials/CRTAWSCredentialsProvider.swift
+++ b/Source/AwsCommonRuntimeKit/auth/credentials/CRTAWSCredentialsProvider.swift
@@ -240,9 +240,8 @@ public final class CRTAWSCredentialsProvider {
                 return
             }
             let pointer = userData.assumingMemoryBound(to: CRTCredentialsProviderShutdownOptions.self)
-            defer {pointer.deinitializeAndDeallocate()}
             pointer.pointee.shutDownCallback()
-
+            pointer.deinitializeAndDeallocate()
         }, shutdown_user_data: pointer)
 
         return shutDownOptionsC
--- a/Source/AwsCommonRuntimeKit/auth/credentials/CRTAWSCredentialsProvider.swift
+++ b/Source/AwsCommonRuntimeKit/auth/credentials/CRTAWSCredentialsProvider.swift
@@ -6,17 +6,17 @@ import AwsCIo
 import AwsCHttp
 
 public final class CRTAWSCredentialsProvider {
-    
+
     let allocator: Allocator
-    
+
     var rawValue: UnsafeMutablePointer<aws_credentials_provider>
-    
+
     init(credentialsProvider: UnsafeMutablePointer<aws_credentials_provider>,
          allocator: Allocator) {
         self.rawValue = credentialsProvider
         self.allocator = allocator
     }
-    
+
     public convenience init(fromProvider impl: CRTCredentialsProvider,
                             shutDownOptions: CRTCredentialsProviderShutdownOptions? = nil,
                             allocator: Allocator = defaultAllocator) throws {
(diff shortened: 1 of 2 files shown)


Source/AwsCommonRuntimeKit/auth/credentials/CRTAWSCredentialsProvider.swift
```diff
@@ -226,7 +231,39 @@ public final class CRTAWSCredentialsProvider {
...
+    static func setUpShutDownOptions(shutDownOptions: CRTCredentialsProviderShutdownOptions?)
+    -> aws_credentials_provider_shutdown_options {
+
+        let pointer: UnsafeMutablePointer<CRTCredentialsProviderShutdownOptions>? = fromOptionalPointer(ptr: shutDownOptions)
+        let shutDownOptionsC = aws_credentials_provider_shutdown_options(shutdown_callback: { userData in
+            guard let userData = userData else {
+                return
+            }
+            let pointer = userData.assumingMemoryBound(to: CRTCredentialsProviderShutdownOptions.self)
+            defer {pointer.deinitializeAndDeallocate()}
```
justinboswell, line 243: After the `shutDownCallback` call, you can just do this without a defer, right?

Source/AwsCommonRuntimeKit/auth/credentials/CRTAWSCredentialsProvider.swift
```diff
@@ -226,7 +231,39 @@ public final class CRTAWSCredentialsProvider {
...
+                                      _ callbackFn: (@convention(c)(OpaquePointer?, Int32, UnsafeMutableRawPointer?) -> Void)?,
+                                      _ userData: UnsafeMutableRawPointer?) -> Int32 {
+    guard let credentialsProvider = delegatePtr?.assumingMemoryBound(to: CRTCredentialsProvider.self) else {
+        return 1
+    }
+    guard let credentialCallbackData = userData?.assumingMemoryBound(to: CRTCredentialsProviderCallbackData.self) else {
+        return 1
+    }
+
+    credentialsProvider.pointee.getCredentials(credentialCallbackData: credentialCallbackData.pointee)
```
justinboswell, line 266: So this is going to pass a raw pointer on to your customer code, would you rather just give them a callback to call with their results, and you can convert it back into gross C stuff? Otherwise, they have to work out the calling mechanism themselves, whereas calling a swift function is idiomatic.

Source/AwsCommonRuntimeKit/auth/credentials/CRTCredentialsProvider.swift
```diff
@@ -3,84 +3,7 @@
...
-        }, shutdown_user_data: pointer)
-
-        return shutDownOptionsC
-    }
-
-    deinit {
-        implementationPtr.deinitializeAndDeallocate()
-        vTablePtr.deinitializeAndDeallocate()
-    }
 }
```
justinboswell, line 9: RED CODE IS BEST CODE

Source/AwsCommonRuntimeKit/auth/credentials/CRTAWSCredentialsProvider.swift
kneekey23: correct, updated code

Source/AwsCommonRuntimeKit/auth/credentials/CRTAWSCredentialsProvider.swift
kneekey23: not sure what you mean. if you look at this PR: https://github.com/awslabs/aws-sdk-swift/pull/362 its how I expose it to the customer with just a simple protocol and I then I have a wrapper but it just calls the crt bindings nothing gross

kneekey23: also im not passing the pointer but the pointee which is the swift type :)

Source/AwsCommonRuntimeKit/auth/credentials/CRTCredentialsProvider.swift
kneekey23: AGREED

Source/AwsCommonRuntimeKit/auth/credentials/CRTAWSCredentialsProvider.swift
justinboswell: Ohhhhhhh

justinboswell (approved):

PR #45, opened 2021-09-27
feat: add sha256 checksum extension

kneekey23:
*Description of changes:* Adds sha256 checksum extension on a string for calculations in middleware in the sdk

### DIFF
--- a/Source/AwsCommonRuntimeKit/Utilities.swift
+++ b/Source/AwsCommonRuntimeKit/Utilities.swift
@@ -42,16 +42,14 @@ extension String {
         return byteCursor.toData().base64EncodedString()
     }
 
-    public func base64EncodedSha256(allocator: Allocator = defaultAllocator, truncate: Int = 0) -> String? {
+    public func base64EncodedSha256(allocator: Allocator = defaultAllocator, truncate: Int = 0) -> String {
         let input: UnsafePointer<aws_byte_cursor> = fromPointer(ptr: self.awsByteCursor)
         let emptyBuffer: UInt8 = 0
         let bufferPtr: UnsafeMutablePointer<UInt8> = fromPointer(ptr: emptyBuffer)
         let buffer = aws_byte_buf(len: 0, buffer: bufferPtr, capacity: 32, allocator: allocator.rawValue)
         let output: UnsafeMutablePointer<aws_byte_buf> = fromPointer(ptr: buffer)
 
-        guard AWS_OP_SUCCESS == aws_sha256_compute(allocator.rawValue, input, output, truncate) else {
-            return nil
-        }
+        aws_sha256_compute(allocator.rawValue, input, output, truncate)
 
         let byteCursor = aws_byte_cursor_from_buf(output)
         return byteCursor.toData().base64EncodedString()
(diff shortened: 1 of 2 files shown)


Source/AwsCommonRuntimeKit/Utilities.swift
```diff
@@ -41,6 +41,21 @@ extension String {
         let byteCursor = aws_byte_cursor_from_buf(output)
         return byteCursor.toData().base64EncodedString()
     }
+
+    public func base64EncodedSha256(allocator: Allocator = defaultAllocator, truncate: Int = 0) -> String? {
```
graebm, line 45: Could this just return `String` instead of `String?`
I don't think it can ever actually fail. aws_sha256_compute() can only fail if allocation fails (that's fatal now) or the output buffer isn't big enough (you're creating the output buffer, so there's no chance of messing that up)

I understand if you want to stick with `String?` because the other existing `base64EncodedMD5()` function returns `String?`

Source/AwsCommonRuntimeKit/Utilities.swift
```diff
@@ -41,6 +41,19 @@ extension String {
         let byteCursor = aws_byte_cursor_from_buf(output)
         return byteCursor.toData().base64EncodedString()
     }
+
+    public func base64EncodedSha256(allocator: Allocator = defaultAllocator, truncate: Int = 0) -> String {
+        let input: UnsafePointer<aws_byte_cursor> = fromPointer(ptr: self.awsByteCursor)
+        let emptyBuffer: UInt8 = 0
+        let bufferPtr: UnsafeMutablePointer<UInt8> = fromPointer(ptr: emptyBuffer)
+        let buffer = aws_byte_buf(len: 0, buffer: bufferPtr, capacity: 32, allocator: allocator.rawValue)
```
graebm, line 49: Is it possible to use [AWS_SHA256_LEN](https://github.com/awslabs/aws-c-cal/blob/e364e9a42075447acf0ff9386a83c1ca19955b89/include/aws/cal/hash.h#L12) instead of `32` here? I don't know how accessible C stuff is from Swift

Test/AwsCommonRuntimeKitTests/crt/UtilityTests.swift
```diff
@@ -22,5 +22,19 @@ class UtilityTests: XCTestCase {
         
         XCTAssertEqual(md5, "iB0/3YSo7maijL0IGOgA9g==")
     }
+    
+    func testSha256() throws {
+        let hello = "Hello"
+        let sha256 = hello.base64EncodedSha256()
+        XCTAssertEqual(sha256, "GF+NsyJx/iX1Yab8k4suJkMG7DBO2lGAB9F2SCY4GWk=")
```
graebm, line 29: probably worth testing a sha of the empty string "" -> "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855" since it's edge-case-y

graebm (approved):
fix (or not) and ship

Source/AwsCommonRuntimeKit/Utilities.swift
justinboswell: It's accessible if it's a constant, not a macro

kneekey23: updated :)

Test/AwsCommonRuntimeKitTests/crt/UtilityTests.swift
kneekey23: added a test with empty string

PR #50, opened 2021-10-20
feat: wrap IMDS Client

kneekey23:
*Description of changes:*
Adds the wrapping of the EC2 IMDS Client for use by the SDK

### DIFF
--- a/Source/AwsCommonRuntimeKit/Utilities.swift
+++ b/Source/AwsCommonRuntimeKit/Utilities.swift
@@ -47,6 +47,10 @@ extension aws_byte_buf {
     func toByteBuffer() -> ByteBuffer {
         return ByteBuffer(ptr: self.buffer, len: self.len, capacity: self.capacity)
     }
+    
+    func toString() -> String {
+        return String(cString: self.buffer)
+    }
 }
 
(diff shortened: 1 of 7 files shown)


Source/AwsCommonRuntimeKit/Utilities.swift
```diff
@@ -49,6 +49,22 @@ extension aws_byte_buf {
     }
 }
 
+extension aws_array_list {
```
justinboswell, line 52: Hahahahaha, one can indeed write super unsafe code in any language ;)

Source/AwsCommonRuntimeKit/auth/imds/CRTIMDSClientIAMProfileCallbackData.swift
```diff
@@ -0,0 +1,14 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+
+public struct CRTIMDSClientIAMProfileCallbackData {
```
justinboswell, line 4: Curiosity: Do all these callback data structures actually need to be public?

PR #59, opened 2022-02-07
fix: update submodules to latest versions

kneekey23:
*Description of changes:* This PR updates all crt git submodules and fixes any build or api issues that came along with that. Also adds a handy script from @graebm to update submodules easily

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -80,7 +80,7 @@ awsCCommonPlatformExcludes.append("source/arch/intel/asm")
 awsCCommonPlatformExcludes.append("source/arch/arm/asm")
 #endif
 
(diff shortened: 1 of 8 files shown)


Package.swift
```diff
@@ -80,7 +80,7 @@ awsCCommonPlatformExcludes.append("source/arch/intel/asm")
 awsCCommonPlatformExcludes.append("source/arch/arm/asm")
 #endif
 
-var awsCIoPlatformExcludes = ["docs", "CODE_OF_CONDUCT.md", "codebuild"] + excludesFromAll
+var awsCIoPlatformExcludes = ["docs", "CODE_OF_CONDUCT.md", "codebuild", "PKCS11.md", "THIRD-PARTY-LICENSES.txt", "include/aws/io/private/pkcs11/v2.40/pkcs11.h", "include/aws/io/private/pkcs11/v2.40/pkcs11f.h", "include/aws/io/private/pkcs11/v2.40/pkcs11t.h"] + excludesFromAll
```
graebm, line 83: don't need to exclude `include/private/pkcs11/...` anymore
those files are gone (moved under `source/...`)

graebm (approved):

Package.swift
kneekey23: fixed

Issue #66, opened 2022-07-13
CRTAWSRetryToken is calling aws_retry_token_release on aws_retry_token after it is released


brennanMKE:
Below are the details I collected as a result of a crash that occurs sporadically. Memory is being released and later accessed resulting in a runtime exception.

```
==30709==ERROR: AddressSanitizer: heap-use-after-free on address 0x00013c542530 at pc 0x000105cabf58 bp 0x00016b4728b0 sp 0x00016b4728a8
WRITE of size 8 at 0x00013c542530 thread T6
    #0 0x105cabf54 in aws_atomic_fetch_sub_explicit+0x1b4 (Storage:arm64+0x100fdbf54)
    #1 0x105cac2c0 in aws_retry_token_release+0x68 (Storage:arm64+0x100fdc2c0)
    #2 0x105b637e4 in CRTAWSRetryToken.deinit+0x1bc (Storage:arm64+0x100e937e4)
    #3 0x105b639c0 in CRTAWSRetryToken.__deallocating_deinit+0x150 (Storage:arm64+0x100e939c0)
    #4 0x111f2a550 in _swift_release_dealloc+0x18 (libswiftCore.dylib:arm64+0x2d2550)
    #5 0x1058a6470 in (2) await resume partial function for RetryerMiddleware.handle<A>(context:input:next:)+0x2e0 (Storage:arm64+0x100bd6470)
    #6 0x10f225e14 in swift::runJobInEstablishedExecutorContext(swift::Job*)+0x80 (libswift_Concurrency.dylib:arm64+0x29e14)
```

I've set breakpoints and logged the pointer for `aws_retry_token` when it is released. It shows the same address. Below **0x160604700** is the same pointer address which is released many times.

```
👋 signRequestToCRT
👋 acquiring token 0x160604700
👋 releasing in aws_retry_token_release 0x160604700
👋 releasing in aws_retry_token_release 0x160604800
👋 releaseToken 2 0x160604700
👋 releasing in aws_retry_token_release 0x160604700
👋 releasing in aws_retry_token_release 0x160604800
👋 CRTAWSRetryToken deinit 0x160604700
👋 releasing in aws_retry_token_release 0x160604700
```

My assumption is this value should only be released once and only after it is no longer in use, which would be the `deinit` function.

brennanMKE:
[...]

PR #67, opened 2022-07-14
removes redundant call to aws_retry_token_release

brennanMKE:
*Issue #, if available:*

#66

*Description of changes:*

Identified a redundant memory release call and removed it. See issue for details.

### DIFF
--- a/Source/AwsCommonRuntimeKit/io/retryer/CRTAWSRetryToken.swift
+++ b/Source/AwsCommonRuntimeKit/io/retryer/CRTAWSRetryToken.swift
@@ -11,7 +11,4 @@ public final class CRTAWSRetryToken {
         self.rawValue = rawValue
     }
 
-    deinit {
-        aws_retry_token_release(rawValue)
-    }
 }
--- a/Source/AwsCommonRuntimeKit/io/retryer/CRTAWSRetryToken.swift
+++ b/Source/AwsCommonRuntimeKit/io/retryer/CRTAWSRetryToken.swift
@@ -11,4 +11,8 @@ public final class CRTAWSRetryToken {
         self.rawValue = rawValue
     }
 
+    deinit {
+        aws_retry_token_release(rawValue)
+    }
+
 }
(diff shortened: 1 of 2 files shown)


Source/AwsCommonRuntimeKit/io/retryer/CRTAWSRetryToken.swift
```diff
@@ -11,7 +11,4 @@ public final class CRTAWSRetryToken {
         self.rawValue = rawValue
     }
 
-    deinit {
-        aws_retry_token_release(rawValue)
-    }
```
graebm, line 16: I don't know enough about Swift/C bindings and object lifetime to say whether this is the right call.

At first glance, the pre-existing deinit here looks right. A "wrapper" swift class ensure the C object stays alive by keeping 1 reference-count on the C object. Then releases the reference-count when the Swift object is destroyed. If this was the last reference-count on the C object, it will be cleaned up. If something else has acquired a reference to the C object, then the C object stays alive until that reference is also realeased.

But there is a double free going on. And I can see that CRTAWSRetryStrategy ALSO has a `releaseToken()` function that will ALSO release a reference-count on the C object.

In Swift, someone should "release" the token by ceasing to reference it, not by manually calling a release() function, right?

But I don't know Swift well enough to understand the nitty gritty of what's going on in this code

graebm: Also, it would be great to add a unit test that reproduces this crash. Then we could be sure that the fix actually fixes it. The unit test could just use these retry classes directly, it wouldn't actually need to do any HTTP calls or anything

brennanMKE: Revised the PR with PRs also prepared for the other packages.

https://github.com/awslabs/aws-sdk-swift/pull/592
https://github.com/awslabs/smithy-swift/pull/422

Issue #69, opened 2022-07-16
HttpStream is activated without checking for an error


brennanMKE:

waahm7:
Fixed.

PR #70, opened 2022-07-16
handles errors for HttpStream

brennanMKE:
*Issue #, if available:*

#69

*Description of changes:*

The status code was not being used to check if there was a failure. This update now checks and throws and error if there is one.


waahm7:
Fixed in main

PR #72, opened 2022-09-15
WIP - C-IO refactor

waahm7:
*Issue #, if available:*

*Description of changes:*
- Refactors Host Resolver to use aws allocator and deallocator


Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -30,26 +28,22 @@ public final class DefaultHostResolver: HostResolver {
...
-        )
-
-        self.config = fromPointer(ptr: config)
 
+        var options = aws_host_resolver_default_options(max_entries: maxHosts,
+                el_group: elg.rawValue,
+                shutdown_options: shutdownCallbackOptionPtr,
+                system_clock_override_fn: nil)
+        rawValue = aws_host_resolver_new_default(allocator.rawValue, &options)
+        config = try! allocator.allocate(capacity: 1)
```
waahm7, line 37: Try! is to just crash the application by throwing exception if we can’t allocate memory. I think swift also uses this approach (https://forums.swift.org/t/allocating-memory-without-a-crash/29758/3). Alternatively, we can just throw an exception which Swift SDK can handle.

Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -30,26 +28,22 @@ public final class DefaultHostResolver: HostResolver {
...
-            impl_data: nil
-        )
-
-        self.config = fromPointer(ptr: config)
 
+        var options = aws_host_resolver_default_options(max_entries: maxHosts,
+                el_group: elg.rawValue,
+                shutdown_options: shutdownCallbackOptionPtr,
+                system_clock_override_fn: nil)
+        rawValue = aws_host_resolver_new_default(allocator.rawValue, &options)
```
waahm7, line 36: Swift doesn’t have a pass by reference. &options implicitly converts options to a pointer to pass to the C function. The pointer generated by the ampersand syntax is only valid for the duration of that function call. We can use this when we don’t need long lived pointers. 
Referece: https://developer.apple.com/forums/thread/674633

Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -58,20 +52,25 @@ public final class DefaultHostResolver: HostResolver {
     }
 
     public func resolve(host: String) async throws -> [HostAddress] {
+        let resolverOptions: UnsafeMutablePointer<ResolverOptions> = try! allocator.allocate(capacity: 1)
+        defer {
+            allocator.release(resolverOptions)
+        }
```
waahm7, line 58: I will try to refactor this into & syntax, as it is being used in only one function call. Keeping it for now to get this approach reviewed when we need a pointer for the lifetime of a function.

Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -30,26 +28,22 @@ public final class DefaultHostResolver: HostResolver {
...
-        self.hostResolverOptionsPtr = fromPointer(ptr: options)
-        self.rawValue = aws_host_resolver_new_default(allocator.rawValue, hostResolverOptionsPtr)
-
-        let config = aws_host_resolution_config(
-            impl: aws_default_dns_resolve,
-            max_ttl: maxTTL,
-            impl_data: nil
-        )
-
-        self.config = fromPointer(ptr: config)
```
waahm7, line 46: This function uses Swift allocate call to callocate memory. I have replaced it with Allocator.allocate

Source/AwsCommonRuntimeKit/io/HostResolver.swift
DmitriyMusatkin: you might want to confirm we still need this. looks like allocate implementation relies on aws_mem_calloc returning null in guard let statement to throw exception, but calloc no longer returns null in latest crt and instead exits.

Source/AwsCommonRuntimeKit/crt/Allocator.swift
```diff
@@ -40,11 +40,14 @@ internal extension Allocator {
...
 
     /**
-     * Releases memory allocated by this allocator.
+     * Releases memory allocated by this allocator. We need to call deinitialize first so that Swift compiler's ARC
+     * can decreases the reference count of instances held by this pointer and release them if needed.
      *
      * - Parameter pointer: The pointer to allocated data.
+     * - Parameter count:
      */
-    func release<T>(_ pointer: UnsafeMutablePointer<T>?) {
```
DmitriyMusatkin, line 47: nit: setting default to 1 seems like an easy way to get into strange memory bugs down the road. as far as i can tell from docs count here is the number of elements that were allocated (ie same thing as capacity in allocate function), so its up to caller of this function to keep track of when to keep default and when to provide value and im guessing most of the time people will forget.

Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -30,26 +28,22 @@ public final class DefaultHostResolver: HostResolver {
...
 
+        var options = aws_host_resolver_default_options(max_entries: maxHosts,
```
graebm, line 32: trivial: I think the reason the old code was crazy, and the reason you have to use `var` here instead of `let` is just that the C function forgot to mark the options arg as `const`

we should fix the C function. "options" structs should always be passed as `const`

Source/AwsCommonRuntimeKit/io/HostResolver.swift
graebm: I'm not sure why `config` is a member variable. This is just another "options struct" for calls to `aws_host_resolver_resolve_host()`.

 (and that function is another one that failed to mark the argument as `const`, which is why Swift is making us do weird tricks with the pointer types)

Source/AwsCommonRuntimeKit/io/HostResolver.swift
graebm: hey look, it's another call to some `_new()` function where we never checked whether it returned NULL

PR #73, opened 2022-09-21
Updates crt to latest version

waahm7:
*Issue #, if available:*

*Description of changes:*


Package.swift
```diff
@@ -125,6 +125,15 @@ let cSettings: [CSetting] = [
 //    .unsafeFlags(cFlags),
     .define("DEBUG_BUILD", .when(configuration: .debug))
 ]
+#if os(Linux)
+let cSettingsIO: [CSetting] = [
+//    .unsafeFlags(cFlags),
+    .define("DEBUG_BUILD", .when(configuration: .debug)),
+    .define("USE_S2N")
+]
```
graebm, line 133: just copy `cSettings`
```
let cSettingsIO: [CSetting] = cSettings + [
    .define("USE_S2N")
]
```

Source/AwsCommonRuntimeKit/http/HttpClientConnectionManager.swift
```diff
@@ -16,30 +17,38 @@ public class HttpClientConnectionManager {
         self.options = options
         self.allocator = allocator
         let shutDownPtr: UnsafeMutablePointer<ShutDownCallbackOptions>? = fromOptionalPointer(ptr: options.shutDownOptions)
+        //Todo: add new http2 options as properties in HttpClientConnectionOptions
```
graebm, line 20: don't worry about it yet

Source/AwsCommonRuntimeKit/http/HttpClientConnectionManager.swift
```diff
@@ -16,30 +17,38 @@ public class HttpClientConnectionManager {
...
-                                                             proxy_options: options.proxyOptions?.rawValue,
-                                                             proxy_ev_settings: options.proxyEnvSettings?.rawValue,
-                                                             max_connections: options.maxConnections,
-                                                             shutdown_complete_user_data: shutDownPtr,
-                                                             shutdown_complete_callback: { (userData) in
-                                                                guard let userData = userData else {
-                                                                    return
-                                                                }
+                initial_window_size: options.initialWindowSize,
+                socket_options: options.socketOptions.rawValue,
```
graebm, line 23: we could make submodule updates painless by zeroing out options-structs, before setting each member. That's the way they're intended to be used

Using the Swift-generated initializer functions like this means things break whenever we take an update and members get added

don't make the change in this PR. let's do that change in its own PR, because it will be a LOT of changed lines

Source/AwsCommonRuntimeKit/http/HttpClientConnectionManager.swift
```diff
@@ -64,21 +73,23 @@ public class HttpClientConnectionManager {
                 return
             }
             let callbackData = userData.assumingMemoryBound(to: HttpClientConnectionCallbackData.self)
-            defer {callbackData.deinitializeAndDeallocate()}
+            defer {
+                callbackData.deinitializeAndDeallocate()
+            }
```
graebm, line 78: Revert lines in this PR that are just style or whitespace changes.

They makes this change hard to review.
And they makes the git history sloppy. On BIG projects like this, someone will inevitably be like "why is this line doing what it's doing" and then go dig through the git history. They're going to see a PR with an interesting description, look at the PR, and be like "dammit that was misleading this is just an arbitrary whitespace change", and it makes the dig harder.

If this code was being auto-formatted before, let's get our auto-formatter configured the same way.

If this code was NOT being auto-formatted, please turn your auto-formatter off for now.

Let's do a separate followup PR where we run an auto-formatter on EVERYTHING and then we can one git commit where a TON of code changes, but the line history will clearly show that this commit was "Autoformat all the code".

This is what we do in nearly all our other repos. Autoformat the code, and have a CI check that enforces it, so that lines aren't arbitrarily changing all the time, and so we're not wasting time nitpicking whitespace style in code reviews

Source/AwsCommonRuntimeKit/http/HttpClientConnectionManager.swift
```diff
@@ -16,30 +17,38 @@ public class HttpClientConnectionManager {
...
-                                                                    return
-                                                                }
+                initial_window_size: options.initialWindowSize,
+                socket_options: options.socketOptions.rawValue,
+                tls_connection_options: options.tlsOptions?.rawValue,
+                http2_prior_knowledge: false,
+                monitoring_options: options.monitoringOptions?.rawValue,
+                host: options.hostName.awsByteCursor,
+                port: options.port,
+                initial_settings_array: nil,
```
graebm, line 29: likewise, please get these spaced the same way they used to be spaced. for the sake of clean diffs

we'll fix the formatting in a future pass

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -5,22 +5,26 @@ import AwsCIo
...
     public var length: Int64
-
+    private var refCount: aws_ref_count
     public init(_ impl: AwsStream, allocator: Allocator = defaultAllocator) {
         self.length = Int64(impl.length)
         self.implPointer = fromPointer(ptr: impl)
-        self.rawValue = aws_input_stream(allocator: allocator.rawValue, impl: self.implPointer, vtable: &vtable)
+        refCount = aws_ref_count()
+        aws_ref_count_init(&refCount, nil, { obj in })
+        self.rawValue = aws_input_stream(impl: self.implPointer, vtable: &vtable, ref_count: refCount)
```
graebm, line 24: this may compile and run, but it's not correct

We need to get it so that C calls to the `acquire()` and `release()` vtable functions can keep this Swift object alive. Let's chat directly about this. Too complex for a quick write up

Package.swift
waahm7: Improved.

Source/AwsCommonRuntimeKit/http/HttpClientConnectionManager.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/http/HttpClientConnectionManager.swift
waahm7: Thank you for the detailed explanation. I have fixed it.

Source/AwsCommonRuntimeKit/http/HttpClientConnectionManager.swift
waahm7: This is a good idea. We can fix it keeping the future changes in mind w.r.t ABI compatibility.

waahm7:
S2N has a work in progress PR (https://github.com/aws/s2n-tls/pull/3512) that will fix the Linux CI.

graebm (approved):

Source/AwsCommonRuntimeKit/http/HttpClientConnectionManager.swift
ganeshnj: Usually prefer a backlog item while adding TODO. 

Could you add "Add support for H2" github issue and add the link here?

ganeshnj (approved):

PR #75, opened 2022-09-28
Improve formatting with SwiftFormat

waahm7:
*Description of changes:*
- Refactors whole repo using [SwiftFormat](https://github.com/nicklockwood/SwiftFormat)
- Used the following command 
_swiftformat . --swiftversion 5.7 --disable wrapMultilineStatementBraces --maxwidth 145 --closingparen same-line_

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -1,0 +1,118 @@
+# file options
+--exclude Pods
+
+# rules with format options
+
+--enable braces
+  --allman false
+
+--enable elseOnSameLine
+  --elseposition same-line
+
+--enable fileHeader
+  --header "//\n// Copyright Amazon.com Inc. or its affiliates.\n// All Rights Reserved.\n//\n// SPDX-License-Identifier: Apache-2.0\n//"
+
+--disable hoistPatternLet
+  --patternlet inline
+
+--disable indent
+  --ifdef outdent
+  --indent 4
+  --indentcase false
+  --xcodeindentation disabled
+
+--enable linebreaks
+  --linebreaks lf
+
+--enable numberFormatting
+  --binarygrouping none
+  --decimalgrouping 3
+  --exponentcase lowercase
+  --exponentgrouping disabled
+  --fractiongrouping disabled
+  --hexgrouping none
+  --hexliteralcase lowercase
+  --octalgrouping none
+
+--enable spaceAroundOperators
+  --operatorfunc spaced
+
+--enable redundantSelf
+  --self init-only
+  --selfrequired 
+
+--enable semicolons
+  --semicolons never
+
+--disable sortedImports
+  --importgrouping testable-bottom
+
+--enable spaceAroundOperators
+  --operatorfunc spaced
+
+--enable trailingClosures
+  --trailingclosures 
+
+--disable trailingCommas
+  --commas inline
+
+--enable trailingSpace
+  --trimwhitespace always
+
+--disable unusedArguments
+  --stripunusedargs closure-only
+
+--enable void
+  --empty void
+
+--disable wrapArguments
+  --closingparen same-line
+  --wraparguments after-first
+  --wrapcollections before-first
+
+# standalone rules
+--disable andOperator
+--disable anyObjectProtocol
+--disable blankLinesAroundMark
+--disable blankLinesAtEndOfScope
+--disable blankLinesAtStartOfScope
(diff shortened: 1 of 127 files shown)


Source/AwsCommonRuntimeKit/Utilities.swift
```diff
@@ -55,8 +62,8 @@ extension aws_array_list {
         var arrayList = self
         var newArray: [String] = Array(repeating: "", count: length)
 
-        for index  in 0..<length {
-            var val: UnsafeMutableRawPointer! = nil
+        for index in 0 ..< length {
```
ganeshnj, line 65: this should be `0..length` whitespaces in range are odd.

Source/AwsCommonRuntimeKit/auth/credentials/CRTCredentials.swift
```diff
@@ -21,9 +24,9 @@ public final class CRTCredentials {
                 expirationTimeout: UInt64,
                 allocator: Allocator = defaultAllocator) {
         self.rawValue = aws_credentials_new(allocator.rawValue,
-                                            accessKey.awsByteCursor,
-                                            secret.awsByteCursor,
-                                            sessionToken?.awsByteCursor ?? "".awsByteCursor, expirationTimeout)
+                                       accessKey.awsByteCursor,
+                                       secret.awsByteCursor,
+                                       sessionToken?.awsByteCursor ?? "".awsByteCursor, expirationTimeout)
```
ganeshnj, line 29: is this autoformat?

aligning params label is common way like it was before.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -308,22 +311,22 @@ extension ByteBuffer: AwsStream {
     }
 }
 
-extension ByteBuffer {
+public extension ByteBuffer {
```
ganeshnj, line 314: this is risky as it will put all methods public by default and can lead to accidental public visibility of an internal only method/prop.

default is internal which is inside module only.

Source/Elasticurl/Elasticurl.swift
```diff
@@ -1,45 +1,47 @@
...
     public var logLevel: LogLevel = .trace
     public var verb: String = "GET"
     public var caCert: String?
     public var caPath: String?
     public var certificate: String?
     public var privateKey: String?
-    public var connectTimeout: Int = 3000
-    public var headers: [String: String] = [String: String]()
+    public var connectTimeout: Int = 3_000
+    public var headers: [String: String] = .init()
```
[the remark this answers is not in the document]
ganeshnj: is more common way
```suggestion
    public var headers: [String: String] = [:]
```

Source/Elasticurl/Elasticurl.swift
```diff
@@ -168,7 +169,7 @@ struct Elasticurl {
             context.alpnList.append("h2")
         }
 
-        if argumentsDict["w"] == nil && argumentsDict["W"] == nil {
+        if argumentsDict["w"] == nil, argumentsDict["W"] == nil {
```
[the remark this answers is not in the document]
ganeshnj: yup, let's keep using operator.

Source/Elasticurl/ElasticurlOptions.swift
```diff
@@ -1,10 +1,14 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
 
 import AwsCommonRuntimeKit
 
-//swiftlint:disable identifier_name
-public struct ElasticurlOptions {
+// swiftlint:disable identifier_name
+public enum ElasticurlOptions {
```
ganeshnj, line 11: if static let are still as they are, there is no point converting to enum.

This is more like obj-c pattern.

.swiftformat
```diff
@@ -0,0 +1,118 @@
+# file options
```
ganeshnj, line 1: are we aligned on using swiftformat?

ganeshnj (changes requested):

.swiftformat
ganeshnj: https://github.com/awslabs/aws-crt-swift/pull/70/files

we would need replacement for CI also, like swiftlint does today.

PR #76, opened 2022-09-29
Fix error handling logic & remove MQTT bindings

waahm7:
*Description of changes:*
- Replace three different error classes with only one error class
- Remove MQTT bindings


.swiftlint.yml
```diff
@@ -34,7 +35,6 @@ opening_brace: error
 return_arrow_whitespace: error
 statement_position:
   severity: error
-todo: warning
```
ganeshnj, line 37: this should stay to keep track of TODOs.

Source/AwsCommonRuntimeKit/crt/CRTError.swift
```diff
@@ -1,12 +1,27 @@
...
-
-import Foundation
 import AwsCCommon
 
-public enum CRTError: Error {
-  case fileNotFound(String)
-  case memoryAllocationFailure
-  case stringConversionError(UnsafePointer<aws_string>?)
-  case crtError(AWSError)
+public struct CRTError: Error {
```
ganeshnj, line 5: Original question: what is the motivation behind not using specific errors? 

this is not ideal way to handle error in Swift. This is more like obj-c version where only one error type exists. This makes the code simple here but less than ideal for consumer.

A user friendly representation for something like https://github.com/awslabs/aws-c-sdkutils/blob/endpoint20-eval/include/aws/sdkutils/sdkutils.h#L26
```c
enum aws_sdkutils_errors {
    AWS_ERROR_SDKUTILS_GENERAL =    AWS_ERROR_ENUM_BEGIN_RANGE(AWS_C_SDKUTILS_PACKAGE_ID),
    AWS_ERROR_SDKUTILS_PARSE_FATAL,
    AWS_ERROR_SDKUTILS_PARSE_RECOVERABLE,
    AWS_ERROR_SDKUTILS_ENDPOINTS_UNSUPPORTED_RULESET,
    AWS_ERROR_SDKUTILS_ENDPOINTS_PARSE_FAILED,
    AWS_ERROR_SDKUTILS_ENDPOINTS_EVAL_INIT_FAILED,
    AWS_ERROR_SDKUTILS_ENDPOINTS_EVAL_FAILED,
    AWS_ERROR_SDKUTILS_ENDPOINTS_EMPTY_RULESET,
    AWS_ERROR_SDKUTILS_ENDPOINTS_RULESET_EXHAUSTED,

    AWS_ERROR_SDKUTILS_END_RANGE = AWS_ERROR_ENUM_END_RANGE(AWS_C_SDKUTILS_PACKAGE_ID)
};
```

would be
```swift
enum CRTSDKUtilsError {
    case general
    case parseFatal
    case parseRecoverable
    case endpointsUnsupportedRuleset
    case endpointsParseFailed
    case endpointsEvalInitFailed
    case endpointsEvalFailed
    case endpointsEmptyRuleset
    case endpointsRulesetExhausted
}

extension CRTSDKUtilsError {
    public var errorCode: Int32 {
        get {
            rawValue
        }
    }

    public var errorMessage: String? {
        get {
            guard let stringPtr = aws_error_str(errorCode) else {
                return nil
            }
            return String(cString: stringPtr)
        }
    }

    public var errorName: String? {
        get {
            guard let stringPtr = aws_error_name(errorCode) else {
                return nil
            }
            return String(cString: stringPtr)
        }
    }
}
```

ganeshnj (changes requested):

.swiftlint.yml
waahm7: I have removed it temporarily as I will be adding Todos during refactoring to take care of later. I will enable it when we merge to main.

Source/AwsCommonRuntimeKit/auth/credentials/CRTAWSCredentialsProvider.swift
```diff
@@ -27,7 +27,7 @@ public final class CRTAWSCredentialsProvider {
                                                                 delegate_user_data: credProviderPtr)
 
         guard let credProvider = aws_credentials_provider_new_delegate(allocator.rawValue, &options) else {
-            throw AWSCommonRuntimeError() }
+            throw CRTError(errorCode: aws_last_error()) }
```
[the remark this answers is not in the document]
waahm7: This returns the last error code on current thread. This code is exactly the same as before. AWSCommonRuntimeError() was also calling aws_last_error.

graebm (approved):
fix & ship

Source/AwsCommonRuntimeKit/auth/credentials/CRTAWSCredentialsProvider.swift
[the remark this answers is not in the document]
graebm: Ganesh: It works in multi-threaded code because the error code is [thread local storage](https://en.wikipedia.org/wiki/Thread-local_storage)

Source/AwsCommonRuntimeKit/crt/CRTError.swift
graebm: I just chatted with Waqar about this.

We have plans to do like you said use enums instead, but we're going to do it at a later stage. It's going to require doing some code-gen whenever we update the C submodules (scan through all the int error codes from the CRT, and create a nicely named swift enum for each one)

In the meantime we're trying to go from 3 different error classes (where it's not clear which one to use) down to 1

Source/AwsCommonRuntimeKit/crt/CRTError.swift
```diff
@@ -1,12 +1,27 @@
...
+
```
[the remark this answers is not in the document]
waahm7: Thanks. done.

Source/AwsCommonRuntimeKit/crt/CRTError.swift
ganeshnj: I was trying to understand a little more this space and realized (I think) why there are 3 errors (1 is redundant)

Finally it should look like
```swift
/// Base enum for all CRT errors (Swift repo)
public enum CRTError: Error {
  case anyErrorThatPureSwiftCodeCanThrow
  case anyErrorThatPureSwiftCodeCanThrow2
  case awsError(AWSError) // case for all C errors
}

/// Base enum for all AWS CRT errors (C repo)
enum AWSError: Error {

}

Ideally we would want to have `enum` for `AWSError` but a `struct` is fine as of now
```

PR #78, opened 2022-10-03
Improve error handling by using a generated enum

waahm7:
*Description of changes:*

- Fix Error handling by creating a generated CrtError enum


Source/AwsCommonRuntimeKit/auth/credentials/CRTAWSCredentialsProvider.swift
```diff
@@ -27,7 +27,7 @@ public final class CRTAWSCredentialsProvider {
                                                                 delegate_user_data: credProviderPtr)
 
         guard let credProvider = aws_credentials_provider_new_delegate(allocator.rawValue, &options) else {
-            throw CRTError(errorCode: aws_last_error()) }
+            throw AWSCommonRuntimeError.CRTError(CRTError.init(fromErrorCode: aws_last_error())) }
```
ganeshnj, line 30: .init is not required for all `CRTError.init` calls

Source/Script/CRTErrorGenerator.swift
```diff
@@ -0,0 +1,80 @@
...
+        self.write("\n".data(using: .utf8)!)
+    }
+
+    func writeTab() {
+        self.write("    ".data(using: .utf8)!)
+    }
+}
+
+@main
+struct CRTErrorGenerator {
```
ganeshnj, line 22: Is there any way this can be tested?

Source/AwsCommonRuntimeKit/io/ClientBootstrap.swift
```diff
@@ -31,7 +31,7 @@ public final class ClientBootstrap {
             user_data: callbackDataPointer
         )
         guard let rawValue = aws_client_bootstrap_new(allocator.rawValue, &options) else {
-            throw CRTError(errorCode: aws_last_error())
+            throw AWSCommonRuntimeError.CRTError(CRTError.init(fromErrorCode: aws_last_error()))
```
ganeshnj, line 34: It would be nice to create a convenience `init` for CRTError than repeating `CRTError.init(fromErrorCode: aws_last_error())`

Source/AwsCommonRuntimeKit/crt/CRTErrorGenerated.swift
```diff
@@ -0,0 +1,306 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0.
+// This file is generated using Script/CRTErrorGenerator.swift.
+// Do not modify this file.
+
+import AwsCCommon
+
+public enum CRTError: Int32, Error {
```
ganeshnj, line 8:
```suggestion

/// Error type for CRT errors thrown from C code
public enum CRTError: Int32, Error {
```
```

File: Source/AwsCommonRuntimeKit/crt/AWSCommonRuntimeError.swift [hunk 5]
```diff
@@ -0,0 +1,6 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+
+public enum AWSCommonRuntimeError: Error {
```
ganeshnj, line 4:
```suggestion
/// Error type for all errors thrown by the Swift code including C runtime errors
public enum AWSCommonRuntimeError: Error {
```

ganeshnj (changes requested):

Source/AwsCommonRuntimeKit/auth/credentials/CRTAWSCredentialsProvider.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/crt/CRTErrorGenerated.swift
waahm7: Thanks, updated

Source/AwsCommonRuntimeKit/crt/AWSCommonRuntimeError.swift
waahm7: Thanks, updated.

Source/Script/CRTErrorGenerator.swift
waahm7: I am not sure if we can write unit tests. I have tested it by brute forcing from 0 to Int32.max and making sure they both generate equal number of cases.

ganeshnj (approved):

Source/AwsCommonRuntimeKit/crt/CRTError.swift
```diff
@@ -1,18 +1,30 @@
...
+    }
+
+    public init(fromErrorCode errorCode: Int) {
+        self = CRTError(rawValue: Int32(errorCode)) ?? .UNKNOWN_ERROR_CODE
+    }
 
-    public let errorCode: Int32
+    /// Use aws_last_error() as the error code
+    public init() {
+        self.init(fromErrorCode: aws_last_error())
```
graebm, line 16: I'd rather this be explicit, like `CRTError.makeFromLastError()`

Swift suggests [Begin names of factory methods with “make"](https://www.swift.org/documentation/api-design-guidelines/)

Source/AwsCommonRuntimeKit/crt/AWSCommonRuntimeError.swift
```diff
@@ -0,0 +1,9 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+
+/// Error type for all errors thrown by the Swift code including C runtime errors
+public enum AWSCommonRuntimeError: Error {
+
+    /// Use aws_last_error() as the error code if no value is provided
+    case AWSCRTError(_ crtError: CRTError = CRTError())
```
graebm, line 8: Again, I'd rather this be explicit

Source/AwsCommonRuntimeKit/crt/CRTError.swift
```diff
@@ -1,18 +1,30 @@
-//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
-//  SPDX-License-Identifier: Apache-2.0.
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0.
 import AwsCCommon
 
-public struct CRTError: Error {
+extension CRTError {
+    public init(fromErrorCode errorCode: Int32) {
+        self = CRTError(rawValue: errorCode) ?? .UNKNOWN_ERROR_CODE
```
graebm, line 7: I'm confused about the `??`

I don't see how `CRTError(rawValue: 0)` could evaluate to `nil`? But I don't know Swift super well yet...

Should this be like
`self.rawValue = errorCode !=0 ? errorCode : UNKNOWN_ERROR_CODE` ???

Source/AwsCommonRuntimeKit/crt/CRTError.swift
```diff
@@ -1,18 +1,30 @@
...
+    }
 
-    public let errorCode: Int32
+    /// Use aws_last_error() as the error code
+    public init() {
+        self.init(fromErrorCode: aws_last_error())
+    }
 
-    public let errorMessage: String
+    public var errorCode: Int32 {
```
graebm, line 19: trivial: I'd be fine if this was just `Int` instead of `Int32`

Source/AwsCommonRuntimeKit/crt/CRTErrorGenerated.swift
```diff
@@ -0,0 +1,307 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0.
+// This file is generated using Script/CRTErrorGenerator.swift.
+// Do not modify this file.
+
+import AwsCCommon
+
+/// Error type for CRT errors thrown from C code
+public enum CRTError: Int32, Error {
```
graebm, line 9: trivial: similar, is it worth making it `Int` instead of `Int32` just to be more swift-like?

Source/AwsCommonRuntimeKit/crt/CRTErrorGenerated.swift
```diff
@@ -0,0 +1,307 @@
...
+
+    case UNKNOWN_ERROR_CODE = -1
+
+    /// AWS-C-COMMON
+    case AWS_ERROR_SUCCESS = 0
```
graebm, line 14: I would be happy to omit SUCCESS from this list

Source/AwsCommonRuntimeKit/crt/CRTErrorGenerated.swift
```diff
@@ -0,0 +1,307 @@
...
+    case AWS_ERROR_FILE_INVALID_PATH = 44
+    case AWS_ERROR_MAX_FDS_EXCEEDED = 45
+    case AWS_ERROR_SYS_CALL_FAILURE = 46
+    case AWS_ERROR_C_STRING_BUFFER_NOT_NULL_TERMINATED = 47
+    case AWS_ERROR_STRING_MATCH_NOT_FOUND = 48
+    case AWS_ERROR_DIVIDE_BY_ZERO = 49
+    case AWS_ERROR_INVALID_FILE_HANDLE = 50
+    case AWS_ERROR_OPERATION_INTERUPTED = 51
+    case AWS_ERROR_DIRECTORY_NOT_EMPTY = 52
+    case AWS_ERROR_PLATFORM_NOT_SUPPORTED = 53
```
graebm, line 67: I'd love to do a transform on these names to make them more Swift-like
ex:
```
AWS_ERROR_PLATFORM_NOT_SUPPORTED -> platformNotSupported
```

there would need to be some special cases since we weren't 100% uniform in how errors were named

Source/AwsCommonRuntimeKit/crt/CRTError.swift
ganeshnj: Int size can be different depending on the arch, but I think it is okay to use Int here.

waahm7: Swift expects it to be Int32 instead of Int for passing to C functions. I tried using an Int and then created a simple init function which takes an int and converts it to Int32 to use it with C calls.

Source/AwsCommonRuntimeKit/crt/CRTError.swift
waahm7: The nil-coalescing operator (??) is used to supply a default value in case the Optional instance is nil. It's shortform of something like 
```
if CRTError(rawValue: errorCode) != nil { 
    self = CRTError(rawValue:errorCode) 
}
else {
    self = .UNKNOWN_ERROR_CODE
}
``` 
So if we pass any integer value that is not defined in CRTError like -100, it will evaluate it to UNKNOWN_ERROR_CODE instead of None, which matches the C implementation.

waahm7: The CRTError (rawValue: errorCode) converts rawValue to an Enum Case so if we pass 1, it will convert it to CRTError.AWS_ERROR_OOM. This is a handy function to convert the integer value of aws_last_error() to a matching enum case.

Source/AwsCommonRuntimeKit/crt/AWSCommonRuntimeError.swift
waahm7: Thanks, updated.

Source/AwsCommonRuntimeKit/crt/CRTError.swift
waahm7: Thanks, updated.

Source/AwsCommonRuntimeKit/crt/CRTErrorGenerated.swift
brennanMKE: See: https://www.swift.org/documentation/api-design-guidelines/#follow-case-conventions

Source/AwsCommonRuntimeKit/crt/CRTErrorGenerated.swift
waahm7: What should we do if there is a bug where we throw the error but error code is zero? Should we just fatal exit or throw errorSuccess or unknownErrorCode?

Source/AwsCommonRuntimeKit/crt/CRTErrorGenerated.swift
waahm7: Thanks, updated.

Source/AwsCommonRuntimeKit/crt/CRTErrorGenerated.swift
graebm: If the error-code is 0 and we need to throw the error, we should turn it into `AWS_ERROR_UNKNOWN`

waahm7: Fixed, thanks.

Source/AwsCommonRuntimeKit/crt/CRTErrorGenerated.swift
```diff
@@ -0,0 +1,306 @@
...
+
+import AwsCCommon
+
+/// Error type for CRT errors thrown from C code
+public enum CRTError: Int32, Error {
+
+    case unknownErrorCode = -1
+
+    /// AWS-C-COMMON
+    case oom = 1
```
waahm7, line 14: Replace it with outOfMemory

PR #79, opened 2022-10-05
feat: bindings for EndpointsRuleEngine

ganeshnj:
*Issue #, if available:*
https://github.com/awslabs/aws-sdk-swift/issues/624

*Description of changes:*

CRT implements the rule set evaluation system and this change creates bindings for Swift SDK. The usage of API is quite straight forward which is 1:1 of C APIs
```swift
// engine must be initialized with a rule set string
let engine = try CRTAWSEndpointsRuleEngine(ruleSetString: ruleSetString)

// populate context with available params
let context = try CRTAWSEndpointsRequestContext()
try context.add(name: "Region", value: "us-west-2")

// resolve
let endpoint = try engine.resolve(context: context)
let url = try endpoint?.getURL()
```


Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
```diff
@@ -0,0 +1,93 @@
...
+    /// - Returns: The type of the resolved endpoint
+    public func getType() -> CRTAWSEndpointsResolvedEndpointType {
+        let type = aws_endpoints_resolved_endpoint_get_type(rawValue)
+        return CRTAWSEndpointsResolvedEndpointType(rawValue: type)
+    }
+    
+    
+    /// Get the URL of the resolved endpoint
+    /// - Returns: The URL of the resolved endpoint
+    public func getURL() throws -> String? {
```
jbelkins, line 31: Why return an optional String from a throwing function?  Is there a scenario where `nil` would be returned, and it is not an error?

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
```diff
@@ -0,0 +1,93 @@
...
+        if success != 0 {
+            throw CRTError.awsError(AWSCommonRuntimeError())
+        }
+        return urlOut.pointee.toString()
+    }
+    
+    
+    /// Get the properties of the resolved endpoint
+    /// - Returns: The properties of the resolved endpoint
+    public func getProperties() throws -> [String: AnyHashable]? {
```
jbelkins, line 43: Same as above re: optionality of return value

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
```diff
@@ -0,0 +1,93 @@
...
+        let propsOut = UnsafeMutablePointer<aws_byte_cursor>.allocate(capacity: 1)
+        let success = aws_endpoints_resolved_endpoint_get_properties(rawValue, propsOut)
+        if success != 0 {
+            throw CRTError.awsError(AWSCommonRuntimeError())
+        }
+        guard let data = propsOut.pointee.toString()?.data(using: .utf8) else {
+            return nil
+        }
+        return try? JSONSerialization.jsonObject(with: data) as? [String: AnyHashable]
```
jbelkins, line 52: Why hide serialization errors with `try?` here?  Better to throw them back to the caller for easier debugging

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
ganeshnj: Yes, for whatever reason CRT return a non-zero response in Line33, here we want to bubble up the error thrown by the CRT than just return a nil.

Secondly, pointee.toString() returns a nullable string. Sending a default empty string is not correct and we will endup in some strange case in the consumer side.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
ganeshnj: Same as above.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
ganeshnj: Good call.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
jbelkins: Agreed, returning an empty string in place of the nil doesn't solve anything, even if it allows the return type to be changed from `String?` to `String`.

What I'm asking is if we should be throwing an error instead of returning nil, i.e.
```
guard let urlString = urlOut.pointee.toString() else {
    throw CRTError.awsError(/* insert appropriate error here */)
}
return urlString
```

Because if returning `nil` without throwing an error is abnormal/unexpected, we should not permit it.  And if it isn't unexpected, then at a minimum our documentation comment for this function should say that it may be nil & why.

ganeshnj: It should be okay throwing error, it won't be `awsError` but rather a custom error. It would be good experience too.

ganeshnj: I was thinking more and trying to see options and convinced that returning `nil` is better here and let the caller handle the issue. Reason being, this is a binding that should just work as wrapper around the C code. Further handling of error/nil must be done in the next layer i.e. AWSClientRuntime.

This also aligns with the existing bindings such as https://github.com/awslabs/aws-crt-swift/blob/2565815e8963455d8298afa4d5a8a44544e2b03c/Source/AwsCommonRuntimeKit/auth/credentials/CRTCredentials.swift#L32

TLDR, keeping it it is.

jbelkins: OK.  We can keep this for now if this follows the convention in the project.  But one of the functions of these bindings should be to enforce guarantees in Swift that can't be enforced in C, so that the output from CRT functions excludes undefined states like nil result and no error.

Put another way, if I wanted to test everything for `nil` before using it, I'd be a Java developer.  ;)

jbelkins (approved):

.github/workflows/ci.yml
```diff
@@ -20,10 +20,6 @@ jobs:
       - uses: actions/checkout@v1
       - name: GitHub Action for SwiftLint
         uses: norio-nomura/action-swiftlint@3.1.0
-      - name: GitHub Action for SwiftLint with --strict
-        uses: norio-nomura/action-swiftlint@3.1.0
-        with:
-          args: --strict
```
ganeshnj, line 26: strict is ignoring the `.swiftlint` config file and redundant to the line 20 after removing strict.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsRequestContext.swift
```diff
@@ -0,0 +1,55 @@
...
+        }
+        let namePtr: UnsafeMutablePointer<aws_byte_cursor> = fromPointer(ptr: name.awsByteCursor)
+        let valuePtr: UnsafeMutablePointer<aws_byte_cursor> = fromPointer(ptr: value.awsByteCursor)
+        let success = aws_endpoints_request_context_add_string(allocator.rawValue,
+                                                               rawValue,
+                                                               namePtr.pointee,
+                                                               valuePtr.pointee)
+        if success != 0 {
+            throw CRTError.awsError(AWSCommonRuntimeError())
+        }
```
waahm7, line 37: We don't need a pointer to byte cursor.
```suggestion
       if aws_endpoints_request_context_add_string(allocator.rawValue,
                                                               rawValue,
                                                               name.awsByteCursor,
                                                               value.awsByteCursor) != AWS_OP_SUCCESS {
            throw CRTError.crtError(AWSError(errorCode: aws_last_error()))
        }
```

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsRequestContext.swift
```diff
@@ -0,0 +1,55 @@
...
+    /// - Parameters:
+    ///   - name: The name of the parameter
+    ///   - value: The value of the parameter
+    ///   - allocator: The allocator to use for the parameter
+    public func add(name: String, value: Bool?, allocator: Allocator = defaultAllocator) throws {
+        guard let value = value else {
+            return
+        }
+        let namePtr: UnsafeMutablePointer<aws_byte_cursor> = fromPointer(ptr: name.awsByteCursor)
+        let success = aws_endpoints_request_context_add_boolean(allocator.rawValue, rawValue, namePtr.pointee, value)
```
waahm7, line 50: Same as above. We don't need a pointer to cursor. If we allocate a pointer, we need to pair it with a deallocation.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
```diff
@@ -0,0 +1,88 @@
...
+public class CRTAWSEndpointResolvedEndpoint {
+    let rawValue: OpaquePointer
+
+    /// Initialize a new resolved endpoint
+    /// - Parameter rawValue: The raw value of the resolved endpoint
+    public init(rawValue: OpaquePointer) {
+        self.rawValue = rawValue
+
+        aws_endpoints_resolved_endpoint_acquire(rawValue)
+    }
```
waahm7, line 17: Does this init needs to be public? We would prefer not exposing pointers to other layers.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
```diff
@@ -0,0 +1,88 @@
...
+            throw CRTError.awsError(AWSCommonRuntimeError())
+        }
+        return errorOut.pointee.toString()
+    }
+
+    /// Get headers of the resolved endpoint
+    /// - Returns: The headers of the resolved endpoint
+    public func getHeaders() throws -> [String: [String]]? {
+        let headersOut: UnsafeMutablePointer<UnsafePointer<aws_hash_table>?>
+            = UnsafeMutablePointer<UnsafePointer<aws_hash_table>?>.allocate(capacity: 1)
```
waahm7, line 66: Same as above

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsRuleEngine.swift
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        self.rawValue = rawValue
+    }
+
+    /// Resolve an endpoint from the rule engine using the provided request context
+    /// - Parameter context: The request context to use for endpoint resolution
+    /// - Returns: The resolved endpoint
+    public func resolve(context: CRTAWSEndpointsRequestContext) throws -> CRTAWSEndpointResolvedEndpoint? {
+        let resolvedEndpoint: UnsafeMutablePointer<OpaquePointer?>? = UnsafeMutablePointer.allocate(capacity: 1)
```
waahm7, line 29: Same as above.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
```diff
@@ -0,0 +1,88 @@
...
+        let success = aws_endpoints_resolved_endpoint_get_error(rawValue, errorOut)
+        if success != 0 {
+            throw CRTError.awsError(AWSCommonRuntimeError())
+        }
+        return errorOut.pointee.toString()
+    }
+
+    /// Get headers of the resolved endpoint
+    /// - Returns: The headers of the resolved endpoint
+    public func getHeaders() throws -> [String: [String]]? {
```
waahm7, line 64: Would it be possible to write tests for this function?

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
```diff
@@ -0,0 +1,88 @@
...
+        if success != 0 {
+            throw CRTError.awsError(AWSCommonRuntimeError())
+        }
+        return urlOut.pointee.toString()
+    }
+
+    /// Get the properties of the resolved endpoint
+    /// - Returns: The properties of the resolved endpoint
+    public func getProperties() throws -> [String: AnyHashable]? {
+        let propsOut = UnsafeMutablePointer<aws_byte_cursor>.allocate(capacity: 1)
```
waahm7, line 40: allocations should be paired with deallocations. Something like
```suggestion
        let propsOut = UnsafeMutablePointer<aws_byte_cursor>.allocate(capacity: 1)
        defer {
            allocator.release(propsOut)
        }
```

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
```diff
@@ -0,0 +1,88 @@
...
+        guard let data = propsOut.pointee.toString()?.data(using: .utf8) else {
+            return nil
+        }
+        return try JSONSerialization.jsonObject(with: data) as? [String: AnyHashable]
+    }
+
+    /// Get the error of the resolved endpoint
+    /// - Parameter allocator: The allocator to use for the error
+    public func getError() throws -> String? {
+        let errorOut = UnsafeMutablePointer<aws_byte_cursor>.allocate(capacity: 1)
```
waahm7, line 54: Same as above

Test/AwsCommonRuntimeKitTests/sdkutils/CRTAWSEndpointsRuleEngineTests.swift
```diff
@@ -0,0 +1,210 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+
+import XCTest
+import Foundation
+@testable import AwsCommonRuntimeKit
+
+class CRTAWSEndpointsRuleEngineTests: CrtXCBaseTestCase {
```
waahm7, line 8: We should be using tracing allocator in tests to detect memory leaks. However, this will fail the tests because currently Swift CRT has a lot of leaks.

```suggestion
//TODO: use tracing allocator
class CRTAWSEndpointsRuleEngineTests: CrtXCBaseTestCase {
```

waahm7 (changes requested):

Source/AwsCommonRuntimeKit/crt/CRTError.swift
```diff
@@ -9,4 +9,6 @@ public enum CRTError: Error {
   case memoryAllocationFailure
   case stringConversionError(UnsafePointer<aws_string>?)
   case crtError(AWSError)
+  // TODO: cleanup error handling https://github.com/awslabs/aws-crt-swift/issues/77
+  case awsError(AWSCommonRuntimeError)
```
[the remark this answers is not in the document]
ganeshnj: is this change merged? I don't think it was when I last checked.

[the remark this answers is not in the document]
waahm7: That changed is not merged to main, but we still should not be using AWSCommonRuntimeError class in new code. We can simply do 
`
   throw crtError(AWSError(errorCode: aws_last_error())`
   We can also add the following function in Aws Error.
`    public static func makeFromLastError() -> AWSError {
        return CRTError(errorCode: aws_last_error())
    }`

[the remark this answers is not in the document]
ganeshnj: as much I didn't want to step on this, but anyway I fixed it.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsRequestContext.swift
```diff
@@ -0,0 +1,55 @@
...
+
+    /// Initialize a new request context
+    /// - Parameter allocator: Allocator to use for request context creation
+    public init(allocator: Allocator = defaultAllocator) throws {
+        guard let rawValue = aws_endpoints_request_context_new(allocator.rawValue) else {
+            throw CRTError.awsError(AWSCommonRuntimeError())
+        }
+
+        self.rawValue = rawValue
+    }
```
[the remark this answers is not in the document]
ganeshnj: fixed.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsRequestContext.swift
```diff
@@ -0,0 +1,55 @@
...
+    /// - Parameters:
+    ///   - name: The name of the parameter
+    ///   - value: The value of the parameter
+    ///   - allocator: The allocator to use for the parameter
+    public func add(name: String, value: String?, allocator: Allocator = defaultAllocator) throws {
+        guard let value = value else {
+            return
+        }
+        let namePtr: UnsafeMutablePointer<aws_byte_cursor> = fromPointer(ptr: name.awsByteCursor)
+        let valuePtr: UnsafeMutablePointer<aws_byte_cursor> = fromPointer(ptr: value.awsByteCursor)
```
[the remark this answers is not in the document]
ganeshnj: No action as of now.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsRequestContext.swift
```diff
@@ -0,0 +1,55 @@
...
+
+    /// Add a string endpoint parameter to the request context
+    /// - Parameters:
+    ///   - name: The name of the parameter
+    ///   - value: The value of the parameter
+    ///   - allocator: The allocator to use for the parameter
+    public func add(name: String, value: String?, allocator: Allocator = defaultAllocator) throws {
+        guard let value = value else {
+            return
+        }
```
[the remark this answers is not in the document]
ganeshnj: because all endpoint params are nullable. They are discarded here because C APIs don't have nullable concept (yet).

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsRequestContext.swift
ganeshnj: fixed.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsRequestContext.swift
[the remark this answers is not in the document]
waahm7: I think it might be better to make value non nullable and the SDK only calls this functions when it has both name & value. Currently, this implementation makes it look like that we can successfully just add a name with no value which isn’t true. However, I am not sure how this fits into the big picture and if it is worth the effort to make this change.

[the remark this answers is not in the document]
ganeshnj: By the endpoints spec, params are nullable hence C must allow taking nullable values. This is safe guard for C code than SDK code.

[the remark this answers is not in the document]
ganeshnj: Also, this is by luck (actually rules are not exercising that path) here both `nil` and `not present` are working in C. @DmitriyMusatkin can share more information here.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsRequestContext.swift
ganeshnj: fixed.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
ganeshnj: fair call. fixed.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
ganeshnj: fixed.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsRuleEngine.swift
ganeshnj: fixed.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
ganeshnj: there is test case which I'm awaiting to be added in C which not only check for headers but also for properties.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
ganeshnj: fixed.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
ganeshnj: fixed.

Test/AwsCommonRuntimeKitTests/sdkutils/CRTAWSEndpointsRuleEngineTests.swift
ganeshnj: We are already using 
https://github.com/awslabs/aws-crt-swift/blob/main/Test/AwsCommonRuntimeKitTests/CrtXCBaseTestCase.swift#L8-L9

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsRequestContext.swift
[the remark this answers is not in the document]
DmitriyMusatkin: so to provide some context, this is basically a question about where crt draws a line for interface.

for ep2, a set of all possible parameters is defined for each service. some of those are required, some are optional. 

so basically, we are guaranteed that param name is never null (because we know all possible params at codegen time), but depending on type of request some parameter values might not be present (ex. bucket name on list buckets request). 

from sdk perspective set of parameters is basically constant, but some have null values. 

CRT currently asks to pass only the parameters that have concrete values (i.e. non null). This has some implementation benefits, but from interface perspective that means that caller needs to filter out all nulls. This does not affect correctness of eval in any way, and if CRT was to take nulls for param values, we would either make it no-op at the time of context construction or add null checks throughout eval (which i hate). 

anyways, discussed it with mike last week and we feel like its better design to not allow nullable types here. should we have a quick meeting next week, when mike is back to close out on this?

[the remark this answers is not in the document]
waahm7: Thanks, that makes sense.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSEndpointsResolvedEndpoint.swift
ganeshnj: test updated.

DmitriyMusatkin (approved):

waahm7 (approved):

DmitriyMusatkin (approved):

PR #84, opened 2022-10-11
Refactor Http Layer

waahm7:
*Description of changes:*
- Refactor Error handling to use only two classes
- Refactor cursor as previous implementation had memory leak issues
    - Nuke previous implementation
    - If we need a cursor for only one call, use closures. 
    - If we need a long lived cursor, copy string to aws_string, and get a cursor to aws_string.
- Refactor HTTP Layer 
    - Fix data classes to update underlaying C struct as well
    - Fix some strong reference cycles 
    - Fix clean up of resources 
    - Fix lifetime management issues of Connection Manager, Connection, and HttpStream
- Added some unit tests  
- Add some missing error checks and throws to functions
- Refactored some callback logic to use Unmanaged
- Nuked some unused code because best code is red code
- Added some todos (Please ignore them for now)


Source/AwsCommonRuntimeKit/crt/ByteCursor.swift
```diff
@@ -3,45 +3,13 @@
...
-        }
-    }
-}
-
 extension aws_byte_cursor {
-    public func toString() -> String? {
+    func toString() -> String? {
         if self.len == 0 { return nil }
-        return String(bytesNoCopy: self.ptr, length: self.len, encoding: .utf8, freeWhenDone: false)
+        return String(bytes: UnsafeBufferPointer(start: self.ptr, count: self.len), encoding: .utf8)
```
waahm7, line 9: BytesNoCopy is no longer supported. https://developer.apple.com/documentation/swift/string/init(bytesnocopy:length:encoding:freewhendone:)

waahm7:
CI is failing because of some missing const in C libraries. I will fix it before merge.

.github/workflows/ci.yml
```diff
@@ -1,13 +1,7 @@
 name: CI
 
 on:
-  push:
-    branches:
-      - '*'
-      - '!main'
   pull_request:
```
TingDaoK, line 4: why? I believe we don't want CI to run on every PR because of security issue. Please just use the same condition as we used other places. like https://github.com/awslabs/aws-c-common/blob/main/.github/workflows/ci.yml#L3-L6

waahm7: @TingDaoK this change is part of https://github.com/awslabs/aws-crt-swift/pull/80.

.github/workflows/ci.yml
```diff
@@ -75,4 +69,4 @@ jobs:
         run: |
           python3 -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz?run=${{ env.RUN }}', 'builder')"
           chmod a+x builder
-          ./builder build -p ${{ env.PACKAGE_NAME }} --target=${{ matrix.target }}-armv8
```
waahm7, line 78: This is from #80

Source/AwsCommonRuntimeKit/io/TlsContextOptions.swift
```diff
@@ -2,86 +2,72 @@
...
-    
-    public init(clientWithMtlsCert cert: inout ByteCursor,
-                key: inout ByteCursor,
-                allocator: Allocator = defaultAllocator) throws {
-        self.rawValue = allocatePointer()
-        if aws_tls_ctx_options_init_client_mtls_pkcs12(rawValue,
-                                                       allocator.rawValue,
-                                                       &cert.rawValue,
-                                                       &key.rawValue) != AWS_OP_SUCCESS {
-            throw CRTError(errorCode: aws_last_error())
```
waahm7, line 39: Nuke this function as we don't want to expose byte cursor. Customers should call the one which takes cert as a String

Source/AwsCommonRuntimeKit/io/TlsContextOptions.swift
```diff
@@ -2,86 +2,72 @@
...
 import AwsCIo
-//swiftlint:disable trailing_whitespace
 public final class TlsContextOptions {
+    private let allocator: Allocator
     var rawValue: UnsafeMutablePointer<aws_tls_ctx_options>
-    
     public static func isAlpnSupported() -> Bool {
         return aws_tls_is_alpn_available()
     }
-
```
waahm7, line 12: Even if I put a tab here to remove this change, git does not detect that change. I think the IDE just reverts it automatically.

Test/AwsCommonRuntimeKitTests/http/HttpTests.swift
```diff
@@ -0,0 +1,145 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+import XCTest
+@testable import AwsCommonRuntimeKit
+import AwsCCommon
+
+class HttpTests: CrtXCBaseTestCase {
```
waahm7, line 7: This is a first iteration of getting a simple http get request working with memory & lifetime issues fixed. This also helps to visualize how users can make a simple get http request. I will improve these tests coverage (other type of requests) and make helper functions more generic in a future PR. 

There are two tests for two implementations of makeRequest.

Source/AwsCommonRuntimeKit/http/CRTHttpResponse.swift
```diff
@@ -4,9 +4,6 @@ import AwsCHttp
 import AwsCIo
 
 public final class CRTHttpResponse: HttpMessage {
```
waahm7, line 6: We can probably just remove this class. We can get the response status code from HttpStream class.

Source/AwsCommonRuntimeKit/AwsCommonRuntimeKit.swift
```diff
@@ -4,6 +4,7 @@ import AwsCAuth
 public struct AwsCommonRuntimeKit {
   public static let version = "@AWS_CRT_SWIFT_VERSION@"
 
+  // Todo: which library to init? Auth does init of http
```
graebm, line 7: it's safe to init them all. They protect against repeated init

Source/AwsCommonRuntimeKit/Utilities.swift
```diff
@@ -12,43 +12,48 @@ func zeroStruct<T>(_ ptr: UnsafeMutablePointer<T>) {
...
-
-        guard AWS_OP_SUCCESS == aws_md5_compute(allocator.rawValue, input, output, truncate) else {
+        var buffer = aws_byte_buf()
+        if aws_byte_buf_init(&buffer, allocator.rawValue, 16) != AWS_OP_SUCCESS {
             return nil
         }
+        guard self.withByteCursorPointer({ strCursorPointer in
+           aws_md5_compute(allocator.rawValue, strCursorPointer, &buffer, truncate)
+        }) == AWS_OP_SUCCESS else {
+            return nil
```
graebm, line 25: fail to clean up the aws_byte_buf if we fail here.
maybe do something like:
- use a `defer` block to ensure the byte_buf is cleaned up, regardless of where this function returns from
OR
- Create an internal helper `class ByteBuf` the calls clean up from its `deinit`

Source/AwsCommonRuntimeKit/Utilities.swift
```diff
@@ -12,43 +12,48 @@ func zeroStruct<T>(_ ptr: UnsafeMutablePointer<T>) {
...
+        }
+        return Data(bytesNoCopy: buffer.buffer, count: buffer.len, deallocator: .custom { _, _ in
+            aws_byte_buf_clean_up(&buffer)
+        }).base64EncodedString()
```
graebm, line 29: just copy the 16 bytes. it's not worth the complexity of nocopy

Source/AwsCommonRuntimeKit/Utilities.swift
graebm: or use the "no alloc" variant of aws_byte_buf:
```
var buffer_data = Array<UInt8>(repeating: 0, count: 16)
var buffer = aws_byte_buf_from_empty_array(....)
```

Source/AwsCommonRuntimeKit/http/HttpRequest.swift
```diff
@@ -2,18 +2,15 @@
 //  SPDX-License-Identifier: Apache-2.0.
 import AwsCHttp
 import AwsCIo
-
+import AwsCCommon
 public final class HttpRequest: HttpMessage {
```
graebm, line 6: utterly trivial: it's nice to have a line of whitespace between the imports and class declaration

Source/AwsCommonRuntimeKit/crt/AWSString.swift
```diff
@@ -3,38 +3,31 @@
...
+// TODO: try to find a better way to get a long lived byte cursor from String without making a copy.
+class AWSStringByteCursor {
+    let awsString: AWSString
+    let string: String
+    let byteCursor: aws_byte_cursor
 
-    init(_ awsString: AWSString) {
-        self.awsString = awsString
-        self.rawValue = aws_byte_cursor_from_string(awsString.rawValue)
+    init(_ string: String, allocator: Allocator = defaultAllocator) {
```
graebm, line 27: for internal classes, we SHOULDN'T provide a default allocator.
if some public class needs to use an internal class, it should be forced to pass along its allocator

If addressing this is a huge revamp, just mark it with TODO

Personally, I'm not sure it's worth supporting per-object custom allocators AT ALL for aws-crt-swift. It's just not a thing that's done in the Swift ecosystem.

Source/AwsCommonRuntimeKit/http/HttpClientConnection.swift
```diff
@@ -23,102 +26,24 @@ public class HttpClientConnection {
...
 
     /// Close the http connection
-    public func close() {
-        manager.releaseConnection(connection: self)
-    }
+// TODO: do we need a explicit close function or deinit is enough?
+//    public func close() throws
+//        try manager.releaseConnection(connection: self)
+//        manager = nil
+//    }
```
graebm, line 33: Yeah, it's common enough that people want to kill a connection because something is going wrong, but getting its refcount down to zero is a complicated process.

but ughghhgh I'm not sure what's going on with this design where "close" really means "give it back to the connection manager for re-use".

In C, the HttpClientConnection is ignorant of the HttpClientConnectionManager, but here in Swift we've tied their implementation together.

I really need to look this through 🤦‍♀️

Source/AwsCommonRuntimeKit/crt/AWSString.swift
waahm7: Makes sense, thanks.

Source/AwsCommonRuntimeKit/auth/imds/CRTIMDSClient.swift
```diff
@@ -2,28 +2,33 @@
...
     ///
     /// - Parameters:
     ///    - resourcePath: `String` path of the resource to query
     public func getResource(resourcePath: String) async throws -> String? {
+        let resourcePathCursor = AWSStringByteCursor(resourcePath)
         return try await withCheckedThrowingContinuation { (continuation: ResourceContinuation) in
             let callbackData = CRTIMDSClientResourceCallbackData(continuation: continuation)
             let pointer: UnsafeMutableRawPointer = fromPointer(ptr: callbackData)
-            aws_imds_client_get_resource_async(rawValue, resourcePath.awsByteCursor, resourceCallback, pointer)
+            aws_imds_client_get_resource_async(rawValue, resourcePathCursor.byteCursor, resourceCallback, pointer)
```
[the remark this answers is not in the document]
waahm7: The pointer is invalid as soon as we leave the closure. So that doesn’t work in async logic where we need a manually managed byte cursor. We can not make a copy because that copy will also be invalid. Here, in particular, I think we can use a closure because of the Continuation which will wait until the callback has fired. However, [SigningConfig](https://github.com/awslabs/aws-crt-swift/blob/b4c6f28a32568c101fa7dc801d5d223b935b7fd2/Source/AwsCommonRuntimeKit/auth/signing/SigningConfig.swift ) is a good example where we need a persistent byte cursor and we cannot use a closure. 

That’s where I ran into the issue of undefined behavior because byte cursor was stored by C struct which was not valid after the init call. We probably can get rid of it after some major refactoring of the signing logic and the way code is organized. I will try to do that in the Auth layer refactoring PR.

Source/AwsCommonRuntimeKit/crt/AWSString.swift
waahm7: Thanks, added a todo. I guess it’s nice to have it for the tests where we it can help detect memory leaks. We can also maybe just override the defaultAllocator with tracing allocator for tests.

Source/AwsCommonRuntimeKit/crt/AWSString.swift
```diff
@@ -3,38 +3,31 @@
...
     }
 }
 
-private struct AWSStringByteCursor: ByteCursor {
-    private let awsString: AWSString
-    public var rawValue: aws_byte_cursor
+// TODO: try to find a better way to get a long lived byte cursor from String without making a copy.
+class AWSStringByteCursor {
+    let awsString: AWSString
+    let string: String
```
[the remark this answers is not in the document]
waahm7: Storing String doesn’t occur any cost because Swift doesn’t make a copy until it is mutated. I needed it in the HttpProxyOption class because of complexity of didSet. I will get rid of it after the revamp of withCStruct or getCStruct.

Source/AwsCommonRuntimeKit/auth/imds/CRTIMDSClient.swift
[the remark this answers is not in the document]
waahm7: HttpProxyOptions have a similar issue.

Source/AwsCommonRuntimeKit/crt/AWSString.swift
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/http/HttpProxyOptions.swift
```diff
@@ -1,23 +1,88 @@
...
-
-    public init(hostName: String, port: UInt16) {
-        self.rawValue = allocatePointer()
-        self.hostName = hostName
+
+    public var authType: HttpProxyAuthenticationType = .none {
+        didSet {
+            rawValue.pointee.auth_type = authType.rawValue
+        }
+    }
```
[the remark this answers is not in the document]
waahm7: Thanks, I have improved it.

Source/AwsCommonRuntimeKit/Utilities.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/http/HttpClientConnection.swift
waahm7: Thanks, resolved.

Source/AwsCommonRuntimeKit/auth/imds/CRTIMDSClient.swift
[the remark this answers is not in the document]
waahm7: Thanks, will resolve in a new PR.

Source/AwsCommonRuntimeKit/AwsCommonRuntimeKit.swift
waahm7: Thanks, will fix it in a future PR. Just doing init of http & auth is confusing.

Source/AwsCommonRuntimeKit/auth/signing/SigningConfig.swift
```diff
@@ -32,11 +36,15 @@ public struct SigningConfig {
...
-                configType: SigningConfigType = .aws) {
+                configType: SigningConfigType = .aws,
+                allocator: Allocator = defaultAllocator) {
+
         self.credentials = credentials
         self.credentialsProvider = credentialsProvider
         self.expiration = expiration
         self.date = date
+        self._service = AWSStringByteCursor(service, allocator: allocator)
+        self._region = AWSStringByteCursor(region, allocator: allocator)
```
waahm7, line 47: This will be fixed by Fix CStruct PR. Will most probably nuke the AWSStringByteCursor class.

Source/AwsCommonRuntimeKit/crt/CommandLine.swift
```diff
@@ -3,7 +3,8 @@
 
 import Foundation
 import AwsCCommon
-//swiftlint:disable trailing_whitespace
+// swiftlint:disable trailing_whitespace
+// Todo: remove/refactor
 public struct CommandLineParser {
```
waahm7, line 8: Will refactor it in future to use https://apple.github.io/swift-argument-parser/documentation/argumentparser/.

Source/AwsCommonRuntimeKit/http/HttpClientConnectionManager.swift
```diff
@@ -6,104 +6,72 @@ import Collections
...
-    let manager: OpaquePointer
-    let allocator: Allocator
+    let rawValue: OpaquePointer
     let options: HttpClientConnectionOptions
 
-    public init(options: HttpClientConnectionOptions, allocator: Allocator = defaultAllocator) {
+    public init(options: HttpClientConnectionOptions, allocator: Allocator = defaultAllocator) throws {
         self.options = options
-        self.allocator = allocator
+        // Todo: fix shutdown options
```
waahm7, line 14: Fixed by #85

Source/AwsCommonRuntimeKit/auth/signing/SigningConfig.swift
```diff
@@ -11,9 +12,12 @@ public struct SigningConfig {
     public let expiration: Int64
     public let signedBodyHeader: SignedBodyHeaderType
     public let signedBodyValue: SignedBodyValue
+    let signedBodyValueCursor: AWSStringByteCursor
```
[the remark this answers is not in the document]
waahm7: Yes, thanks.

Source/AwsCommonRuntimeKit/http/HttpStreamCalbackCore.swift
```diff
@@ -0,0 +1,111 @@
...
+
+func onComplete(stream: UnsafeMutablePointer<aws_http_stream>?,
+                errorCode: Int32,
+                userData: UnsafeMutableRawPointer!) {
+
+    let httpStreamCbData = Unmanaged<HttpStreamCallbackCore>.fromOpaque(userData).takeUnretainedValue()
+    let stream = httpStreamCbData.stream!
+    let onStreamCompleteFn = httpStreamCbData.requestOptions.onStreamComplete
+    onStreamCompleteFn(stream, CRTError(code: errorCode))
+    httpStreamCbData.stream = nil
```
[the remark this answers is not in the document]
waahm7: 🥳

Source/AwsCommonRuntimeKit/http/HttpStream.swift
```diff
@@ -3,19 +3,20 @@
 import AwsCHttp
 
 public class HttpStream {
-    var httpStream: UnsafeMutablePointer<aws_http_stream>?
+    let rawValue: UnsafeMutablePointer<aws_http_stream>
+    var callbackData: HttpStreamCallbackCore
+    private var activated: Bool = false
```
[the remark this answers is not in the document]
waahm7: Fixed, thanks

Source/AwsCommonRuntimeKit/io/TlsConnectionOptions.swift
```diff
@@ -9,34 +9,33 @@ public final class TlsConnectionOptions {
...
 	public func setServerName(_ serverName: String) throws {
-		var byteCur = serverName.newByteCursor()
-		if aws_tls_connection_options_set_server_name(rawValue,
-                                                      self.allocator.rawValue,
-                                                      &byteCur.rawValue) != AWS_OP_SUCCESS {
-			throw CRTError(errorCode: aws_last_error())
+		if (serverName.withByteCursorPointer { serverNameCursorPointer in
+			aws_tls_connection_options_set_server_name(rawValue,
+                                                       self.allocator.rawValue,
+					                                   serverNameCursorPointer)
```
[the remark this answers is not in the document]
waahm7: Git issue, seems fine in IDE.

Source/AwsCommonRuntimeKit/http/HttpStream.swift
```diff
@@ -26,15 +27,32 @@ public class HttpStream {
...
+        }
+        return Int(status)
     }
 
-    ///Activates the client stream.
-    public func activate() {
-        aws_http_stream_activate(httpStream)
+    /// Activates the client stream.
+    /// Multiple calls to activate have no effect if the stream is already activated
+    /// This is thread safe because HttpStream is an actor
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/http/HttpClientConnection.swift
```diff
@@ -23,102 +26,27 @@ public class HttpClientConnection {
     }
 
     /// Close the http connection
-    public func close() {
-        manager.releaseConnection(connection: self)
+    public func close() throws {
+        aws_http_connection_close(rawValue)
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

graebm (approved):

PR #87, opened 2022-10-24
IO Refactor

waahm7:
*Description of changes:*
- Fixes Retry Strategy which had weird double implementation.
- Fixes Stream class. Previous implementation had undefined behavior and was exposing internal CRT stuff.
- Use core classes whereever possible.


Source/AwsCommonRuntimeKit/io/StreamCore.swift
```diff
@@ -0,0 +1,93 @@
...
+
+    typealias RawType = aws_input_stream_vtable
+    func withCStruct<Result>(_ body: (aws_input_stream_vtable) -> Result) -> Result {
+        return body(_rawValue.pointee)
+    }
+
+    deinit {
+        allocator.release(_rawValue)
+    }
+}
```
waahm7, line 46: I don’t like this, however, I can’t think of a better way. Previous implementation had undefined behavior as you are not supposed to store pointer by &.

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -4,40 +4,36 @@
...
 
-// swiftlint:disable trailing_whitespace
+public protocol AwsStream {
+    var isEndOfStream: Bool { get }
+    var length: UInt { get }
+
+    func seek(offset: Int64, streamSeekType: StreamSeekType) -> Bool
+    /// Data.count should not greater than length.
+    func read(length: Int) -> Data
+}
```
waahm7, line 14: As this protocol is public, we should not be exposing internals of CRT.

Source/AwsCommonRuntimeKit/io/ClientBootstrap.swift
```diff
@@ -1,33 +1,32 @@
...
 
+    public convenience init(eventLoopGroup: EventLoopGroup,
+                            hostResolver: HostResolver,
+                            allocator: Allocator = defaultAllocator,
+                            shutdownCallback: ShutdownCallback? = nil) throws {
+        try self.init(clientBootstrapOptions: ClientBootstrapOptions(eventLoopGroup: eventLoopGroup,
+                                                                     hostResolver: hostResolver,
+                                                                     shutdownCallback: shutdownCallback),
+                      allocator: allocator)
+    }
```
waahm7, line 28: We have a couple of options here

- Nuke convenience init and just have init from clientBootstrapOptions
- Nuke clientBootstrapOptions and take elg, hostResolver etc directly like we are doing it in HostResolver.
- Keep both.

 I like both but it is not consistent with other classes.

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -4,40 +4,36 @@
...
-    func seek(offset: Int64, basis: aws_stream_seek_basis) -> Bool
-    func read(buffer: inout aws_byte_buf) -> Bool
-}
-
 extension FileHandle: AwsStream {
     @inlinable
-    public var status: aws_stream_status {
-        return aws_stream_status(is_end_of_stream: self.length == self.offsetInFile, is_valid: true)
+    public var isEndOfStream: Bool {
+        self.length == self.offsetInFile
```
graebm, line 36: under the hood `.length` is seeking to the end of the stream, that's an expensive isEndOfStream check 

in our other language bindings, in the read() functions, when we hit the end of the file we set some member variable like `self.endOfFile = true` ... and then later in the aws_input_stream_get_status() functions we report that we've hit the end

I see how you can't do that, due to this being a protocol and you can't add a member variable

maybe we need to update the protocol
remove `isEndOfStream`
add some way for `read()` to report when it's hit the end of the file, and we'll store that fact in InputStreamCore so we can report it later when `aws_input_stream_get_status()` is called

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -49,61 +45,18 @@ extension FileHandle: AwsStream {
     }
```
graebm, line 45: above this, seekToEndOfFile() is used, but that's a deprecated function because it doesn't report errors, we should use seekToEnd()

likewise seek(toFileOffset) is deprecated and seek(toOffset) should be used so we don't crash the application on errors

which means the lenght() protocol function should be throwing

...

and now that I'm thinking about it since a failed seek can REALLY mess up the stream, maybe we should just have InputStreamCore store `self.valid = false` any time one of its protocol functions throw, and we can report that via `aws_input_stream_get_status()`

Source/AwsCommonRuntimeKit/io/HostAddress.swift
```diff
@@ -4,13 +4,13 @@
...
-    let useCount: Int
-    let connectionFailureCount: Int
-    let weight: UInt8
+    public var host: String?
+    public var address: String?
+    public var recordType: AddressRecordType
+    public var expiry: UInt64
+    public var useCount: Int
+    public var connectionFailureCount: Int
+    public var weight: UInt8
```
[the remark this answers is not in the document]
waahm7: Thanks, This is more like the public API for host address. Internally, we are using the C struct directly. So I have nuked these variables.

Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -4,98 +4,48 @@
...
-        )
-
-        self.config = fromPointer(ptr: config)
-
-    }
-
-    deinit {
-        config.deinitializeAndDeallocate()
-        aws_host_resolver_release(rawValue)
+        self.config = HostResolutionConfig(maxTTL: maxTTL)
```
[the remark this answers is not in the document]
waahm7: Thanks, we need the aws_host_resolution_config at two places (resolve & client bootstrap init). So I have created a function to create aws_host_resolution_config.

Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -4,98 +4,48 @@
...
     }
 
     public func resolve(host: String) async throws -> [HostAddress] {
-         return try await withCheckedThrowingContinuation({ (continuation: HostResolvedContinuation) in
-             resolve(host: host, continuation: continuation)
+        return try await withCheckedThrowingContinuation({ (continuation: HostResolvedContinuation) in
+            HostResolverCore(hostResolver: self, host: host, continuation: continuation, allocator: allocator).retainedResolve()
```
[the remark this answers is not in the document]
waahm7: Thanks, I have renamed it to HostResolveCore. In my head the "Core" naming is for anything that has manual memory management. We can just use a closure but we will still need to balance the retain & release calls. We can create a new naming convention if we want to distinguish between something that outlives a class vs something that keeps the class alive till the duration of an async operation. Let's discuss more on this on Monday.

Source/AwsCommonRuntimeKit/io/HostResolverCore.swift
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    private func getRetainedSelf() -> UnsafeMutableRawPointer {
+        return Unmanaged<HostResolverCore>.passRetained(self).toOpaque()
+    }
+
+    /// This function does a manual retain on HostResolverCore
+    /// to keep it until until onHostResolved callback has fired which will do the release.
+    func retainedResolve() {
+        let retainedSelf = getRetainedSelf()
```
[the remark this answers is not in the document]
waahm7: Closures that are passed to C can not have a context. So we need to pass Self manually if we want to have a context in the callback.

Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -4,98 +4,48 @@
...
-    func resolve(host: String) async throws -> [HostAddress]
-}
-
-public final class DefaultHostResolver: HostResolver {
-    public var rawValue: UnsafeMutablePointer<aws_host_resolver>
-    public var config: UnsafeMutablePointer<aws_host_resolution_config>
+public class HostResolver {
+    let rawValue: UnsafeMutablePointer<aws_host_resolver>
+    let config: HostResolutionConfig
     private let allocator: Allocator
```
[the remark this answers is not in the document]
waahm7: It is used in the resolve call where we need to convert host:String -> AwsString

Source/AwsCommonRuntimeKit/io/HostResolverCore.swift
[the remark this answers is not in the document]
waahm7: https://github.com/apple/swift/blob/main/docs/HowSwiftImportsCAPIs.md#function-pointers

Source/AwsCommonRuntimeKit/io/Stream.swift
waahm7: I am not sure if Python binding has a bug. We are doing 
```
if (bytes_read == 0) {
    impl->is_end_of_stream = true;
} else {
    dest->len += bytes_read;
}
```
Let say stream has a length of 10. In the first read, we read 10 bytes. IsEnd should be true but it will be false until we do another read. 
I am thinking out about replacing isEndOfStream with currentOffset. and we can do a simple currentOffset == (pre cached) length check. Also can you please let me know what makes seeking to the end of a file expensive? I think, it should be O(1).

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -4,40 +4,36 @@
 import AwsCIo
 import Foundation
 
-// swiftlint:disable trailing_whitespace
+public protocol AwsStream {
+    var isEndOfStream: Bool { get }
+    var length: UInt { get }
+
+    func seek(offset: Int64, streamSeekType: StreamSeekType) -> Bool
```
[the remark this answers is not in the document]
waahm7: Thanks, it is throwable now.

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -4,40 +4,36 @@
...
+    /// Data.count should not greater than length.
+    func read(length: Int) -> Data
```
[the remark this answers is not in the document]
waahm7: Thanks, I have updated the API.

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -49,61 +45,18 @@ extension FileHandle: AwsStream {
...
     @inlinable
-    public func seek(offset: Int64, basis: aws_stream_seek_basis) -> Bool {
+    public func seek(offset: Int64, streamSeekType: StreamSeekType) -> Bool {
         let targetOffset: UInt64
-        if basis.rawValue == AWS_SSB_BEGIN.rawValue {
-            targetOffset = self.offsetInFile + UInt64(offset)
-        } else {
-            targetOffset = self.offsetInFile - UInt64(offset)
+        switch streamSeekType {
+        case .begin: targetOffset = self.offsetInFile + UInt64(offset)
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -49,61 +45,18 @@ extension FileHandle: AwsStream {
...
+        case .end: targetOffset = self.offsetInFile - UInt64(offset)
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -49,61 +45,18 @@ extension FileHandle: AwsStream {
...
         }
         self.seek(toFileOffset: targetOffset)
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/io/StreamCore.swift
```diff
@@ -0,0 +1,83 @@
...
+
+private func doRead(_ stream: UnsafeMutablePointer<aws_input_stream>!,
+                    _ buffer: UnsafeMutablePointer<aws_byte_buf>!) -> Int32 {
+    let awsStream = Unmanaged<AwsInputStreamCore>.fromOpaque(stream.pointee.impl).takeUnretainedValue().awsStream
+    let length = buffer.pointee.capacity
+    let data = awsStream.read(length: length)
+
+    // Invalid data length
+    if data.count > length || (data.count == 0 && !awsStream.isEndOfStream) {
+        return aws_raise_error(Int32(AWS_IO_STREAM_READ_FAILED.rawValue))
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -4,40 +4,36 @@
 import AwsCIo
 import Foundation
 
-// swiftlint:disable trailing_whitespace
+public protocol AwsStream {
+    var isEndOfStream: Bool { get }
+    var length: UInt { get }
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -4,40 +4,36 @@
 import AwsCIo
 import Foundation
 
-// swiftlint:disable trailing_whitespace
+public protocol AwsStream {
+    var isEndOfStream: Bool { get }
```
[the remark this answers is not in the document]
waahm7: Thanks, nuked this function.

Source/AwsCommonRuntimeKit/io/Stream.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/io/HostResolveCore.swift
```diff
@@ -0,0 +1,70 @@
...
+    let length = aws_array_list_length(hostAddresses!)
+    var addresses = [HostAddress]()
+
+    for index in 0..<length {
+        var address: UnsafeMutableRawPointer! = nil
+        aws_array_list_get_at_ptr(hostAddresses!, &address, index)
+        let hostAddressCType = address.bindMemory(to: aws_host_address.self, capacity: 1).pointee
+        if let hostAddress = HostAddress(hostAddress: hostAddressCType) {
+            addresses.append(hostAddress)
+        }
```
waahm7, line 66: Is it possible that some host address doesn't have a value for host or address? Should we crash if that happens?

Source/AwsCommonRuntimeKit/io/StreamCore.swift
```diff
@@ -0,0 +1,83 @@
...
+}
+
+private func doSeek(_ stream: UnsafeMutablePointer<aws_input_stream>!,
+                    _ offset: Int64,
+                    _ seekBasis: aws_stream_seek_basis) -> Int32 {
+    let inputStream = Unmanaged<AwsInputStreamCore>.fromOpaque(stream.pointee.impl).takeUnretainedValue()
+    if inputStream.awsStream.seek(offset: offset, streamSeekType: StreamSeekType(rawValue: seekBasis.rawValue)!) {
+        return AWS_OP_SUCCESS
+    }
+    return aws_raise_error(Int32(AWS_IO_STREAM_INVALID_SEEK_POSITION.rawValue))
```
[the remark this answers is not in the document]
waahm7: Thanks, I have improved it.

Source/AwsCommonRuntimeKit/io/HostResolveCore.swift
graebm: yeah, that would be freaky and unexpected
I agree, this is a good reason to crash

graebm: yeah, I just looked into the C code ... they're both always present

graebm (commented):
still working through it ... but here's what I had at the end of day

Source/AwsCommonRuntimeKit/io/HostAddress.swift
```diff
@@ -4,31 +4,27 @@
...
 
-    init(hostAddress: aws_host_address) {
-        self.host = String(awsString: hostAddress.host)
-        self.address = String(awsString: hostAddress.address)
+    init?(hostAddress: aws_host_address) {
+        guard let host = String(awsString: hostAddress.host) else {
+            return nil
+        }
+        guard let address = String(awsString: hostAddress.address) else {
+            return nil
```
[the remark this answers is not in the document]
waahm7: Updated, thanks.

Source/AwsCommonRuntimeKit/io/Stream.swift
waahm7: Thank you for the detailed explanation. I have already updated this after discussion in the office.

Source/AwsCommonRuntimeKit/io/HostResolveCore.swift
```diff
@@ -0,0 +1,70 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+import AwsCIo
+
+typealias HostResolvedContinuation = CheckedContinuation<[HostAddress], Error>
+
+/// Core classes have manual memory management.
+/// You have to balance the retain & release calls in all cases to avoid leaking memory.
+class HostResolveCore {
+    let continuation: HostResolvedContinuation
```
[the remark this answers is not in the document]
waahm7: I tried it, and it crashes during debugging due to accessing invalid address access. If I run it, it works, but, logs have the following error: “SWIFT TASK CONTINUATION MISUSE: resolve(host:) leaked its continuation!”. Normally, closure have context, so you do not have to do anything manual and access continuation through the context.

In our case, it is probably safer to just do a manual retain & release. We have to pass this as a pointer to C and Swift preferred approach is to not store pointers.

Some intresting discussion : https://forums.swift.org/t/swift-function-as-a-callback-to-a-c-function/37409/9

Source/AwsCommonRuntimeKit/io/HostResolveCore.swift
```diff
@@ -0,0 +1,70 @@
...
+        let hostStr = AWSString(host, allocator: allocator)
+        let retainedSelf = core.getRetainedSelf()
+        withUnsafePointer(to: hostResolver.getHostResolutionConfig()) { hostResolutionConfigPointer in
+            if aws_host_resolver_resolve_host(hostResolver.rawValue,
+                    hostStr.rawValue,
+                    onHostResolved,
+                    hostResolutionConfigPointer,
+                    retainedSelf) != AWS_OP_SUCCESS {
+                //TODO: this is wrong. Sometimes it triggers the error callback and sometimes it doesn't.
+                // I have a fix in progress in aws-c-io
```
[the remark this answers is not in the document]
waahm7: Not yet, I have updated the PR with a TODO of async callback. I might spend some more time and get it merged afterwards.

Source/AwsCommonRuntimeKit/io/HostResolveCore.swift
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /// This function does a manual retain on HostResolverCore
+    /// to keep it until until onHostResolved callback has fired which will do the release.
+    static func retainedResolve(hostResolver: HostResolver, host: String, continuation: HostResolvedContinuation, allocator: Allocator) {
+        let core = HostResolveCore(continuation)
+        let hostStr = AWSString(host, allocator: allocator)
+        let retainedSelf = core.getRetainedSelf()
+        withUnsafePointer(to: hostResolver.getHostResolutionConfig()) { hostResolutionConfigPointer in
+            if aws_host_resolver_resolve_host(hostResolver.rawValue,
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -71,8 +71,8 @@ public class ByteBuffer: Codable {
         return self
     }
 
-    public func put(_ value: ByteBuffer, offset: UInt = 0, maxBytes: UInt? = nil) {
-        var end: UInt = UInt(value.length)
+    public func put(_ value: ByteBuffer, offset: UInt = 0, maxBytes: UInt? = nil) throws {
+        var end: UInt = try UInt(value.length())
```
graebm, line 75: don't make this function `throws`

ByteBuffer.length() cannot ever actually throw
the protocol API can throw
but the ByteBuffer's implementation never does
just do `!` that it succeeds

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -273,40 +273,25 @@ public class ByteBuffer: Codable {
...
-        if basis.rawValue == AWS_SSB_BEGIN.rawValue {
-            targetOffset = offset
-
-        } else {
-            targetOffset = Int64(length) - offset
-        }
-        currentIndex = Int(targetOffset)
-        return true
+    public func seek(offset: UInt64) throws {
+        currentIndex = Int(offset)
```
graebm, line 283: check that currentIndex doesn't exceed capacity

(oh yeah, just remembered, you were doing that safety check in the wrapper function, but I don't think you should, for efficiency's sake. Sorry about getting so hung up on efficiency on the aws_input_stream implementation, but this is is literally where our programs spend most of their time)

Source/AwsCommonRuntimeKit/crt/ContinuationCore.swift
```diff
@@ -0,0 +1,20 @@
...
+    init(continuation: CheckedContinuation<T, Error>) {
+        self.continuation = continuation
+    }
+
+    func passRetained() -> UnsafeMutableRawPointer {
+        return Unmanaged<ContinuationCore>.passRetained(self).toOpaque()
+    }
+
+    func release() {
+        Unmanaged.passUnretained(self).release()
```
graebm, line 18: utterly trivial: seems like typing `<ContinuationCore>` isn't necessary? since you only typed it in one and not the other?

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -4,106 +4,80 @@
...
-                    _ offset: Int64,
-                    _ seekBasis: aws_stream_seek_basis) -> Int32 {
-    let inputStream = Unmanaged<AwsInputStream>.fromOpaque(stream.pointee.impl).takeUnretainedValue()
-    if inputStream.awsStream.seek(offset: offset, basis: seekBasis) {
-        return AWS_OP_SUCCESS
-    }
-    return AWS_OP_ERR
-}
+        guard let baseAddress = buffer.baseAddress else {
+            throw CommonRunTimeError.crtError(CRTError(code: Int32(AWS_IO_STREAM_READ_FAILED.rawValue)))
```
graebm, line 75: just crash if `buffer.baseAddress` isn't set

would it be simpler if our `read()` API was more like [copyBytes(to:count:)](https://developer.apple.com/documentation/foundation/data/1780297-copybytes)? taking 2 separate pointer and count args? right now I see you warning users not to change `buffer.count` and then you have to check here that `buffer.baseAddress` is actually set.

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -4,106 +4,80 @@
...
-        awsStream = impl
-        rawValue = aws_input_stream()
-        rawValue.vtable = UnsafePointer<aws_input_stream_vtable>(&vtable)
-        rawValue.impl = Unmanaged<AwsInputStream>.passUnretained(self).toOpaque()
-    }
+    /// (Optional) throws Seek not supported error by default.
+    func seek(offset: UInt64) throws
+
+    /// buffer count should not be modified.
+    func read(buffer: UnsafeMutableBufferPointer<UInt8>) throws -> Int
```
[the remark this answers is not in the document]
waahm7: Thanks, FileHandle.read() returns nil when end of file. I will do something similar.

Source/AwsCommonRuntimeKit/io/StreamCore.swift
```diff
@@ -0,0 +1,102 @@
...
+            if offset > 0 {
+                return aws_raise_error(Int32(AWS_IO_STREAM_INVALID_SEEK_POSITION.rawValue))
+            }
+            targetOffset = length - UInt64(abs(offset))
+        }
+        try iStreamable.seek(offset: targetOffset)
+        iStreamCore.isEndOfStream = false
+        return AWS_OP_SUCCESS
+    } catch {
+        return aws_raise_error(Int32(AWS_IO_STREAM_SEEK_FAILED.rawValue))
```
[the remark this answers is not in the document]
waahm7: Good idea, I have done something similar in delegate credential provider. I will implement that here as well.

Source/AwsCommonRuntimeKit/io/retryer/CRTAWSRetryStrategy.swift
```diff
@@ -3,113 +3,90 @@
...
+            return aws_retry_strategy_new_standard(allocator.rawValue, retryOptionsPointer)
+        }) else {
             throw CommonRunTimeError.crtError(.makeFromLastError())
         }
-
-        self.init(retryStrategy: retryer, allocator: allocator)
+        self.rawValue = rawValue
     }
 
     public func acquireToken(timeout: UInt64 = 0, partitionId: String) async throws -> CRTAWSRetryToken {
```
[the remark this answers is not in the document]
waahm7: Thanks, this PR is supposed to refactor the IO layer so I have updated it to a Time Interval which has seconds by default. However, I think this parameter is not being used at all in C. Let’s discuss it further tomorrow.

Source/AwsCommonRuntimeKit/io/Stream.swift
waahm7: Thanks, fixed by https://developer.apple.com/documentation/foundation/data/1779913-copybytes.

Source/AwsCommonRuntimeKit/crt/ContinuationCore.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
waahm7: Thanks, I found an interesting tidbit that we can satisfy a throwing function protocol requirement with a non throwing function.

Source/AwsCommonRuntimeKit/io/retryer/CRTStandardRetryOptions.swift
```diff
@@ -0,0 +1,24 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+import AwsCIo
+
+public struct CRTStandardRetryOptions: CStruct {
```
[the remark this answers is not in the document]
waahm7: Good catch, fixed.

Source/AwsCommonRuntimeKit/io/retryer/CRTStandardRetryOptions.swift
```diff
@@ -0,0 +1,24 @@
...
+    /// capacity for partitions. Defaults to 500
+    public var initialBucketCapacity: Int
+    public var exponentialBackoffRetryOptions: CRTExponentialBackoffRetryOptions
+
+    public init(initialBucketCapacity: Int = 500, exponentialBackoffRetryOptions: CRTExponentialBackoffRetryOptions) {
```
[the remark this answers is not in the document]
waahm7: Thanks, It had a default value indirectly. It's fixed by not using an option struct.

Source/AwsCommonRuntimeKit/crt/CommonRuntimeError.swift
```diff
@@ -1,6 +1,27 @@
...
 // SPDX-License-Identifier: Apache-2.0.
 
+import AwsCCommon
+
 public enum CommonRunTimeError: Error {
     case crtError(CRTError)
 }
+
+public struct CRTError {
+
```
waahm7, line 11: CRTError no longer conforms to Error protocol. We should never throw CRTError directly and it just complicates catching the error.

Source/AwsCommonRuntimeKit/io/HostResolveCore.swift
```diff
@@ -0,0 +1,70 @@
...
+/// You have to balance the retain & release calls in all cases to avoid leaking memory.
+class HostResolveCore {
+    let continuation: HostResolvedContinuation
+
+    init(_ continuation: HostResolvedContinuation) {
+        self.continuation = continuation
+    }
+
+    private func getRetainedSelf() -> UnsafeMutableRawPointer {
+        return Unmanaged<HostResolveCore>.passRetained(self).toOpaque()
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -4,106 +4,80 @@
 import AwsCIo
 import Foundation
 
-// swiftlint:disable trailing_whitespace
+public protocol IStreamable {
```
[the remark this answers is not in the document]
waahm7: Thanks, done.

Source/AwsCommonRuntimeKit/io/Stream.swift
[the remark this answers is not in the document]
waahm7: Fixed.

Source/AwsCommonRuntimeKit/io/StreamCore.swift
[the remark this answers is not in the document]
waahm7: Fixed.

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -4,106 +4,80 @@
...
-        length = Int64(impl.length)
-        awsStream = impl
-        rawValue = aws_input_stream()
-        rawValue.vtable = UnsafePointer<aws_input_stream_vtable>(&vtable)
-        rawValue.impl = Unmanaged<AwsInputStream>.passUnretained(self).toOpaque()
-    }
+    /// (Optional) throws Seek not supported error by default.
+    func seek(offset: UInt64) throws
+
+    /// buffer count should not be modified.
```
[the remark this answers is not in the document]
waahm7: Thanks, done.

Source/AwsCommonRuntimeKit/Utilities.swift
```diff
@@ -145,6 +145,19 @@ extension UnsafePointer: PointerConformance {}
...
+        _ arg1: String?, _ body: (UnsafePointer<aws_byte_cursor>?) -> Result
+) -> Result {
+    if let arg1 = arg1 {
+        return arg1.withCString { arg1C in
+            return withUnsafePointer(to: aws_byte_cursor_from_c_str(arg1C)) { byteCursorPointer in
+                return body(byteCursorPointer)
+            }
+        }
+    }
+    return body(nil)
```
epau, line 158: Nit: Prefer `guard` when applicable and prefer to omit `return` keyword from closures when possible.

```suggestion
    guard let arg1 = arg1 else {
        return body(nil)
    }
    return arg1.withCString { arg1C in
        withUnsafePointer(to: aws_byte_cursor_from_c_str(arg1C)) { byteCursorPointer in
            body(byteCursorPointer)
        }
    }
```

Source/AwsCommonRuntimeKit/crt/CommonRuntimeError.swift
```diff
@@ -1,6 +1,29 @@
...
 // SPDX-License-Identifier: Apache-2.0.
 
+import AwsCCommon
+import Foundation
+
 public enum CommonRunTimeError: Error {
     case crtError(CRTError)
 }
+
+public struct CRTError: Equatable {
```
epau, line 11: I'm likely missing a ton of context but something doesn't seem right to essentially have two types that have the same name, except one uses the abbreviation and one does not. Is it possible to clean this up to reduce confusion and make things easier to understand?

Any reason why `CRTError` can't conform to `Error`? And perhaps we can nuke `CommonRunTimeError`.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -273,40 +273,37 @@ public class ByteBuffer: Codable {
...
 
-extension ByteBuffer: AwsStream {
-    public var status: aws_stream_status {
-        return aws_stream_status(is_end_of_stream: self.currentIndex == array.count, is_valid: true)
-    }
+extension ByteBuffer: IStreamable {
 
-    public var length: UInt {
-        return UInt(array.count)
+    public func length() -> UInt64 {
```
epau, line 278: Nit: ooc why change to a function type? imo it'd be preferable to keep as a computed var to stay consistent with how Swift models a similar property on it's structures like Array.count, Data.count, String.count etc. 

Also nit, `count` would be a preferable name here to stay consistent with Swift naming, such as UnsafeMutableBufferPointer.count.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -273,40 +273,37 @@ public class ByteBuffer: Codable {
...
+        return UInt64(array.count)
     }
 
-    public func seek(offset: Int64, basis: aws_stream_seek_basis) -> Bool {
-        let targetOffset: Int64
-        if basis.rawValue == AWS_SSB_BEGIN.rawValue {
-            targetOffset = offset
+    public func seek(offset: Int64, streamSeekType: StreamSeekType) throws {
```
epau, line 282: Nit: Doc comments for this (and all public methods and properties) would be awesome.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -273,40 +273,37 @@ public class ByteBuffer: Codable {
...
-            let result = buffer.buffer.advanced(by: buffer.len)
-            let resultBufferPointer = UnsafeMutableBufferPointer.init(start: result, count: dataArray.count)
-            dataArray.copyBytes(to: resultBufferPointer)
-            self.currentIndex = arrayEnd
-            buffer.len += dataArray.count
-            return true
+    public func read(buffer: UnsafeMutableBufferPointer<UInt8>) -> Int? {
+        let endIndex = self.currentIndex + min(array.count - self.currentIndex, buffer.count)
+        if self.currentIndex == endIndex {
+            return nil
```
epau, line 300: Nit: Prefer `guard`

```suggestion
        guard self.currentIndex != endIndex else {
            return nil
```

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -273,40 +273,37 @@ public class ByteBuffer: Codable {
...
-        let dataArray = array[self.currentIndex..<(arrayEnd)]
-        if dataArray.count > 0 {
-            let result = buffer.buffer.advanced(by: buffer.len)
-            let resultBufferPointer = UnsafeMutableBufferPointer.init(start: result, count: dataArray.count)
-            dataArray.copyBytes(to: resultBufferPointer)
-            self.currentIndex = arrayEnd
-            buffer.len += dataArray.count
-            return true
+    public func read(buffer: UnsafeMutableBufferPointer<UInt8>) -> Int? {
+        let endIndex = self.currentIndex + min(array.count - self.currentIndex, buffer.count)
```
epau, line 298: Nit: It'd be great to stay consistent with our usage of `self`. Can we either change `self.currentIndex` in this function to just `currentIndex` or update the contents of `func seek(offset: Int64, streamSeekType: StreamSeekType)` to use `self.currentIndex`?

Source/AwsCommonRuntimeKit/io/ClientBootstrap.swift
```diff
@@ -1,26 +1,26 @@
 //  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 //  SPDX-License-Identifier: Apache-2.0.
 import AwsCIo
-import Foundation
 
-public final class ClientBootstrap {
+public class ClientBootstrap {
```
epau, line 5: ooc why remove `final`?

Source/AwsCommonRuntimeKit/io/ClientBootstrap.swift
```diff
@@ -1,26 +1,26 @@
...
+        guard let rawValue = (withUnsafePointer(to: hostResolver.getHostResolutionConfig()) { hostResolutionConfigPointer in
+            var options = aws_client_bootstrap_options(
+                    event_loop_group: elg.rawValue,
+                    host_resolver: hostResolver.rawValue,
+                    host_resolution_config: hostResolutionConfigPointer,
+                    on_shutdown_complete: shutdownOptions.shutdown_callback_fn,
+                    user_data: shutdownOptions.shutdown_callback_user_data
+            )
+           return aws_client_bootstrap_new(allocator.rawValue, &options)
+        }) else {
```
epau, line 23: Nit: This can be a bit tricky to read, especially at first glance, given folks are used to seeing the contents of the `else` block right after the `guard` line. But here we have a closure definition, then the `else`.

Can we define a closure to encapsulate this logic to make the `guard` simpler to read?
Something like:

```suggestion
        let getRawValue: () -> UnsafeMutablePointer<aws_client_bootstrap> = {
            withUnsafePointer(to: hostResolver.getHostResolutionConfig()) { hostResolutionConfigPointer in
                var options = aws_client_bootstrap_options(
                        event_loop_group: elg.rawValue,
                        host_resolver: hostResolver.rawValue,
                        host_resolution_config: hostResolutionConfigPointer,
                        on_shutdown_complete: shutdownOptions.shutdown_callback_fn,
                        user_data: shutdownOptions.shutdown_callback_user_data
                )
               return aws_client_bootstrap_new(allocator.rawValue, &options)
        }
        
        guard let rawValue = getRawValue() else {
```

Source/AwsCommonRuntimeKit/io/HostAddress.swift
```diff
@@ -4,31 +4,21 @@
...
-        self.host = nil
-        self.address = nil
-        self.recordType = .typeA
-        self.expiry = 0
-        self.useCount = 0
-        self.connectionFailureCount = 0
-        self.weight = 0
+    public init(host: String,
+                address: String,
+                recordType: AddressRecordType = .typeA) {
```
epau, line 19: Nit: It's preferable to format functions with multiline arguments in Swift like so: 

```suggestion
    public init(
        host: String,
        address: String,
        recordType: AddressRecordType = .typeA
    ) {
```

I'll look into linter support for this...

Source/AwsCommonRuntimeKit/io/HostAddress.swift
```diff
@@ -4,31 +4,21 @@
...
-    let host: String?
-    let address: String?
-    let recordType: AddressRecordType
-    let expiry: UInt64
-    let useCount: Int
-    let connectionFailureCount: Int
-    let weight: UInt8
+    public var host: String
+    public var address: String
+    public var recordType: AddressRecordType
```
epau, line 9: Any particular reason to define these as `var`s? If not, can we make them `let`s?

Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -4,98 +4,93 @@
...
-                maxHosts: Int,
-                maxTTL: Int,
-                allocator: Allocator = defaultAllocator,
-                shutdownCallback: ShutdownCallback? = nil) throws {
-        let shutdownCallbackCore = ShutdownCallbackCore(shutdownCallback)
+    public static func makeDefault(eventLoopGroup: EventLoopGroup,
+                                   maxHosts: Int,
+                                   maxTTL: Int,
+                                   allocator: Allocator = defaultAllocator,
+                                   shutdownCallback: ShutdownCallback? = nil) throws -> HostResolver {
```
epau, line 16: Nit: A more Swifty format

```suggestion
    public static func makeDefault(
        eventLoopGroup: EventLoopGroup,
        maxHosts: Int,
        maxTTL: Int,
        allocator: Allocator = defaultAllocator,
        shutdownCallback: ShutdownCallback? = nil
    ) throws -> HostResolver {
```

Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -4,98 +4,93 @@
...
+        try HostResolver(eventLoopGroup: eventLoopGroup, maxHosts: maxHosts, maxTTL: maxTTL, allocator: allocator, shutdownCallback: shutdownCallback)
+    }
+
+    init(eventLoopGroup: EventLoopGroup,
+         maxHosts: Int,
+         maxTTL: Int,
+         allocator: Allocator = defaultAllocator,
+         shutdownCallback: ShutdownCallback? = nil) throws {
```
epau, line 24: Nit: A more Swifty format

```suggestion
    init(
        eventLoopGroup: EventLoopGroup,
        maxHosts: Int,
        maxTTL: Int,
        allocator: Allocator = defaultAllocator,
        shutdownCallback: ShutdownCallback? = nil
    ) throws {
```

Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -4,98 +4,93 @@
...
-                    el_group: elg.rawValue,
-                    shutdown_options: shutdownCallbackCorePointer,
-                    system_clock_override_fn: nil)
-            //TODO: use const in C-IO to avoid mutable pointer here
-            return withUnsafeMutablePointer(to: &options) { aws_host_resolver_new_default(allocator.rawValue, $0) }
+            var options = aws_host_resolver_default_options()
+            options.max_entries = maxHosts
+            options.el_group = eventLoopGroup.rawValue
+            options.shutdown_options = shutdownCallbackCorePointer
+            return aws_host_resolver_new_default(allocator.rawValue, &options)
```
epau, line 33: Similar to a comment above, can we define this logic in a `getRawValue` closure to make the `guard` statement easier to read.

Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -4,98 +4,93 @@
...
-                            _ hostName: UnsafePointer<aws_string>!,
+private func onHostResolved(_ resolver: UnsafeMutablePointer<aws_host_resolver>?,
+                            _ hostName: UnsafePointer<aws_string>?,
                             _ errorCode: Int32,
-                            _ hostAddresses: UnsafePointer<aws_array_list>!,
+                            _ hostAddresses: UnsafePointer<aws_array_list>?,
                             _ userData: UnsafeMutableRawPointer!) {
-    let options = userData.assumingMemoryBound(to: ResolverOptions.self)
+    let hostResolverCore = Unmanaged<ContinuationCore<[HostAddress]>>.fromOpaque(userData).takeRetainedValue()
+    if errorCode != AWS_OP_SUCCESS {
```
epau, line 79: Nit: prefer `guard` 

```suggestion
    guard errorCode == AWS_OP_SUCCESS else {
```

Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -4,98 +4,93 @@
...
+        hostResolverCore.continuation.resume(throwing: CommonRunTimeError.crtError(CRTError(code: errorCode)))
+        return
+    }
 
-    let length = aws_array_list_length(hostAddresses)
-    var addresses: [HostAddress] = Array(repeating: HostAddress(), count: length)
+    // Success
+    let length = aws_array_list_length(hostAddresses!)
+    var addresses = [HostAddress]()
```
epau, line 86: This looks great as is but I just want to highlight that Swift has some nice tools to write code in a more functional programming style which has the nice effect of allowing us to define `addreses` as a constant -> `let addresses`

```swift
let addresses = (0..<length).map { index in 
    var address: UnsafeMutableRawPointer! = nil
    aws_array_list_get_at_ptr(hostAddresses!, &address, index)
    let hostAddressCType = address.bindMemory(to: aws_host_address.self, capacity: 1).pointee
    return HostAddress(hostAddress: hostAddressCType)
}
```

No need to make any changes here, just wanted to highlight :)

Source/AwsCommonRuntimeKit/crt/CommonRuntimeError.swift
```diff
@@ -1,6 +1,29 @@
...
+    public let code: Int32
+    public let message: String
+    public let name: String
+
+    public init<T: BinaryInteger>(code: T) {
+        if code > INT32_MAX || code <= 0 {
+            self.code = AWS_OP_ERR // Error Unknown
```
graebm, line 18: trivial: ~AWS_OP_ERR~ -> `AWS_ERROR_UNKNOWN` or set `-1` if you really want to set something different

AWS_OP_ERR and AWS_OP_SUCCESS are not themselves error codes. They're the values we return from a function that has set a thread-local error-code

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -4,106 +4,99 @@
 import AwsCIo
 import Foundation
 
-// swiftlint:disable trailing_whitespace
+/// This is for streaming input.
+public protocol IStreamable {
```
epau, line 8: Can we improve the naming here? What does the `I` stand for? 
Would just `Stream` or `InputStream` be a better fit? Its ok if the name doesn't end in `--able`, the Swift standard library has many protocols that don't follow that naming convention such as `Sequence` and `Collection`.

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -4,106 +4,99 @@
...
 
-private var vtable = aws_input_stream_vtable(seek: doSeek,
-        read: doRead,
-        get_status: doGetStatus,
-        get_length: doGetLength,
-        acquire: { _ = Unmanaged<AwsInputStream>.fromOpaque($0!.pointee.impl).retain() },
-        release: { Unmanaged<AwsInputStream>.fromOpaque($0!.pointee.impl).release() })
+    /// Optional, throws length not supported error by default
+    func length() throws -> UInt64
```
epau, line 11: Why provide a default implementation if it just throws an error by default?

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
waahm7: Some implementations of length can throw like the file handle, so we can't have simple var.

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -4,106 +4,99 @@
...
+    /// - Throws: Throws Seek not supported error by default.
+    func seek(offset: Int64, streamSeekType: StreamSeekType) throws
 
-    func seek(offset: Int64, basis: aws_stream_seek_basis) -> Bool
-    func read(buffer: inout aws_byte_buf) -> Bool
+    /// Read up to buffer.count bytes into the buffer.
+    /// buffer.count must not be modified.
+    /// Return the number of bytes copied.
+    /// Return nil if the end of file has been reached.
+    /// Return 0 if data is not yet available.
```
epau, line 27: Nit: Proper doc comment formatting and slight changes to the docs (feel free to ignore or modify)
 
```suggestion
    /// Reads up to `buffer.count` bytes and returns the number of bytes read.
    /// buffer.count must not be modified.
    ///
    /// - Parameters:
    ///     - buffer: The buffer to read from
    /// - Returns: `nil` if the end of the file has been reached. Returns `0` if the data is not yet available. Otherwise returns the number of bytes read.
```

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -4,106 +4,99 @@
...
-            data.copyBytes(to: result, count: data.count)
-            buffer.len += data.count
-            return true
+    public func seek(offset: Int64, streamSeekType: StreamSeekType) throws {
+        let targetOffset: UInt64
+        switch streamSeekType {
+        case .begin:
+            if offset < 0 {
+                throw CommonRunTimeError.crtError(CRTError(code: AWS_IO_STREAM_INVALID_SEEK_POSITION.rawValue))
+            }
```
epau, line 74: Nit: Prefer `guard` 

```suggestion
            guard offset >= 0 else {
                throw CommonRunTimeError.crtError(CRTError(code: AWS_IO_STREAM_INVALID_SEEK_POSITION.rawValue))
            }
```

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -4,106 +4,99 @@
...
+            targetOffset = UInt64(offset)
+        case .end:
+            let length = try self.length()
+            if offset > 0 || abs(offset) > length {
```
epau, line 78: Nit: Prefer `guard`

```suggestion
            guard offset <= 0, abs(offset) <= length else {
```

Source/AwsCommonRuntimeKit/io/StreamCore.swift
```diff
@@ -0,0 +1,95 @@
...
+    var rawValue: aws_input_stream
+    let iStreamable: IStreamable
+    var isEndOfStream: Bool = false
+    private let allocator: Allocator
+    private var vtable = aws_input_stream_vtable(seek: doSeek,
+            read: doRead,
+            get_status: doGetStatus,
+            get_length: doGetLength,
+            acquire: { _ = Unmanaged<IStreamCore>.fromOpaque($0!.pointee.impl).retain() },
+            release: { Unmanaged<IStreamCore>.fromOpaque($0!.pointee.impl).release() })
```
epau, line 18: Nit: More Swifty format

```suggestion
    private var vtable = aws_input_stream_vtable(
        seek: doSeek,
        read: doRead,
        get_status: doGetStatus,
        get_length: doGetLength,
        acquire: { _ = Unmanaged<IStreamCore>.fromOpaque($0!.pointee.impl).retain() },
        release: { Unmanaged<IStreamCore>.fromOpaque($0!.pointee.impl).release() }
    )
```

graebm (approved):
fix & ship

Source/AwsCommonRuntimeKit/io/StreamCore.swift
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    deinit {
+        allocator.release(vtablePointer)
+    }
+}
+
+private func doSeek(_ stream: UnsafeMutablePointer<aws_input_stream>!,
+                    _ offset: Int64,
+                    _ seekBasis: aws_stream_seek_basis) -> Int32 {
```
epau, line 40: Nit: More Swifty format

```suggestion
private func doSeek(
    _ stream: UnsafeMutablePointer<aws_input_stream>!,
    _ offset: Int64,
    _ seekBasis: aws_stream_seek_basis
) -> Int32 {
```

Source/AwsCommonRuntimeKit/io/StreamCore.swift
```diff
@@ -0,0 +1,95 @@
...
+        return AWS_OP_SUCCESS
+    } catch CommonRunTimeError.crtError(let crtError) {
+        return aws_raise_error(crtError.code)
+    } catch {
+        return aws_raise_error(Int32(AWS_IO_STREAM_SEEK_FAILED.rawValue))
+    }
+}
+
+private func doRead(_ stream: UnsafeMutablePointer<aws_input_stream>!,
+                    _ buffer: UnsafeMutablePointer<aws_byte_buf>!) -> Int32 {
```
epau, line 56: Nit: More Swifty format

```suggestion
private func doRead(
    _ stream: UnsafeMutablePointer<aws_input_stream>!,
    _ buffer: UnsafeMutablePointer<aws_byte_buf>!
) -> Int32 {
```

Source/AwsCommonRuntimeKit/io/StreamCore.swift
```diff
@@ -0,0 +1,95 @@
...
+        return AWS_OP_SUCCESS
+    } catch CommonRunTimeError.crtError(let crtError) {
+        return aws_raise_error(crtError.code)
+    } catch {
+        return aws_raise_error(Int32(AWS_IO_STREAM_READ_FAILED.rawValue))
+    }
+}
+
+private func doGetStatus(_ stream: UnsafeMutablePointer<aws_input_stream>!,
+                         _ result: UnsafeMutablePointer<aws_stream_status>!) -> Int32 {
```
epau, line 77: Nit: More Swifty format

```suggestion
private func doGetStatus(
    _ stream: UnsafeMutablePointer<aws_input_stream>!,
    _ result: UnsafeMutablePointer<aws_stream_status>!
) -> Int32 {
```

Source/AwsCommonRuntimeKit/io/TlsContextOptions.swift
```diff
@@ -5,47 +5,39 @@ import AwsCIo
...
     }
-    #endif
 
-    #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
-    public init(clientWithMtlsPkcs12Path path: String,
-                password: String,
-                allocator: Allocator = defaultAllocator) throws {
+    init(mtlsPkcs12FromPath path: String,
+         password: String,
+         allocator: Allocator) throws {
```
epau, line 25: Nit: More Swifty format

```suggestion
    init(
        mtlsPkcs12FromPath path: String,
        password: String,
        allocator: Allocator
    ) throws {
```

Source/AwsCommonRuntimeKit/io/StreamCore.swift
```diff
@@ -0,0 +1,95 @@
...
+    let iStreamCore = Unmanaged<IStreamCore>.fromOpaque(stream.pointee.impl).takeUnretainedValue()
+    result.pointee = aws_stream_status(is_end_of_stream: iStreamCore.isEndOfStream, is_valid: true)
+    return AWS_OP_SUCCESS
+}
+
+private func doGetLength(_ stream: UnsafeMutablePointer<aws_input_stream>!,
+                         _ result: UnsafeMutablePointer<Int64>!) -> Int32 {
```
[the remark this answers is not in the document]
waahm7: Thanks, I will create a separate PR which fixes this issue everywhere and use the lint file in #91.

Source/AwsCommonRuntimeKit/crt/CommonRuntimeError.swift
waahm7: The idea is that we want to have CommonRuntimeError which can have both Swift as well as C errors. Currently, it doesn't have any Swift Errors but it might in the future. CRTError just encapsulates the C error. We couldn't think of a better name for CRTError. Please free to recommend if you any better suggestions.

Source/AwsCommonRuntimeKit/io/ClientBootstrap.swift
waahm7: Any reason why we want to have final?

waahm7: The codebase had inconsistent final with some classes for no reason. We decided to remove that to keep it consistent.

Source/AwsCommonRuntimeKit/io/retryer/CRTAWSRetryStrategy.swift
```diff
@@ -2,114 +2,123 @@
...
-
-    var rawValue: UnsafeMutablePointer<aws_retry_strategy>
-
-    init(retryStrategy: UnsafeMutablePointer<aws_retry_strategy>,
-         allocator: Allocator) {
-        self.rawValue = retryStrategy
-        self.allocator = allocator
-    }
+//TODO: rename class to RetryStrategy or CRTRetryStrategy. We have inconsistent CRT/AWS as a prefix of some classes.
+// I am not renaming it for now because it messes up the git change log. Will create a separate PR for just renaming.
```
epau, line 8: Food for thought: We don't need to use any prefixes (here and throughout this codebase). There won't be any name collisions given names just need to be unique within the same module.

waahm7: Thanks, I will remove the inconsistent prefix

Source/AwsCommonRuntimeKit/crt/CommonRuntimeError.swift
waahm7: Thanks, resolved.

Source/AwsCommonRuntimeKit/crt/CommonRuntimeError.swift
epau: Would consumers need to handle C errors differently than Swift errors? Do we expect to model a Swift error differently than a C error? Could we have a single error struct that has multiple initializers to handle being initialized from C and from Swift?

waahm7: Ganesh wanted the ability to differentiate between C & Swift Errors and model them differently.

Source/AwsCommonRuntimeKit/io/ClientBootstrap.swift
epau: Ok gotcha, I think it can be boiled down to whether or not we want consumers to be able to subclass the class or not.
If yes -> don't use `final`
If no -> use `final

When in doubt, it's preferable to not support subclassing and use `final`. This has some nice, while minor, runtime performance boosts via static dispatch since the compiler knows the location of the methods since there are no subclasses.

Source/AwsCommonRuntimeKit/io/TlsContext.swift
```diff
@@ -26,3 +24,8 @@ public class TlsContext {
         aws_tls_ctx_release(rawValue)
     }
 }
+
+public enum TlsMode {
```
epau, line 28: Nit: I think it'd be preferable to name this `TLSMode`

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/io/ClientBootstrap.swift
waahm7: Please correct me if I am wrong. My understanding is that consumers can not subclass this unless we make this open explicitly.

Source/AwsCommonRuntimeKit/crt/CommonRuntimeError.swift
epau: Ok gotcha, I'll make a note of this and we can revisit cleaning up the error model when ganesh is back. I have some ideas but I also have very little context so would love ganesh's thoughts.

Source/AwsCommonRuntimeKit/io/HostAddress.swift
waahm7: These are var For consistency. We decided to keep all struct members as var instead of forcing consumers to create new structs if they want to update a value.

Source/AwsCommonRuntimeKit/io/HostResolver.swift
waahm7: Thanks, I will update it in a separate formatting PR.

Source/AwsCommonRuntimeKit/io/HostResolver.swift
waahm7: Thanks, I will update it in a separate formatting PR.

epau (approved):
Nothing blocking, just formatting/clean-up changes. 
I'll make a note to investigate adding rules to the linter to catch these and ideally also have some sort of formatter to auto apply these rules when applicable.

Source/AwsCommonRuntimeKit/io/Stream.swift
waahm7: Unfortunately, Swift already has Stream & InputStream classes. I stands for Input.

Source/AwsCommonRuntimeKit/io/Stream.swift
waahm7: This code is called by C. It can have some alternate implementation if some stream doesn't support length.

Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -4,98 +4,93 @@
...
+                                   maxHosts: Int,
+                                   maxTTL: Int,
+                                   allocator: Allocator = defaultAllocator,
+                                   shutdownCallback: ShutdownCallback? = nil) throws -> HostResolver {
+        try HostResolver(eventLoopGroup: eventLoopGroup, maxHosts: maxHosts, maxTTL: maxTTL, allocator: allocator, shutdownCallback: shutdownCallback)
+    }
+
+    init(eventLoopGroup: EventLoopGroup,
+         maxHosts: Int,
+         maxTTL: Int,
```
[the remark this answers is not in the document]
waahm7: Updated, thanks. In other language bindings, we don't even have maxTTL. Maybe we should just nuke it.

Source/AwsCommonRuntimeKit/io/StreamCore.swift
waahm7: Thanks, I will update it in a separate formatting PR.

Source/AwsCommonRuntimeKit/io/StreamCore.swift
waahm7: Thanks, I will update it in a separate formatting PR.

Source/AwsCommonRuntimeKit/io/TlsContextOptions.swift
waahm7: Thanks, I will update it in a separate formatting PR.

Source/AwsCommonRuntimeKit/io/Stream.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/io/Stream.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -4,106 +4,99 @@
...
+            length = try self.offset()
         } else {
-            targetOffset = self.offsetInFile - UInt64(offset)
+            savedPos = self.offsetInFile
+            self.seekToEndOfFile()
+            length = self.offsetInFile
         }
-        self.seek(toFileOffset: targetOffset)
-        return true
+        try self.seek(toOffset: savedPos)
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/io/Stream.swift
```diff
@@ -4,106 +4,99 @@
...
+    public func seek(offset: Int64, streamSeekType: StreamSeekType) throws {
+        let targetOffset: UInt64
+        switch streamSeekType {
+        case .begin:
+            if offset < 0 {
+                throw CommonRunTimeError.crtError(CRTError(code: AWS_IO_STREAM_INVALID_SEEK_POSITION.rawValue))
+            }
+            targetOffset = UInt64(offset)
+        case .end:
+            let length = try self.length()
```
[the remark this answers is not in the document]
waahm7: Unfortunately, in file handle, that's the only way we have found so far to calculate length.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -273,40 +273,37 @@ public class ByteBuffer: Codable {
...
+        if self.currentIndex == endIndex {
+            return nil
         }
-        return !self.status.is_end_of_stream
+
+        let dataArray = array[self.currentIndex..<endIndex]
+        dataArray.copyBytes(to: buffer, count: dataArray.count)
```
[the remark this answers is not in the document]
waahm7: Thanks, updated.

Source/AwsCommonRuntimeKit/io/TlsContext.swift
waahm7: Thanks, updated.

Source/AwsCommonRuntimeKit/io/HostResolver.swift
waahm7: Thanks, updated

Source/AwsCommonRuntimeKit/crt/CommonRuntimeError.swift
waahm7: I would love to clean this up too but haven't found a better way yet.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
waahm7: In this particular case, I have added the documentation in the protocol for Seek & Read functions.

Source/AwsCommonRuntimeKit/io/ClientBootstrap.swift
waahm7: Thanks, incorporated.

Source/AwsCommonRuntimeKit/io/HostResolver.swift
waahm7: Thanks, fixed.

PR #88, opened 2022-10-25
Fix Structs & Remove AWSStringByteCursor

waahm7:
*Description of changes:*
- Make C Struct consistent by creating a CStruct protocol. 
- Add helper functions for C Structs
- Make C Structs on the fly when needed.
- Nuke AWSStringByteCursor


Source/AwsCommonRuntimeKit/crt/CStruct.swift
```diff
@@ -0,0 +1,41 @@
...
+) -> Result {
+    if let arg1 = arg1 {
+        return arg1.withCStruct { cStruct in
+            return withUnsafePointer(to: cStruct as! T) { structPointer in
+                body(structPointer)
+            }
+        }
+    }
+    return body(nil)
+}
```
waahm7, line 41: We will need to define more overrides of this functions as T1, T2, T3.... I tried to find a way to make it generic by passing an array or something but could not get it to work.

Source/AwsCommonRuntimeKit/crt/CStruct.swift
```diff
@@ -0,0 +1,41 @@
...
+            return withUnsafePointer(to: cStruct) { body($0) }
+        }
+    }
+
+    func withCPointer<Result>(_ body: (UnsafePointer<RawType>) -> Result) -> Result {
+        return withCStruct { cStruct in
+            return withUnsafePointer(to: cStruct) { body($0) }
+        }
+    }
+}
```
waahm7, line 28: I experimented with multiple approaches to come up with a consistent approch for all different structs. Defining a protocol helps with keep it consistent and create functions that we can apply to any C Struct like withCPointer. However, some structs had shutdown options which require special hanlding like HttpClientConnectionOptions.swift

Source/AwsCommonRuntimeKit/http/HttpProxyOptions.swift
```diff
@@ -3,43 +3,36 @@
...
-        allocator.release(rawValue)
+        return withByteCursorFromStrings(basicAuthUsername ?? "",
+                                         basicAuthPassword ?? "",
+                                         hostName) {
+            userNamePointer, passwordPointer, hostPointer in
+            cProxyOptions.host = hostPointer
+            cProxyOptions.auth_username = userNamePointer
+            cProxyOptions.auth_password = passwordPointer
+            return body(cProxyOptions)
+        }
```
waahm7, line 36: The problem with just having this function without any protocol was that ProxyOptions are nullable. So 
`proxyOptions.withCStruct() {cStruct in }` wouldn't work if proxyOptions is null. Another solution I tried was to make it a static function at https://github.com/awslabs/aws-crt-swift/pull/88/commits/0459fd145dc2e95fbd2ff882723743736c168607. That approach can also work but there is no relation between different structs which is more flexible but less consistent.

waahm7: I am not sure if this complexity is worth the consistency. The main problem with static function approach I think is that we cannot create a generic helper function to avoid nested closures.

Source/AwsCommonRuntimeKit/auth/imds/CRTIMDSClient.swift
```diff
@@ -2,7 +2,7 @@
 //  SPDX-License-Identifier: Apache-2.0.
 
 import AwsCAuth
-// TODO: fix pointers and look for other errors
+// TODO: fix pointers and look for other errors, continue continuation if error during sync call, Use core class and Unmanaged
```
waahm7, line 5: This class has the previous buggy callback implementation of callbacks. I will fix it as part of future PR (Auth layer Refactor)

Source/AwsCommonRuntimeKit/auth/imds/CRTIMDSClient.swift
```diff
@@ -27,11 +27,12 @@ public class CRTIMDSClient {
...
     ///    - resourcePath: `String` path of the resource to query
     public func getResource(resourcePath: String) async throws -> String? {
-        let resourcePathCursor = AWSStringByteCursor(resourcePath, allocator: allocator)
         return try await withCheckedThrowingContinuation { (continuation: ResourceContinuation) in
-            let callbackData = CRTIMDSClientResourceCallbackData(continuation: continuation)
-            let pointer: UnsafeMutableRawPointer = fromPointer(ptr: callbackData)
-            aws_imds_client_get_resource_async(rawValue, resourcePathCursor.byteCursor, resourceCallback, pointer)
+            resourcePath.withByteCursor { resourcePathCursor in
+                let callbackData = CRTIMDSClientResourceCallbackData(continuation: continuation)
+                let pointer: UnsafeMutableRawPointer = fromPointer(ptr: callbackData)
```
graebm, line 33: I don't see anything happening to retain callbackData and prevent it from being cleaned up before the e
completion-callback fires?


shouldn't we be doing:
`let pointer = Unmanaged.passRetained(callbackData).toOpaque()`

instead of that using that weird `fromPointer()` function .. that weirdly allocates a POINTER  on the heap but doesn't seem to do anything to keep the thing it's pointing AT alive

feel free to make this a TODO if this kind of thing needs to be fixed a million times in the codebase

Source/AwsCommonRuntimeKit/http/HttpClientConnectionManager.swift
```diff
@@ -13,27 +13,8 @@ public class HttpClientConnectionManager {
         self.options = options
```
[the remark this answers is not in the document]
waahm7: Thanks, resolved.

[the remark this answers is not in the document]
waahm7: I will look for other places where we can get rid of this.

Source/AwsCommonRuntimeKit/auth/signing/SigV4HttpRequestSigner.swift
```diff
@@ -53,53 +51,43 @@ public class SigV4HttpRequestSigner {
...
-                            .crtError(CRTError(code: errorCode)))
-                }
-            }
 
-        }, callbackPointer)
+                    if let continuation = callback.pointee.continuation {
+                        if errorCode == 0,
+                           let signingResult = signingResult {
+
+                            let signedRequest = aws_apply_signing_result_to_http_request(callback.pointee.request.rawValue,
```
graebm, line 74: sorry, commenting on old code... feel free to TODO this in the future

I haven't looked into what circumstance `continuation` might not be set ...
but don't we ALWAYS want to do `aws_apply_signing_result_to_http_request()`, even if continuation isn't set?

Source/AwsCommonRuntimeKit/crt/CStruct.swift
```diff
@@ -0,0 +1,70 @@
...
+    return withOptionalCStructPointer(to: arg1) { arg1Pointer in
+        return withOptionalCStructPointer(to: arg2) { arg2Pointer in
+            return withOptionalCStructPointer(to: arg3) { arg3Pointer in
+                return withOptionalCStructPointer(to: arg4) { arg4Pointer in
+                    return body(arg1Pointer, arg2Pointer, arg3Pointer, arg4Pointer)
+                }
+            }
+        }
+    }
+}
```
graebm: seriously, trailing newlines pls

Source/AwsCommonRuntimeKit/auth/signing/SigningConfig.swift
```diff
@@ -4,20 +4,16 @@
 import AwsCAuth
 
 // TODO: verify callback logic, fix pointers, and maybe error handling
-public struct SigningConfig {
+public class SigningConfig: CStruct {
```
graebm, line 7: why change these all to `class`?

Source/AwsCommonRuntimeKit/http/HttpMonitoringOptions.swift
```diff
@@ -2,28 +2,25 @@
 //  SPDX-License-Identifier: Apache-2.0.
 import AwsCHttp
 
-public class HttpMonitoringOptions {
+public class HttpMonitoringOptions: CStruct {
     /// Minimum amount of throughput, in bytes per second, for a connection to be considered healthy
-    public let minThroughputBytesPerSecond: Int
+    public var minThroughputBytesPerSecond: UInt64
```
graebm, line 7: I genuinely don't know if we want members of options structs/classes to all be `let` or `var`, but we should probably be consistent

it's pretty inconsistent right now

Source/AwsCommonRuntimeKit/io/SocketOptions.swift
```diff
@@ -3,60 +3,42 @@
...
+    typealias RawType = aws_socket_options
+    func withCStruct<Result>(_ body: (aws_socket_options) -> Result) -> Result {
+        var cSocketOptions = aws_socket_options()
+        cSocketOptions.type = socketType.rawValue
+        cSocketOptions.domain = SocketDomain.ipv4.rawValue
+        cSocketOptions.connect_timeout_ms = connectTimeoutMs
+        cSocketOptions.keep_alive_interval_sec = 0
+        cSocketOptions.keep_alive_timeout_sec = 0
+        cSocketOptions.keep_alive_max_failed_probes = 0
+        cSocketOptions.keepalive = false
```
graebm, line 41: forgot to actually set some values, i.e. instead of 0 this line should be:
`cSocketOptions.keep_alive_interval_sec = keepaliveIntervalSec`

Source/AwsCommonRuntimeKit/http/HttpClientConnectionOptions.swift
```diff
@@ -70,4 +71,41 @@ public struct HttpClientConnectionOptions {
...
+                cManagerOptions.socket_options = socketOptionsPointer
+                cManagerOptions.tls_connection_options = tlsOptions?.rawValue
+                cManagerOptions.http2_prior_knowledge = false
+                cManagerOptions.monitoring_options = monitoringOptionsPointer
+                cManagerOptions.host = hostNameCursor
+                cManagerOptions.port = port
+                cManagerOptions.initial_settings_array = nil
+                cManagerOptions.num_initial_settings = 0
+                cManagerOptions.max_closed_streams = 0
+                cManagerOptions.http2_conn_manual_window_management = false
```
graebm, line 99: just delete the lines where we set fields 0/false, they're not necessary anymore

Source/AwsCommonRuntimeKit/io/TlsConnectionOptions.swift
```diff
@@ -5,35 +5,35 @@ import AwsCIo
 
 public final class TlsConnectionOptions {
 	private let allocator: Allocator
-    var rawValue: UnsafeMutablePointer<aws_tls_connection_options>
-
+    private var _rawValue: UnsafeMutablePointer<aws_tls_connection_options>
```
graebm, line 8: I thought it was weird that you didn't convert this one to CStruct

But looking at it, it is a bit different. This options-struct has init/cleanup functions and setters, instead of being a plain old non-owning struct.

We **could** still use the CStruct protocol though. its `withCStruct()` function would call init(), then the setters, then invoke the body callback, and then call clean_up() afterwards

The setters can't ACTUALLY fail if you look into them. They can only fail if allocation fails, but we don't let allocation fail anymore, so we can skip checking those errors

OR

keep TlsConnectionOptions a class, but have it *feel* like the other options-structs by using swift properties, instead of setter functions. And we could still have a `withCStruct()` function that just delivered the pre-existing struct

Source/AwsCommonRuntimeKit/io/TlsConnectionOptions.swift
```diff
@@ -5,35 +5,35 @@ import AwsCIo
...
+		_rawValue.pointee.timeout_ms = 30_000
         #else
-		rawValue.pointee.timeout_ms = 3_000
+		_rawValue.pointee.timeout_ms = 3_000
         #endif
     }
 
 	public func setAlpnList(_ alpnList: String) throws {
-		if aws_tls_connection_options_set_alpn_list(rawValue, self.allocator.rawValue, alpnList) != AWS_OP_SUCCESS {
+		if aws_tls_connection_options_set_alpn_list(_rawValue, self.allocator.rawValue, alpnList) != AWS_OP_SUCCESS {
```
graebm, line 22: like I said above, this function can't *really* fail, it's from the bad old days where any allocation could fail

Source/AwsCommonRuntimeKit/io/retryer/CRTExponentialBackoffRetryOptions.swift
```diff
@@ -21,4 +21,14 @@ public struct CRTExponentialBackoffRetryOptions {
...
     }
+
+    typealias RawType = aws_exponential_backoff_retry_options
+    func withCStruct<Result>(_ body: (aws_exponential_backoff_retry_options) -> Result) -> Result {
+        var cExponentialBackoffRetryOptions = aws_exponential_backoff_retry_options()
+        cExponentialBackoffRetryOptions.el_group = options.backOffRetryOptions.eventLoopGroup.rawValue
+        cExponentialBackoffRetryOptions.max_retries = options.backOffRetryOptions.maxRetries
+        cExponentialBackoffRetryOptions.backoff_scale_factor_ms = options.backOffRetryOptions.backOffScaleFactor
+        cExponentialBackoffRetryOptions.jitter_mode = options.backOffRetryOptions.jitterMode.rawValue
+        cExponentialBackoffRetryOptions.generate_random = ni
```
graebm: `ni`?

Source/AwsCommonRuntimeKit/http/HttpMonitoringOptions.swift
waahm7: I think It should be var, as we make c structs on the fly, there is no point in making it let and forcing users to create new objects if they want to modify something.

Source/AwsCommonRuntimeKit/io/retryer/CRTExponentialBackoffRetryOptions.swift
waahm7: Fixed, thanks.

Source/AwsCommonRuntimeKit/io/SocketOptions.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/auth/signing/SigV4HttpRequestSigner.swift
waahm7: Continuation should always be set as we pass it as part of user data. Will refactor when I refactor this class to use core classes and learning more about it.

waahm7: This is a side effect of treating user data as nullable which will be fixed by !

Source/AwsCommonRuntimeKit/auth/signing/SigV4HttpRequestSigner.swift
```diff
@@ -53,53 +51,43 @@ public class SigV4HttpRequestSigner {
...
+        //TODO: fix callback
+        config.withCPointer { configPointer in
+            configPointer.withMemoryRebound(to: aws_signing_config_base.self, capacity: 1) { configBasePointer in
+                let callbackPointer: UnsafeMutablePointer<SigningCallbackData> = fromPointer(ptr: callbackData)
+                aws_sign_request_aws(allocator.rawValue,
+                        signable,
+                        configBasePointer, { (signingResult, errorCode, userData) -> Void in
+                    guard let userData = userData else {
+                        return
+                    }
```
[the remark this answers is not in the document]
waahm7: Thanks, I will fix this everywhere.

Source/AwsCommonRuntimeKit/auth/signing/SigningConfig.swift
waahm7: I thought classes have additional features, so why not. Plus, it will help us avoid needless copies by value. I will do a bit of dive deep on Struct vs Classes and make it consistent.

Source/AwsCommonRuntimeKit/http/HttpClientConnectionOptions.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/auth/signing/SigningConfig.swift
waahm7: Thanks, turning it to struct because they are thread safe and recommeded by Swift.

Source/AwsCommonRuntimeKit/auth/imds/CRTIMDSClient.swift
```diff
@@ -294,6 +273,24 @@ public class CRTIMDSClient {
...
+//TODO: fix callback
+private func onGetCredentialsCallback(credentialsPointer: OpaquePointer?,
+                                      errorCode: Int32,
+                                      userData: UnsafeMutableRawPointer!) {
+        let pointer = userData.assumingMemoryBound(to: CRTCredentialsProviderCallbackData.self)
+        if errorCode != AWS_OP_SUCCESS {
+            pointer.pointee.continuation?.resume(throwing: CommonRunTimeError.crtError(CRTError(code: errorCode)))
+            return
+        }
+        if let credentialsPointer = credentialsPointer,
```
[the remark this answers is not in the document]
waahm7: Thanks, updated.

Source/AwsCommonRuntimeKit/auth/imds/CRTIMDSClient.swift
```diff
@@ -294,6 +273,24 @@ public class CRTIMDSClient {
...
+        let crtCredentials = CRTCredentials(rawValue: credentialsPointer) {
+            pointer.pointee.continuation?.resume(returning: crtCredentials)
+        } else {
+            pointer.pointee.continuation?.resume(throwing: CommonRunTimeError.crtError(CRTError(code: errorCode)))
```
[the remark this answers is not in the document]
waahm7: Thanks, updated.

Source/AwsCommonRuntimeKit/http/HttpMonitoringOptions.swift
waahm7: - Changing all to var
- Callbacks are a let because of special handling required with retain/release.

Source/AwsCommonRuntimeKit/io/TlsConnectionOptions.swift
waahm7: Thanks, fixed. There is one other class TlsContextOptions which has a similar issue. Looking into if we can make it a struct instead.

Source/AwsCommonRuntimeKit/io/TlsConnectionOptions.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/auth/signing/SigV4HttpRequestSigner.swift
waahm7: I have refactored it a bit to remove some confusion. I will refactor it properly to use core class etc in the Auth Layer refactor PR.

Source/AwsCommonRuntimeKit/io/retryer/CRTExponentialBackoffRetryOptions.swift
```diff
@@ -1,14 +1,14 @@
...
-    let maxRetries: Int
-    let backOffScaleFactor: UInt32
-    let generateRandom: GenerateRandom?
-    let jitterMode: CRTExponentialBackoffJitterMode
+public struct CRTExponentialBackoffRetryOptions: CStruct {
+    public var eventLoopGroup: EventLoopGroup
+    public var maxRetries: Int
+    public var backOffScaleFactor: UInt32
+    public var jitterMode: CRTExponentialBackoffJitterMode
+    public let generateRandom: GenerateRandom?
```
waahm7, line 11: It's a callback, so I haven't made it a var.

Source/AwsCommonRuntimeKit/io/TlsContextOptions.swift
```diff
@@ -2,14 +2,14 @@
...
+public class TlsContextOptions: CStruct {
     private let allocator: Allocator
-    var rawValue: UnsafeMutablePointer<aws_tls_ctx_options>
+    private var rawValue: UnsafeMutablePointer<aws_tls_ctx_options>
     public static func isAlpnSupported() -> Bool {
         return aws_tls_is_alpn_available()
     }
 
-    public init(defaultClientWithAllocator allocator: Allocator = defaultAllocator) {
+    public init(allocator: Allocator = defaultAllocator) {
```
graebm, line 12: the reason this had the weird `defaultClientWithAllocator` name was to make it clear that this was the "default_client" constructor. This class has a LOT of constructors. There's also a "default_server" constructor in C. It's not bound to Swift, but the reasoning behind giving the "default client" function an explicit name was to disambiguate it from other functions we might bind in the future.

It's clunky that we name it "withAllocator" here, because that was the only arg whose name we could customize. I propose that we use factory methods instead of public init functions, so we can give them all good names without abusing the argument names like we are now. Ex:
```
static public makeDefaultClient(allocator: Allocator? = defaultAllocator) -> TlsContextOptions {...}
static public makeClientWithMtls(certificatePath: String, keyPath: String, ...) -> TlsContextOptions {...}
static public makeClientWithMtls(pkcs12Path: String, password: String, ...) -> TlsContextOptions {...}
```

Source/AwsCommonRuntimeKit/io/retryer/CRTExponentialBackoffRetryOptions.swift
waahm7: Yeah, that makes sense. Thanks. I thought it might give the wrong perception that callback can be updated if it hasn’t been triggered yet. But now, this makes more sense as Struct are a value type. I will update it.

Source/AwsCommonRuntimeKit/io/TlsConnectionOptions.swift
```diff
@@ -3,37 +3,39 @@
...
+		self.alpnList = alpnList
+		self.serverName = serverName
     }
 
-	public func setAlpnList(_ alpnList: String) throws {
-		if aws_tls_connection_options_set_alpn_list(_rawValue, self.allocator.rawValue, alpnList) != AWS_OP_SUCCESS {
-			throw CommonRunTimeError.crtError(.makeFromLastError())
+	typealias RawType = aws_tls_connection_options
+	func withCStruct<Result>(_ body: (aws_tls_connection_options) -> Result) -> Result {
+		var rawValue: UnsafeMutablePointer<aws_tls_connection_options> = allocator.allocate(capacity: 1)
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/io/TlsConnectionOptions.swift
```diff
@@ -3,37 +3,39 @@
...
 import AwsCIo
 
-public final class TlsConnectionOptions {
+public struct TlsConnectionOptions: CStruct {
 	private let allocator: Allocator
-    private var _rawValue: UnsafeMutablePointer<aws_tls_connection_options>
-	var rawValue: UnsafePointer<aws_tls_connection_options> { UnsafePointer(_rawValue) }
-	init(_ context: TlsContext, allocator: Allocator) {
+	public var context: TlsContext
+	public var alpnList: String?
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/io/TlsContextOptions.swift
```diff
@@ -2,14 +2,14 @@
...
         self.allocator = allocator
         self.rawValue = allocator.allocate(capacity: 1)
         aws_tls_ctx_options_init_default_client(rawValue, allocator.rawValue)
```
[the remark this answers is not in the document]
waahm7: Thanks, I didn't know about that. I will refactor this class as part of IO refactor PR that I am working on.

Source/AwsCommonRuntimeKit/io/TlsContextOptions.swift
waahm7: Makes sense, thanks.

Source/AwsCommonRuntimeKit/io/TlsConnectionOptions.swift
```diff
@@ -3,37 +3,37 @@
 
 import AwsCIo
 
-public final class TlsConnectionOptions {
+public struct TlsConnectionOptions: CStruct {
 	private let allocator: Allocator
-    var rawValue: UnsafeMutablePointer<aws_tls_connection_options>
+	public var context: TlsContext
+	public var alpnList: [String]
```
graebm, line 9: trivial: have it be optional `alpnList: [String]?`

ALPN is an optional extension during TLS handshakes. Probably in our code passing an empty array, resulting in an empty string, results in the extension not being used ... but I'd rather our API be more explicit about it

waahm7: Thanks, fixed.

graebm (approved):
man this is beautiful now

Issue #97, opened 2022-11-07
SigV4a Signer


ganeshnj:
### Describe the feature

--/seps/accepted/shared/mrap.md

### Use Case

S3 Multi-region Access Points (MRAP) 

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

waahm7:
SigV4a support was added in v0.5.0.

PR #105, opened 2022-11-21
Refactor signing

waahm7:
*Issue #, if available:*
https://github.com/awslabs/aws-crt-swift/issues/97

*Description of changes:*
- Refactor the signing logic
- Add support for Sigv4a


Source/AwsCommonRuntimeKit/auth/signing/SigningConfig.swift
```diff
@@ -2,121 +2,101 @@
...
-        self.service = service
-        self.region = region
-        self.signedBodyHeader = signedBodyHeader
-        self.signedBodyValue = signedBodyValue
-        self.flags = flags
-        self.shouldSignHeader = shouldSignHeader
-        self.signatureType = signatureType
-        self.signingAlgorithm = signingAlgorithm
-        self.configType = configType
+public enum SignatureType {
```
waahm7, line 7: I just consolidated all the enums related to signing in one file.

Source/AwsCommonRuntimeKit/auth/signing/SigV4HttpRequestSigner.swift
```diff
@@ -2,14 +2,123 @@
 //  SPDX-License-Identifier: Apache-2.0.
 
 import AwsCAuth
+import Foundation
 
-public class SigV4HttpRequestSigner {
-    public var allocator: Allocator
+//TODO: update file name
+public class HttpRequestSigner {
```
graebm, line 8: Don't combine the signing-config class with the signer class. This is one of those things where the config is often passed around as its own thing, to be used later.

That said, `SigV4HttpRequestSigner` was a bad name for this class. We eventually want to support...:
- different signing algorithms (currently SigV4 and SigV4a, but there have been other algorithms in the past (SigV2, SigV3) and surely SigV5 will come some day.
- signing different types of objects (HTTP requests, chunks within an HTTP request, etc)

aws-crt-python has 1 [AwsSigningConfig](https://github.com/awslabs/aws-crt-python/blob/191b52d307793ac105d7389b0bce646917857ed1/awscrt/auth.py#L458) class, and currently only exposes 1 function [aws_sign_request(request, config)](https://github.com/awslabs/aws-crt-python/blob/191b52d307793ac105d7389b0bce646917857ed1/awscrt/auth.py#L732) for signing HTTP requests

aws-crt-java also has 1 [AwsSigningConfig](https://github.com/awslabs/aws-crt-java/blob/main/src/main/java/software/amazon/awssdk/crt/auth/signing/AwsSigningConfig.java) class, but then it has multiple functions for signing: [AwsSigner.signRequest(...)](https://github.com/awslabs/aws-crt-java/blob/dc195bbf0e698f3dd71fb1c23cdfa3db9bdea710/src/main/java/software/amazon/awssdk/crt/auth/signing/AwsSigner.java#L26), [AwsSigner.signChunk(...)](https://github.com/awslabs/aws-crt-java/blob/dc195bbf0e698f3dd71fb1c23cdfa3db9bdea710/src/main/java/software/amazon/awssdk/crt/auth/signing/AwsSigner.java#L50)

waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/auth/signing/SigV4HttpRequestSigner.swift
```diff
@@ -31,58 +47,79 @@ public class SigV4HttpRequestSigner {
     ///    - `config`: The `SigningConfig` to use when signing.
     /// - `Throws`: An error of type `AwsCommonRuntimeError` which will pull last error found in the CRT
     /// - `Returns`: Returns a signed http request `HttpRequest`
-    public func signRequest(request: HttpRequest, config: SigningConfig) async throws -> HttpRequest {
-        typealias SignedContinuation = CheckedContinuation<HttpRequest, Error>
+    public static func signRequest(request: HttpRequest, config: SigningConfig, allocator: Allocator = defaultAllocator) async throws -> HttpRequest {
```
graebm, line 50: Is the `HttpRequest` that gets returned, the same `HttpRequest` that was passed in? If so, that's confusing

Seems better to do either:
1) Mutate the request that gets passed in, do not return a value
OR
2) Do not mutate the request that gets passed in, return a new HttpRequest

In theory, I like 2. But in reality, we should probably do 1, since there's lots of complexity over how to do ref-counting on the body stream when we make a copy of the request

Source/AwsCommonRuntimeKit/auth/signing/SigningConfig.swift
```diff
@@ -2,121 +2,233 @@
 //  SPDX-License-Identifier: Apache-2.0.
 
 import AwsCAuth
+import Foundation
 
-// TODO: verify callback logic, fix pointers, and maybe error handling
-public struct SigningConfig: CStruct {
-    public typealias ShouldSignHeader = (String) -> Bool
+public struct SigningConfig: CStructWithUserData {
```
graebm, line 7: related to my earlier comment about naming the signer...

if we name it AwsSigner, everything in this file should probably also get an "AWS" prefix
or we leave the prefix off and the current names are fine

Source/AwsCommonRuntimeKit/auth/signing/SigV4HttpRequestSigner.swift
waahm7: Thanks, I agree that doing both is a bit weird. I have updated it to approach 2.

Source/AwsCommonRuntimeKit/auth/signing/SigV4HttpRequestSigner.swift
```diff
@@ -2,12 +2,28 @@
 //  SPDX-License-Identifier: Apache-2.0.
 
 import AwsCAuth
+import Foundation
 
-public class SigV4HttpRequestSigner {
-    public var allocator: Allocator
+//TODO: update file name
+public class HttpRequestSigner {
```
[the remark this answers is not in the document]
waahm7: Thanks, I have removed the HttpRequest for now.

Source/AwsCommonRuntimeKit/auth/signing/SigningConfig.swift
waahm7: Thanks, I have not changed it for now.

Source/AwsCommonRuntimeKit/auth/signing/SigV4HttpRequestSigner.swift
```diff
@@ -31,58 +28,81 @@ public class SigV4HttpRequestSigner {
...
-    public func signRequest(request: HttpRequest, config: SigningConfig) async throws -> HttpRequest {
-        typealias SignedContinuation = CheckedContinuation<HttpRequest, Error>
+    public static func signRequest(request: HttpRequest, config: SigningConfig, allocator: Allocator = defaultAllocator) async throws {
 
-        //TODO: Fix. This is weird. We should not throw make from last error here.
-        if config.configType != .aws || (config.credentialsProvider == nil && config.credentials == nil) {
-            throw CommonRunTimeError.crtError(.makeFromLastError())
+        guard let signable = aws_signable_new_http_request(allocator.rawValue, request.rawValue) else {
+            fatalError("Unable to create a new singable http request")
+        }
```
waahm7, line 35: This can only fail if we fail to init a list due to size. We can either do fatalError here or define a new error to throw.

Source/AwsCommonRuntimeKit/auth/signing/SigV4HttpRequestSigner.swift
```diff
@@ -31,58 +47,79 @@ public class SigV4HttpRequestSigner {
...
     /// - `Throws`: An error of type `AwsCommonRuntimeError` which will pull last error found in the CRT
     /// - `Returns`: Returns a signed http request `HttpRequest`
-    public func signRequest(request: HttpRequest, config: SigningConfig) async throws -> HttpRequest {
-        typealias SignedContinuation = CheckedContinuation<HttpRequest, Error>
+    public static func signRequest(request: HttpRequest, config: SigningConfig, allocator: Allocator = defaultAllocator) async throws -> HttpRequest {
 
-        //TODO: Fix. This is weird. We should not throw make from last error here.
-        if config.configType != .aws || (config.credentialsProvider == nil && config.credentials == nil) {
-            throw CommonRunTimeError.crtError(.makeFromLastError())
+        let signable = aws_signable_new_http_request(allocator.rawValue, request.rawValue)
```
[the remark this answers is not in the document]
waahm7: I have added a fatal error because C just returns null and doesn't set any error. We can also create a new error code to throw.

Source/AwsCommonRuntimeKit/auth/signing/SigV4HttpRequestSigner.swift
waahm7: I just realized that the request signing is supposed to be async. I am not really sure about approach 1. Looking more into it.

waahm7: I have added a comment as discussed after failing to find a better approach for now.

graebm (approved):

PR #106, opened 2022-11-22
Refactor sdkutils

waahm7:
*Description of changes:*
- Refactor SdkUtils parsing config & profile logic
- Add & Refactor implementation of endpoint 2.0 #79 


Source/AwsCommonRuntimeKit/Utilities.swift
```diff
@@ -87,6 +87,19 @@ extension aws_array_list {
...
+
+    func awsStringListToStringArray() -> [String] {
+        var arrayList = self
+        var result = [String]()
+
+        for index in 0..<self.length {
+                var valPtr: UnsafeMutableRawPointer! = nil
+                aws_array_list_get_at(&arrayList, &valPtr, index)
+                let strPtr = valPtr.bindMemory(to: aws_string.self, capacity: 1)
+                result.append(String(awsString: strPtr)!)
```
waahm7, line 99: I am not sure if we should just crash if we can not convert AwsString to String. Ganesh created a new stringConversionError and used that.

waahm7: Personally, I think this should not happen unless there is something funky going on with memory.

Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointsResolvedEndpoint.swift
```diff
@@ -0,0 +1,87 @@
...
+
+        var iter = aws_hash_iter_begin(cHeaders)
+        while !aws_hash_iter_done(&iter) {
+            // Get the key
+            let keyPtr = iter.element.key.bindMemory(to: aws_string.self, capacity: 1)
+            let key = String(awsString: keyPtr)!
+
+            // Get the value
+            let arrayPtr = iter.element.value.bindMemory(to: aws_array_list.self, capacity: 1)
+            headers[key] = arrayPtr.pointee.awsStringListToStringArray()
```
waahm7, line 77: If we can’t parse should we crash, throw a newStringConversionError, or return nil?

waahm7: I am not sure if we are expecting any parsing errors from C's get_headers API.

DmitriyMusatkin: that api will return error if endpoint resolved to an error. otherwise it will init out var to a list (that potentially can have 0 elements)

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSProfileCollection.swift
```diff
@@ -24,34 +26,39 @@ public class CRTAWSProfileCollection {
...
-    public init(configProfileCollection: CRTAWSProfileCollection,
-                credentialProfileCollection: CRTAWSProfileCollection,
-                source: CRTAWSProfileSourceType,
-                allocator: Allocator = defaultAllocator) {
-        self.rawValue = aws_profile_collection_new_from_merge(allocator.rawValue,
-                                                              configProfileCollection.rawValue,
-                                                              credentialProfileCollection.rawValue)
+    /// Create a new profile collection by merging a config-file-based profile
+    /// collection and a credentials-file-based profile collection
+    public init?(configProfileCollection: CRTAWSProfileCollection,
```
graebm, line 50: for all these init functions, it's debatable whether they should be `init?(...)` or `init(...) throws`

throws seems a bit safer... since the error might not be something dead-obvious like "file doesn't exist". It might be a parsing error that the user wouldn't realize unless we told them

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSProfileCollection.swift
```diff
@@ -24,34 +26,39 @@ public class CRTAWSProfileCollection {
         self.rawValue = profilePointer
     }
 
-    public init(fromBuffer buffer: ByteBuffer,
-                source: CRTAWSProfileSourceType,
-                allocator: Allocator = defaultAllocator) {
+    /// Create a new profile collection by parsing text in a buffer. Primarily for testing.
+    init(fromBuffer buffer: ByteBuffer,
```
graebm, line 30: this `init()` should be `throws` or `?` like the other init functions. `aws_profile_collection_new_from_buffer()` could fail

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSProfileProperty.swift
```diff
@@ -6,17 +6,19 @@ import AwsCSdkUtils
 public struct CRTAWSProfileProperty {
     let rawValue: OpaquePointer
 
-    public init(rawValue: OpaquePointer) {
+    init(rawValue: OpaquePointer) {
         self.rawValue = rawValue
     }
 
+    /// Returns the property's string value
     public var value: String? {
```
graebm, line 14: I think this can be ~String?~ -> `String`?

The [comments](https://github.com/awslabs/aws-c-sdkutils/blob/4068c2c1f1342a79f8a42aeed848c7efabc1161f/include/aws/sdkutils/aws_profile.h#L117-L121) say:
```
/**
 * Returns a reference to the property's string value
 */
AWS_SDKUTILS_API
const struct aws_string *aws_profile_property_get_value(const struct aws_profile_property *property);
```

which *sounds* like it's always there.

And looking a the source code ... `aws_profile_property->value` is never NULL: https://github.com/awslabs/aws-c-sdkutils/blob/4068c2c1f1342a79f8a42aeed848c7efabc1161f/source/aws_profile.c#L256

Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointsResolvedEndpoint.swift
```diff
@@ -0,0 +1,87 @@
...
+    /// Get the type of the resolved endpoint
+    /// - Returns: The type of the resolved endpoint
+    public func getType() -> CRTAWSEndpointsResolvedEndpointType {
+        let type = aws_endpoints_resolved_endpoint_get_type(rawValue)
+        return CRTAWSEndpointsResolvedEndpointType(rawValue: type)
+    }
+
+    /// Get the URL of the resolved endpoint
+    /// - Returns: The URL of the resolved endpoint
+    public func getURL() throws -> String? {
```
graebm, line 26: should this be `throws -> String` instead of `throws -> String?`

in what circumstance do we not-error but also not-return-a-string?

Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointsResolvedEndpoint.swift
```diff
@@ -0,0 +1,87 @@
...
+        var url = aws_byte_cursor()
+        guard aws_endpoints_resolved_endpoint_get_url(rawValue, &url) == AWS_OP_SUCCESS else {
+            throw CommonRunTimeError.crtError(.makeFromLastError())
+        }
+        return url.toString()
+    }
+
+    /// Get the properties of the resolved endpoint
+    /// - Returns: The properties of the resolved endpoint
+    public func getProperties() throws -> [String: AnyHashable]? {
```
graebm, line 36: similar, should the return value have a `?` on it?

guess this question applies to all the getters on this class...

Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointsResolvedEndpoint.swift
graebm: also, we should probably document the circumstances under which errors might be thrown. The header files say:
> Valid only if resolved endpoint has endpoint type and will error otherwise.

but I'm not 100% sure what that means, not being super familiar with all this...

is it possible for an endpoint to be just totally invalid and every function on it fails?
why bother creating an object that's invalid? why not just have its creation function report failure?

Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointsRuleEngine.swift
```diff
@@ -0,0 +1,61 @@
...
+            }
+
+            guard let ruleSet = (ruleSet.withByteCursor { ruleSetCursor in
+                aws_endpoints_ruleset_new_from_string(allocator.rawValue, ruleSetCursor)
+            }) else {
+                throw CommonRunTimeError.crtError(.makeFromLastError())
+            }
+
+            defer {
+                aws_partitions_config_release(partitions)
```
graebm, line 32: possible memory leak: if `aws_endpoints_ruleset_new_from_string()` fails, then `partitions` will leak

we need to declare the deferred release of each thin immediately after its successfully created

Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointsResolvedEndpoint.swift
DmitriyMusatkin: it might be good to explain here that result of resolution is either endpoint or error. errors only define message, so trying to use url or header getters on them will throw. this error is different from engine itself having an error which is reported using the typical c return.

Does swift support algebraic types? ideally this would be typed as ResolutionOutcome = Endpoint | Error. And getters would be specified only on corresponding classes meaning that using getter on the wrong type is impossible

Source/AwsCommonRuntimeKit/Utilities.swift
waahm7: Agreed, thanks.

Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointsResolvedEndpoint.swift
waahm7: Swift does support that we can define certain methods only on certain types. But this should not be an issue because C throws an invalid argument error. Is it possible that C returns Success (ResolutionOutcome was as exepected) but the byte_cursor length is 0?

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSProfileCollection.swift
waahm7: I agree throw is better than init?. The C code just returns null, either we should introduce proper error handling in C or just throw sdkutils_general error for now.

Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointsResolvedEndpoint.swift
DmitriyMusatkin: We dont have to mimic c interface completely in swift if swift language provides better constructs for what we want. 
From conceptual perspective output of resolution is (Endpoint(url, headers, optional properties) OR Error(message)), i.e. its algebraic sum type that swift supports through a slightly weird enum syntax.

DmitriyMusatkin: With that said. The api is already used by swift, so there might not be much incentive to rewrite interface

Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointsRuleEngine.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSProfileCollection.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSProfileProperty.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointsResolvedEndpoint.swift
waahm7: Thanks a lot for the help, I have updated the API and error handling as discussed.

Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointsResolvedEndpoint.swift
waahm7: Thanks, Fixed. None of the getters are nullable anymore.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSProfileCollection.swift
```diff
@@ -17,50 +19,63 @@ public class CRTAWSProfileCollection {
...
+         source: CRTAWSProfileSourceType,
+         allocator: Allocator = defaultAllocator) throws {
         var byteArray = buffer.toByteArray()
         let byteCount = byteArray.count
-        let byteBuf = byteArray.withUnsafeMutableBufferPointer { pointer -> aws_byte_buf in
+        var byteBuf = byteArray.withUnsafeMutableBufferPointer { pointer -> aws_byte_buf in
```
graebm, line 36: trivial: would withUnsafe~Mutable~BufferPointer() work?

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSProfile.swift
```diff
@@ -6,17 +6,16 @@ import AwsCSdkUtils
 public struct CRTAWSProfile {
     let rawValue: OpaquePointer
```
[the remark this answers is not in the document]
waahm7: Thanks, that is a really great catch. Justin wrote this, so I am not sure who to consult with. Going through the code, I noticed that that this code has nested hashtables, and everything gets cleaned up when we destroy collection. A simple solution would be to just keep a reference to collection in profile. We can also do a complete copy but due to nested nature, we might do unnecessary copies that are not needed.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSProfileCollection.swift
```diff
@@ -17,50 +19,63 @@ public class CRTAWSProfileCollection {
...
             let byteBuf = aws_byte_buf(len: byteCount,
-                                       buffer: pointer.baseAddress,
-                                       capacity: byteCount,
-                                       allocator: allocator.rawValue)
+                    buffer: pointer.baseAddress,
+                    capacity: byteCount,
+                    allocator: allocator.rawValue)
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSProfileCollection.swift
waahm7: Thanks, good catch.

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSProfile.swift
```diff
@@ -3,11 +3,14 @@
 
 import AwsCSdkUtils
 
-public struct CRTAWSProfile {
+public class CRTAWSProfile {
     let rawValue: OpaquePointer
+    // Keep a reference of collection to keep it alive
+    let collection: CRTAWSProfileCollection
```
graebm, line 9: oh, sneaky. nice simple fix

This is how we USED TO do the bindings, before we added ref-counting to everything. If a few different objects all had their lifetimes tied together, then the binding for one object would keep pointers to the bindings for all the other objects too, to keep them alive.

It starts falling apart when the web of entanglements gets too complex, or things start having async shutdown procedures, but it works here 👍

graebm (approved):

PR #107, opened 2022-11-25
Refactor common

waahm7:
*Description of changes:*
- Remove unused code
- Refactor ByteBuffer to use Data and remove unused code.


Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -2,285 +2,69 @@
...
-#if os(Linux)
-     import Glibc
- #else
-     import Darwin
- #endif
 
-// swiftlint:disable identifier_name superfluous_disable_command
-// Todo: this is being used. Refactor
-public class ByteBuffer: Codable {
+public class ByteBuffer {
```
waahm7, line 8: This class is used in the SDK. Orginallay it was an [uint8] buffer with a current index and conforms to IStreamable protocol. I have done the following the changes. Ideally this should have been an extension over the data class to conform to IStreamable protocol. However, because this is being used in SDK, I haven’t completely nuked it.

1. Use Data instead of [uint8] as this is more natural to Swift for ByteBuffer in memory. There is an ByteBuffer implementation but that belongs to SwiftNIO package so we can't use it. Another advantage is the Swift copy on write optimization by using Data directly. 
2. Remove all the endiness & putInt32 etc functions. As per my understanding, they are not used in the SDK and add needless complexity to this class.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -2,285 +2,69 @@
...
+    public func put(_ value: UInt8) {
+        data.append(value)
     }
 
-    public func put(_ value: UInt8) -> ByteBuffer {
-        array.append(value)
-        return self
+    public func put(_ value: [UInt8]) {
+        data.append(contentsOf: value)
     }
```
waahm7, line 39: Returing self is not needed https://github.com/awslabs/aws-crt-swift/pull/9#discussion_r526556883

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -2,285 +2,69 @@
...
-    public func getDouble(_ index: Int) -> Double {
-        let result = from(Array(array[index..<index + MemoryLayout<UInt64>.size]), UInt64.self)
-        return currentEndianness == .little ? Double(bitPattern: result.littleEndian)
-            : Double(bitPattern: result.bigEndian)
-    }
-
-    public func toByteArray() -> [UInt8] {
-        return array
+        data.append(value)
     }
```
waahm7, line 52: I have added a function getData, and all the simple wrapper functions of Data functionality can be executed directly if needed.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -2,285 +2,69 @@
...
-        }
-        let dataArray = Array(array[currentIndex..<(arrayEnd)])
-        if dataArray.count > 0 {
-            _ = buffer.put(dataArray)
-            self.currentIndex = arrayEnd
-            return dataArray.count
-        } else {
-            return 0
-        }
-    }
```
waahm7, line 273: Duplicate of read function.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -290,129 +74,19 @@ extension ByteBuffer: IStreamable {
...
-                // Stream error occured
-                throw stream.streamError!
-            } else if read == 0 {
-                // EOF
-                break
-            }
-            allocate(read)
-            putByte(buffer.pointee)
-        }
-    }
```
waahm7, line 343: I didn’t find any usage in SDK. Also, this feels like a bad function which just loads the entire input stream to memory.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -290,129 +74,19 @@ extension ByteBuffer: IStreamable {
...
-    ///   - filePath:  The path at which you would like a new file to be created at and written to.
-    /// - Returns: `FileHandle`
-    public func toFile(filePath: String) -> FileHandle? {
-        let fileManager = FileManager.default
-        if fileManager.createFile(atPath: filePath, contents: self.toData()) {
-            return FileHandle(forReadingAtPath: filePath)
-        } else {
-            return nil
-        }
-    }
```
waahm7, line 373: Just a simple wrapper over data object. So nuked.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -290,129 +74,19 @@ extension ByteBuffer: IStreamable {
...
-
-    /// Converts the array of bytes to `Foundation.InputStream`
-    ///
-    /// - Returns: `InputStream`
-    public func toStream() -> InputStream {
-        return InputStream(data: self.toData())
-    }
```
waahm7, line 380: Same as above.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -290,129 +74,19 @@ extension ByteBuffer: IStreamable {
...
-        var byteCursor = aws_byte_cursor_from_array(self.array, self.array.count)
-
-        var bytes = [UInt8](repeating: 0, count: Int(AWS_SHA256_LEN))
-        let result: ByteBuffer = bytes.withUnsafeMutableBufferPointer { pointer in
-            var buffer = aws_byte_buf(len: 0, buffer: pointer.baseAddress, capacity: Int(AWS_SHA256_LEN), allocator: allocator.rawValue)
-            aws_sha256_compute(allocator.rawValue, &byteCursor, &buffer, truncate)
-            return buffer.toByteBuffer()
-        }
-        return result
-    }
```
waahm7, line 404: I have moved this function to be an extension of Data instead which makes more sense.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -290,129 +74,19 @@ extension ByteBuffer: IStreamable {
...
-
-    func base64EncodedSha256(allocator: Allocator = defaultAllocator, truncate: Int = 0) -> String {
-        return sha256(allocator: allocator, truncate: truncate).toData().base64EncodedString()
-    }
```
waahm7, line 408: Nuked, as sha256 now returns Data. User can directly use base64EncodedString()

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -290,129 +74,19 @@ extension ByteBuffer: IStreamable {
...
-
-    func encodeToHexString() -> String {
-        var hexString = ""
-        for byte in array {
-            hexString += String(format: "%02x", UInt8(byte))
-        }
-
-        return hexString
-    }
```
waahm7, line 417: This is being used in the SDK. It is a simple helper function that belongs in the SDK and not here.

Source/Elasticurl/CommandLine.swift
```diff
@@ -4,8 +4,8 @@
 import Foundation
 import AwsCCommon
 // swiftlint:disable trailing_whitespace
-// Todo: remove/refactor
-public struct CommandLineParser {
+
+struct CommandLineParser {
```
waahm7, line 8: I have moved this class to ElasticCurl and made it private for now. We can nuke it in the future and used a built in command line parser.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -2,285 +2,68 @@
...
-     import Darwin
- #endif
 
-// swiftlint:disable identifier_name superfluous_disable_command
-// Todo: this is being used. Refactor
-public class ByteBuffer: Codable {
+public class ByteBuffer {
+
+    private var data: Data
+    private var currentIndex: Data.Index
```
waahm7, line 11: Why not a simple Int? https://forums.swift.org/t/data-popfirst-removefirst-adjust-indices/35965/5

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
waahm7: I have discussed it with @epau and SDK is using byte buffer only at [here](https://github.com/awslabs/aws-sdk-swift/blob/f24c44bd8788aff87a75063639f365e38794bf9c/AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift#L60)

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -2,417 +2,93 @@
...
-    }
-
-    public func put(_ value: ByteBuffer, offset: UInt = 0, maxBytes: UInt? = nil) {
-        var end: UInt = UInt(value.length())
-        if let maxBytes = maxBytes {
-             end = maxBytes
-        }
-        let bytesToTake = value.array[Int(offset)..<Int(end)]
-        array.append(contentsOf: bytesToTake)
+    public func put(_ value: [UInt8]) {
```
graebm, line 42: let's take these put() functions out if no one is using them
red code is best code

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -2,417 +2,93 @@
...
-#if os(Linux)
-     import Glibc
- #else
-     import Darwin
- #endif
 
-// swiftlint:disable identifier_name superfluous_disable_command
-// Todo: this is being used. Refactor
-public class ByteBuffer: Codable {
+public final class ByteBuffer: Decodable {
```
graebm, line 8: add a comment explaining what this class is, so I don't keep getting confused by why it's here

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -2,417 +2,93 @@
...
-        return currentEndianness
+    public init(data: Data) {
+        self.data = data
+        currentIndex = data.startIndex
     }
 
-    public func order(_ endianness: Endianness) -> ByteBuffer {
-        currentEndianness = endianness
-        return self
+    public init(bufferPointer: UnsafeMutablePointer<UInt8>, length: Int, capacity: Int) {
```
graebm, line 28: if there are no put functions, can we stop taking mutable data?

can we change `var data` to `let data`?

Source/AwsCommonRuntimeKit/sdkutils/CRTAWSProfileCollection.swift
```diff
@@ -31,13 +31,15 @@ public class CRTAWSProfileCollection {
     init(fromBuffer buffer: ByteBuffer,
```
graebm, line 31: why not take Data? instead of our ByteBuffer thing

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/Utilities.swift
```diff
@@ -67,10 +74,11 @@ extension TimeInterval {
...
 
-//Todo: refactor
-extension aws_byte_buf {
-    func toByteBuffer() -> ByteBuffer {
-        return ByteBuffer(ptr: self.buffer, len: self.len, capacity: self.capacity)
+extension aws_byte_cursor {
+    func toString() -> String? {
+        if self.len == 0 { return nil }
+        let data = Data(bytesNoCopy: self.ptr, count: self.len, deallocator: .none)
+        return String(data: data, encoding: .utf8)
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
waahm7: Thanks, nuked.

Source/AwsCommonRuntimeKit/Utilities.swift
```diff
@@ -46,6 +38,21 @@ extension String {
...
 
+extension Data {
+    func sha256(truncate: Int = 0, allocator: Allocator = defaultAllocator) -> Data {
+        self.withUnsafeBytes { bufferPointer in
+            var byteCursor = aws_byte_cursor_from_array(bufferPointer.baseAddress, count)
+            let bufferSize = Int(AWS_SHA256_LEN)
+            var bufferData = Data(count: bufferSize)
+            bufferData.withUnsafeMutableBytes { bufferDataPointer in
+                var buffer = aws_byte_buf_from_empty_array(bufferDataPointer.baseAddress, bufferSize)
+                aws_sha256_compute(allocator.rawValue, &byteCursor, &buffer, truncate)
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/http/HttpRequest.swift
```diff
@@ -3,49 +3,69 @@
...
+
 public class HttpRequest: HttpMessage {
 
-    public override init(allocator: Allocator = defaultAllocator) throws {
-        try super.init(allocator: allocator)
+    public init(method: String = "GET",
+                path: String = "/",
+                headers: HttpHeaders? = nil,
+                body: IStreamable? = nil,
+                allocator: Allocator = defaultAllocator) throws {
```
graebm, line 13: you could remove `throws` from this

`aws_http_message_set_request_method()` can't *really* fail
`aws_http_message_set_request_path()` can't *really* fail

these return `int` because they were written in the bad old days when allocations could fail

Source/AwsCommonRuntimeKit/auth/imds/CRTIMDSInstanceInfo.swift
```diff
@@ -5,20 +5,20 @@ import AwsCAuth
...
+    public let instanceId: String
+    public let billingProducts: [String]
+    public let instanceType: String
+    public let accountId: String
+    public let imageId: String
+    public let pendingTime: Date
+    public let architecture: String
+    public let kernelId: String
+    public let ramDiskId: String
+    public let region: String
```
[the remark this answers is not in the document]
waahm7: They don’t have to be always present, the only difference is whether we want to return nil or “” for empty value. Looking at the previous code, we were replacing nil with “”. The C code fetches a json from the instance, and the instance might have “” as the json value. The C code treats “” as a valid byte cursor. I thought it is a better API design to keep “” as empty instead of Nil.

Source/AwsCommonRuntimeKit/http/HttpRequest.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/http/HttpRequest.swift
```diff
@@ -3,49 +3,69 @@
...
-        set(value) {
-            guard let value = value else { return }
-            if (value.withByteCursor { valueCursor in
-                aws_http_message_set_request_method(self.rawValue, valueCursor)
-            }) != AWS_OP_SUCCESS {
-                self.method = nil
-            }
+    public func setMethod(method: String) throws {
+        guard (method.withByteCursor { methodCursor in
+            aws_http_message_set_request_method(self.rawValue, methodCursor)
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/http/HttpRequest.swift
```diff
@@ -3,49 +3,69 @@
...
+            let iStreamCore = IStreamCore(iStreamable: body, allocator: allocator)
+            aws_http_message_set_body_stream(self.rawValue, &iStreamCore.rawValue)
+        }
     }
 
-    public override init(headers: HttpHeaders, allocator: Allocator = defaultAllocator) throws {
-        try super.init(headers: headers, allocator: allocator)
+    public func getMethod() throws -> String {
+        var method = aws_byte_cursor()
+        guard aws_http_message_get_request_method(rawValue, &method) == AWS_OP_SUCCESS else {
```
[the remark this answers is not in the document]
waahm7: Thanks, I have changed it to a Swift property with getter & setter.

Source/AwsCommonRuntimeKit/http/HttpStreamCalbackCore.swift
```diff
@@ -52,11 +52,9 @@ private func onResponseHeaders(stream: UnsafeMutablePointer<aws_http_stream>?,
     var headers = [HttpHeader]()
```
graebm, line 52: utterly trivial: this filename misspells "Calback"

this could just be a TODO

Source/AwsCommonRuntimeKit/http/HttpStreamCalbackCore.swift
```diff
@@ -52,11 +52,9 @@ private func onResponseHeaders(stream: UnsafeMutablePointer<aws_http_stream>?,
...
 
     for cHeader in UnsafeBufferPointer(start: headerArray, count: headersCount) {
-        if let name = cHeader.name.toString(),
-           let value = cHeader.value.toString() {
-            let swiftHeader = HttpHeader(name: name, value: value)
-            headers.append(swiftHeader)
-        }
+        let name = cHeader.name.toString()
+        let value = cHeader.value.toString()
+        headers.append(HttpHeader(name: name, value: value))
```
graebm, line 57: we're not even copying the `compression` member

possible solutions:
- Give HttpHeader a private init that takes the C struct (and remembers to copy the compression)
- remove the `compression` member altogether until we're really ready to expose it to users (we haven't exposed it to any other language bindings ... it's a feature we're still kind of figuring out...)

graebm (approved):
fix & ship

Source/AwsCommonRuntimeKit/http/HttpStreamCalbackCore.swift
waahm7: Thanks, I will fix it as part of renaming PR.

Source/AwsCommonRuntimeKit/http/HttpStreamCalbackCore.swift
waahm7: I have already refactored the headers logic in #108. We can refactor it further as part of that PR.

PR #108, opened 2022-11-29
Refactor HttpHeaders

waahm7:
*Description of changes:*
- Add more unit tests
- Refactor HttpHeaders
- Minor improvements


Source/AwsCommonRuntimeKit/http/HttpMessage.swift
```diff
@@ -29,54 +28,53 @@ public class HttpMessage {
...
+    func getHeader(atIndex index: Int) -> HttpHeader {
         var header = aws_http_header()
-        if aws_http_message_get_header(self.rawValue, &header, index) == AWS_OP_SUCCESS {
-            return HttpHeader(rawValue: header)
+        guard aws_http_message_get_header(self.rawValue, &header, index) == AWS_OP_SUCCESS else {
+            fatalError("Index out of range")
         }
-        return nil
+        return HttpHeader(rawValue: header)
     }
```
waahm7, line 68: Crash if index is invalid. This is same as Swift's API https://developer.apple.com/documentation/swift/array/remove(at:)-1p2pj

Package.swift
```diff
@@ -6,7 +6,7 @@ let excludesFromAll = ["tests", "cmake", "CONTRIBUTING.md",
                         "CMakeLists.txt", "README.md"]
 var packageTargets: [Target] = []
 
-var package = Package(name: "AwsCrt",
+var package = Package(name: "aws-crt-swift",
```
waahm7, line 9: Change from #93

Source/AwsCommonRuntimeKit/http/HttpRequest.swift
```diff
@@ -35,21 +35,17 @@ public class HttpRequest: HttpMessage {
     /// Creates an http request which can be passed to a connection.
     /// - Parameters:
     ///   - method: Http method to use. Must be a valid http method and not empty.
-    ///   - path: Path of Http Request. Should not be empty.
+    ///   - path: Path of Http Request. Must not be empty.
```
graebm, line 38:
```suggestion
    ///   - path: Path and query string for Http Request. Must not be empty.
```

Source/Elasticurl/Elasticurl.swift
```diff
@@ -272,30 +267,22 @@ struct Elasticurl {
...
-            if headers.add(name: "Host", value: host),
-               headers.add(name: "User-Agent", value: "Elasticurl"),
-               headers.add(name: "Accept", value: "*/*"),
-               headers.add(name: "Swift", value: "Version 5.4") {
-                for header in context.headers {
-                    _ = headers.add(name: header.key, value: header.value)
-                }
-            }
+            var headers = [HttpHeader]()
+            headers.append(HttpHeader(name: "Host", value: host))
```
graebm, line 271: trivial: seems like an alternate constructor that accepted `name` and `value` as unnamed args would be nice

graebm (approved):
trivial fix & ship

Source/Elasticurl/Elasticurl.swift
waahm7: Thanks, I have added a addHeader function that takes name & value and removed compression.

PR #110, opened 2022-12-05
Fix naming

waahm7:
*Description of changes:*
- Update naming to make it consistent everywhere


Source/AwsCommonRuntimeKit/http/HttpClientConnectionManagerCallbackCore.swift
```diff
@@ -15,12 +15,19 @@ class HttpClientConnectionManagerCallbackCore {
...
-        aws_http_connection_manager_acquire_connection(connectionManager.rawValue, onConnectionSetup, getRetainedSelf())
+    static func acquireConnection(
+        continuation: ConnectionContinuation,
+        connectionManager: HttpClientConnectionManager) {
+
+        let callbackCore = HttpClientConnectionManagerCallbackCore(continuation: continuation,
+                                                                   connectionManager: connectionManager)
+        aws_http_connection_manager_acquire_connection(connectionManager.rawValue,
+                                                       onConnectionSetup,
+                                                       callbackCore.passRetained())
```
epau, line 30: Nit: Format

```suggestion
    static func acquireConnection(
        continuation: ConnectionContinuation,
        connectionManager: HttpClientConnectionManager
    ) {
        let callbackCore = HttpClientConnectionManagerCallbackCore(
            continuation: continuation,
            connectionManager: connectionManager
        )
        aws_http_connection_manager_acquire_connection(
            connectionManager.rawValue,
            onConnectionSetup,
            callbackCore.passRetained()
        )
```

Source/AwsCommonRuntimeKit/io/retryer/RetryToken.swift
```diff
@@ -4,7 +4,7 @@
 import AwsCIo
 
 /// This is just a wrapper for aws_retry_token which user can not create themself but pass around once acquired.
-public class CRTAWSRetryToken {
+public class RetryToken {
```
epau, line 7: It might be good to have the `AWS` prefix, since it appears that this token is aws specific given the underlying C type is `aws_retry_token`?

Source/AwsCommonRuntimeKit/auth/credentials/AWSCredentialsProvider.swift
```diff
@@ -330,22 +330,23 @@ extension AWSCredentialsProvider.Source {
...
-                x509Options.endpoint = endPointCursor
-                return withOptionalCStructPointer(
-                    proxyOptions,
-                    tlsConnectionOptions) { proxyOptionsPointer, tlsConnectionOptionsPointer in
-
-                    x509Options.proxy_options = proxyOptionsPointer
-                    x509Options.tls_connection_options = tlsConnectionOptionsPointer
-                    return aws_credentials_provider_new_x509(allocator.rawValue, &x509Options)
-                }})
+            guard let provider: UnsafeMutablePointer<aws_credentials_provider> = (
```
epau, line 333: Does `withByteCursorFromStrings` need to be wrapped in parentheses?

Source/AwsCommonRuntimeKit/io/retryer/RetryToken.swift
epau: I could go either way here, so defer to your best judgement

Source/AwsCommonRuntimeKit/auth/credentials/AWSCredentialsProvider.swift
waahm7: Yes, we need to need wrap withByteCursorFromStrings or not use trailing closure.

Source/AwsCommonRuntimeKit/http/HttpProxyEnvSettings.swift
```diff
@@ -5,11 +5,11 @@ import AwsCHttp
 public struct ProxyEnvSettings: CStruct {
```
graebm, line 5: the filename changed ~ProxyEnvSettings.swift~ -> `HttpProxyEnvSettings.swift`
but the class name didn't change

Test/AwsCommonRuntimeKitTests/http/HttpClientConnectionManagerTests.swift
```diff
@@ -3,17 +3,17 @@
 import XCTest
 @testable import AwsCommonRuntimeKit
 
-class HttpClientConnectionManagerTests: CrtXCBaseTestCase {
+class HttpClientConnectionManagerTests: XCBaseTestCase {
```
graebm, line 6: Why are these classes still "Http" and not "HTTP"?
it looked like we were doing all-caps acronyms

Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointsRequestContext.swift
```diff
@@ -4,7 +4,7 @@
 import AwsCSdkUtils
 
 /// Request context used for resolving endpoint
-public class CRTAWSEndpointsRequestContext {
+public class EndpointsRequestContext {
```
graebm, line 7: Seems like you were keeping the "AWS" prefix on classes that were AWS-specific like `AWSCredentials`, `AWSSigner`, `AWSProfile`, etc

but endpoints is also just an AWS thing and you removed the prefix ???

Source/AwsCommonRuntimeKit/io/retryer/RetryToken.swift
waahm7: Thanks, I have prefixed it with `aws` as Retry Strategy is only used in Auth layer which is specific to AWS. Btw, In C, everything is prefixed with `aws`.

Source/AwsCommonRuntimeKit/http/HttpProxyEnvSettings.swift
waahm7: Thanks, good catch.

Test/AwsCommonRuntimeKitTests/http/HttpClientConnectionManagerTests.swift
waahm7: Thanks, I have updated it because HTTP commonly appears as all upper case. 

Swift’s official guide states: “Acronyms and initialisms that commonly appear as all uppercase in American English should be uniformly up- or down-cased according to case conventions. Other acronyms should be treated as ordinary words.” 

Reference: https://www.swift.org/documentation/api-design-guidelines/#general-conventions

Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointsRequestContext.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/auth/credentials/AWSCredentialsProvider.swift
epau: In that case, its preferable to not use the trailing closure

waahm7: Thanks, fixed.

graebm (approved):
🧙‍♂️

Issue #99, opened 2022-11-07
Event Steaming encoding decoding


ganeshnj:
### Describe the feature

Bindings for encode and decode messages.

### Use Case

https://github.com/awslabs/smithy-swift/pull/460

CRT already has support for event stream, creating binding is preferred to not create duplication in Swift.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

waahm7:
I have added the support in 0.5.3.

PR #124, opened 2022-12-08
Bind Event-Stream Encode Decode

waahm7:
*Issue #, if available:* https://github.com/awslabs/aws-crt-swift/issues/99

*Description of changes:*
- Binds event stream encode decode


Package.swift
```diff
@@ -116,6 +125,22 @@ awsCCalPlatformExcludes.append("source/windows")
...
+awsCChecksumsExcludes.append("source/generic")
+#elseif arch(x86_64) || arch(i386)
+// include src/intel/asm
+awsCChecksumsExcludes.append("source/arm")
+awsCChecksumsExcludes.append("source/generic")
+#else
+// includes source/generic
+awsCChecksumsExcludes.append("source/arm")
+awsCChecksumsExcludes.append("source/intel")
+#endif
```
waahm7, line 142: We can’t exclude everything and specify what to include because [sources](https://developer.apple.com/documentation/packagedescription/target/sources) property recursively search for source files. It will include everything under `source/` and we will need to exclude as we are doing here.

waahm7: Even if we use sources and tell it to include some files, the Swift compiler still complains to explicitly exclude them.
` found 20 file(s) which are unhandled; explicitly declare them as resources or exclude from the target`

Source/AwsCommonRuntimeKit/event-stream/EventStreamHeader.swift
```diff
@@ -0,0 +1,70 @@
...
+
+public enum EventStreamHeaderValue: Equatable {
+    case bool(value: Bool)
+    case byte(value: Int8)
+    case int16(value: Int16)
+    case int32(value: Int32)
+    case int64(value: Int64)
+    case byteBuf(value: Data)
+    case string(value: String)
+    /// Timestamp is in milliseconds
```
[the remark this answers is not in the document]
waahm7: I have updated it to type TimeInterval, and the param name similar to https://developer.apple.com/documentation/foundation/date/1780353-init

Source/AwsCommonRuntimeKit/event-stream/EventStreamHeader.swift
```diff
@@ -0,0 +1,70 @@
...
+            value = .string(
+                value: aws_event_stream_header_value_as_string(rawValue).toString())
+        case AWS_EVENT_STREAM_HEADER_TIMESTAMP:
+            value = .timestamp(
+                value: aws_event_stream_header_value_as_timestamp(rawValue))
+        case AWS_EVENT_STREAM_HEADER_UUID:
+            let uuid = UUID(uuid: rawValue.pointee.header_value.static_val)
+            value = .uuid(value: uuid)
+        default:
+            fatalError("Unexpected header value type found.")
```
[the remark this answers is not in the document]
waahm7: This is just a side effect of C enums being non frozen. The C api will throw unknown header type if it doesn’t matches any known header type at https://github.com/awslabs/aws-c-event-stream/blob/main/source/event_stream.c#L1179

Source/AwsCommonRuntimeKit/event-stream/EventStreamMessage.swift
```diff
@@ -0,0 +1,121 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+
+import AwsCEventStreams
+import Foundation
+
+public struct EventStreamMessage {
+    var headers: [EventStreamHeader] = [EventStreamHeader]()
+    var payload: Data?
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/event-stream/EventStreamMessage.swift
```diff
@@ -0,0 +1,120 @@
...
+                        value)
+                case .byteBuf(var value):
+                    return value.withUnsafeMutableBytes {
+                        let bytes = $0.bindMemory(to: UInt8.self).baseAddress!
+                        return aws_event_stream_add_bytebuf_header(
+                            rawHeaders,
+                            headerName,
+                            UInt8(header.name.count),
+                            bytes,
+                            UInt16($0.count),
```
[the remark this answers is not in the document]
waahm7: Thanks, updated. I considered adding a new Swift error with a proper message but I think it is not worth it to break the API to make it worse for a rare error.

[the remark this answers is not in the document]
graebm: there's already an error for this: AWS_ERROR_EVENT_STREAM_MESSAGE_INVALID_HEADERS_LEN

this function is already marked as throwing

Source/AwsCommonRuntimeKit/event-stream/EventStreamHeader.swift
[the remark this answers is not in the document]
graebm: If we're going to use a built-in Swift type, let's use `Date`

Seems like Swift would prefer that
>On its own, a time interval does not specify a unique point in time, or even a span between specific times.

Source/AwsCommonRuntimeKit/event-stream/EventStreamMessage.swift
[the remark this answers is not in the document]
waahm7: This is invalid_headers_len error, which is not applicable to this situation. I thought overflow is better or we can define new error codes.

Source/AwsCommonRuntimeKit/event-stream/EventStreamHeader.swift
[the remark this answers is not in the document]
waahm7: Thanks, updated. It still has the issue of not getting the exact same value when decoding. I have updated the header comparison function to take that into account. Once @ganeshnj is back, we can confirm it with him if prefers int64 instead.

Source/AwsCommonRuntimeKit/event-stream/EventStreamMessage.swift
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/crt/Utilities.swift
```diff
@@ -90,6 +90,14 @@ extension Date {
         aws_date_time_init_epoch_secs(&date, self.timeIntervalSince1970)
         return date
     }
+
+    var millisecondsSince1970: Int64 {
+        Int64((self.timeIntervalSince1970 * 1000.0).rounded())
+    }
+
+    init(milliseconds: Int64) {
```
graebm, line 98: trivial
```suggestion
    init(millisecondsSince1970: Int64) {
```

Source/AwsCommonRuntimeKit/event-stream/EventStreamHeader.swift
```diff
@@ -5,7 +5,7 @@ import AwsCEventStreams
 import Foundation
 
 public struct EventStreamHeader {
-    /// Header name length can not be greater than UInt8.max
+    /// Header name length can not be greater than Int8.max
```
graebm, line 8: let's provide proper constants

```suggestion
    /// Header name length can not be greater than EventStreamHeader.maxNameLength
```

could just be a static?
`public static let maxNameLength = AWS_EVENT_STREAM_HEADER_NAME_LEN_MAX`

Source/AwsCommonRuntimeKit/event-stream/EventStreamHeader.swift
```diff
@@ -16,13 +16,13 @@ public enum EventStreamHeaderValue: Equatable {
     case int16(value: Int16)
     case int32(value: Int32)
     case int64(value: Int64)
-    /// Data length can not be greater then UInt16.max
+    /// Data length can not be greater than Int16.max
```
graebm, line 19:
```suggestion
    /// Data length can not be greater than EventStreamHeader.maxValueLength
```

Source/AwsCommonRuntimeKit/event-stream/EventStreamHeader.swift
```diff
@@ -16,13 +16,13 @@ public enum EventStreamHeaderValue: Equatable {
...
     case byteBuf(value: Data)
-    /// String length can not be greater then UInt16.max
+    /// String length can not be greater than Int16.max
```
graebm, line 21:
```suggestion
    /// String length can not be greater than EventStreamHeader.maxValueLength
```

Source/AwsCommonRuntimeKit/event-stream/EventStreamHeader.swift
```diff
@@ -68,7 +67,11 @@ extension EventStreamHeaderValue {
 
 extension EventStreamHeader: Equatable {
     public static func == (lhs: EventStreamHeader, rhs: EventStreamHeader) -> Bool {
-        lhs.name == rhs.name &&
+        if case let EventStreamHeaderValue.timestamp(value1) = lhs.value,
+           case let EventStreamHeaderValue.timestamp(value2) = rhs.value {
+            return value1.millisecondsSince1970 == value2.millisecondsSince1970
```
graebm, line 72: forgot to compare `.name` if they're dates

Package.swift
```diff
@@ -198,6 +228,20 @@ packageTargets.append(contentsOf: [
...
     ),
+    .target(
+        name: "AwsChecksums",
+        dependencies: ["AwsCCommon"],
+        path: "aws-common-runtime/aws-checksums",
+        exclude: awsCChecksumsExcludes,
+        cSettings: cSettings
+    ),
+    .target(
+        name: "AwsCEventStreams",
```
graebm, line 239: trivial: 
```suggestion
        name: "AwsCEventStream",
```

graebm (approved):
fix & ship

Source/AwsCommonRuntimeKit/event-stream/EventStreamHeader.swift
waahm7: Thanks, good catch.

Source/AwsCommonRuntimeKit/event-stream/EventStreamHeader.swift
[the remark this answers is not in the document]
ganeshnj: In such cases, an `unknown` case with some string message comes handy where customers write logic against. But if we are throwing in C code, makes sense to keep consistent in Swift.

Source/AwsCommonRuntimeKit/crt/Utilities.swift
```diff
@@ -68,17 +78,37 @@ extension aws_date_time {
...
+    init(millisecondsSince1970: Int64) {
+        self = Date(timeIntervalSince1970: TimeInterval(millisecondsSince1970) / 1000)
+    }
 }
 
 extension TimeInterval {
-    var millisecond: UInt64 {
-        UInt64(self*1000)
+    var millisecond: Double {
+        (self*1000).rounded()
```
ganeshnj, line 111: why rounding is required here?

Source/AwsCommonRuntimeKit/event-stream/EventStreamMessage.swift
```diff
@@ -0,0 +1,137 @@
...
+                        rawHeaders,
+                        headerName,
+                        UInt8(header.name.count),
+                        value)
+                case .byteBuf(var value):
+                    if value.count > EventStreamHeader.maxValueLength {
+                        throw CommonRunTimeError.crtError(
+                            .init(
+                                code: AWS_ERROR_EVENT_STREAM_MESSAGE_INVALID_HEADERS_LEN.rawValue))
+                    }
```
ganeshnj, line 88: should this check be in the C code?

So, consumer can throw the C emitted error and avoid duplicating this logic?

Source/AwsCommonRuntimeKit/event-stream/EventStreamMessage.swift
```diff
@@ -0,0 +1,137 @@
...
+    }
+}
+
+extension EventStreamMessage {
+    func addHeader(header: EventStreamHeader, rawHeaders: UnsafeMutablePointer<aws_array_list>) throws {
+        if header.name.count > EventStreamHeader.maxNameLength {
+            throw CommonRunTimeError.crtError(
+                .init(
+                    code: AWS_ERROR_EVENT_STREAM_MESSAGE_INVALID_HEADERS_LEN.rawValue))
+        }
```
ganeshnj, line 49: similar to other check in L84.

Test/AwsCommonRuntimeKitTests/event-stream/EventStreamTests.swift
```diff
@@ -0,0 +1,178 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+
+import XCTest
+import AwsCEventStream
+@testable import AwsCommonRuntimeKit
+
+class EventStreamTests: XCBaseTestCase {
```
ganeshnj, line 8: very nice ❤️

ganeshnj (changes requested):

Source/AwsCommonRuntimeKit/crt/Utilities.swift
waahm7: The C api expects an integer for milliseconds. So we can either do rounding or cast directly to an int.

Source/AwsCommonRuntimeKit/event-stream/EventStreamMessage.swift
waahm7: There are already checks for this in C api. Problem is that it is defined in C as follow.

```
char *value;
int8 value_length;
```
So we need to cast `value.count` to `Int8` before passing to C, which is a crash in Swift if it is greater than Int8.max value. This check is there to make it a runtime error instead of crash.

Source/AwsCommonRuntimeKit/crt/Utilities.swift
ganeshnj: that makes sense, then should type be `Int`?

waahm7: Thanks, fixed.

ganeshnj (approved):

PR #130, opened 2022-12-21
fix: Various fixes to help aws-swift-sdk update to 0.5.0+

epau:
*Description of changes:*
This PR contains various tweaks that make it easier for the aws-sdk-swift repo to update to version 0.5.0+

### DIFF
--- a/Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
+++ b/Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
@@ -70,9 +70,3 @@ extension ByteBuffer: IStreamable {
         return dataSlice.count
     }
 }
-
-extension ByteBuffer {
-    public func encodeToHexString() -> String {
-        data.encodeToHexString()
-    }
-}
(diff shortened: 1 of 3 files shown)


Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
```diff
@@ -72,7 +72,7 @@ extension ByteBuffer: IStreamable {
 }
 
 extension ByteBuffer {
-    func encodeToHexString() -> String {
-        data.map { String(format: "%02x", $0) }.joined()
+    public func encodeToHexString() -> String {
+        data.encodeToHexString()
     }
 }
```
waahm7, line 78: Trivial: I would prefer if we can remove this from ByteBuffer class to make it more simple. We can always do `byteBuffer.getData().encodeToHexString()`

waahm7 (approved):

Source/AwsCommonRuntimeKit/crt/ByteBuffer.swift
epau: sgtm! i'll remove it then!

PR #131, opened 2022-12-30
Feat: Setup CodeBuild Integeration Tests

waahm7:
*Description of changes:*
- Setup CodeBuild and add proxy tests.
- Refactor http tests to have a common http tests fixture.
- Update CI to include submodules check
- Fix OSX CI to fail when unit tests fail


.github/workflows/ci.yml
```diff
@@ -64,3 +65,16 @@ jobs:
           python3 -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz?run=${{ env.RUN }}', 'builder')"
           chmod a+x builder
           ./builder build -p ${{ env.PACKAGE_NAME }} --target=${{ matrix.target }}-armv8
+
+  check-submodules:
+    runs-on: ubuntu-20.04 # latest
+    steps:
+      - name: Checkout Source
+        uses: actions/checkout@v2
```
DmitriyMusatkin, line 73: use latest v3?

might be worth switching to latest ubuntu lts - 22.04 as well

builder.json
```diff
@@ -38,6 +38,7 @@
     },
     "macos": {
       "!test_steps": [
+        "swift test",
```
DmitriyMusatkin, line 41: minor: you can also consider switching to +test_steps instead of !, so you dont have to repeat swift test part

Test/AwsCommonRuntimeKitTests/http/HTTPClientTestFixture.swift
```diff
@@ -0,0 +1,117 @@
...
+        var httpResponse = HTTPResponse()
+        let httpRequestOptions = try getHTTPRequestOptions(
+                method: method,
+                endpoint: endpoint,
+                path: path,
+                body: requestBody,
+                response: &httpResponse)
+
+        // Retry the request multiple times to reduce flakiness
+        for i in 1...3 where httpResponse.statusCode != expectedStatus {
```
DmitriyMusatkin, line 32: where is flakiness coming from? due to use of external servers?

are all tests using external servers? or there are some that mock server and should not be flaky?

.github/workflows/ci.yml
waahm7: Thanks, fixed.

builder.json
waahm7: Thanks, I didn't know about +. I have fixed it.

Test/AwsCommonRuntimeKitTests/http/HTTPClientTestFixture.swift
waahm7: Yes, currently, we are not mocking anything and using external servers like `httpbin.org`.

DmitriyMusatkin: lets make a note or update function name to indicate it will do retries? 
it seems a bit dangerous and someone will likely use it in future for cases where test should fail right away without retrying

waahm7: Thanks, I have added a numRetries parameter to make it more configurable.

DmitriyMusatkin (approved):

Issue #98, opened 2022-11-07
HTTP2


ganeshnj:
### Describe the feature

Support for H2 to support event streaming.

### Use Case

Event Stream.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #132, opened 2023-01-03
Feat: Bind http2

waahm7:
*Issue #:* https://github.com/awslabs/aws-crt-swift/issues/98
*Description of changes:*
- Bind HTTP2 & HTTP2StreamManager


Source/AwsCommonRuntimeKit/http/HTTPVersion.swift
```diff
@@ -0,0 +1,26 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+import AwsCHttp
+
+public enum HTTPVersion {
+    case unknown // Invalid version
+    case version_1_0
```
waahm7, line 7: We can remove this as SDK do not support version 1.0

Source/AwsCommonRuntimeKit/http/HTTPRequest.swift
```diff
@@ -4,7 +4,7 @@ import AwsCHttp
 import AwsCIo
 import AwsCCommon
 
-public class HTTPRequest: HTTPMessage {
+public class HTTPRequest: HTTPRequestBase {
```
waahm7, line 7: I have not renamed this class to avoid breaking the API.

Source/AwsCommonRuntimeKit/http/HTTP1Stream.swift
```diff
@@ -0,0 +1,22 @@
...
+    init(
+        httpConnection: HTTPClientConnection,
+        options: aws_http_make_request_options,
+        callbackData: HTTPStreamCallbackCore) throws {
+        guard let rawValue = withUnsafePointer(
+                to: options, { aws_http_connection_make_request(httpConnection.rawValue, $0) }) else {
+            throw CommonRunTimeError.crtError(.makeFromLastError())
+        }
+        self.httpConnection = httpConnection
+        try super.init(rawValue: rawValue, callbackData: callbackData)
```
graebm, line 20: probably this can't really throw,
but if it can, you probably need to delete the aws_http_stream from the aws_http_connection_make_request() call above

Source/AwsCommonRuntimeKit/http/HTTPClientConnection.swift
```diff
@@ -25,21 +40,27 @@ public class HTTPClientConnection {
...
     /// - Returns: An `HTTPStream` containing the `HTTPClientConnection`
     public func makeRequest(requestOptions: HTTPRequestOptions) throws -> HTTPStream {
         let httpStreamCallbackCore = HTTPStreamCallbackCore(requestOptions: requestOptions)
         do {
-            return try HTTPStream(httpConnection: self,
-                                  options: httpStreamCallbackCore.getRetainedHttpMakeRequestOptions(),
-                                  callbackData: httpStreamCallbackCore)
+            return try HTTP1Stream(httpConnection: self,
+                                   options: httpStreamCallbackCore.getRetainedHttpMakeRequestOptions(),
+                                   callbackData: httpStreamCallbackCore)
```
graebm, line 63: debatable: we could move this to an `HTTP1ClientConnection` class

Source/AwsCommonRuntimeKit/http/HTTPClientConnectionManager.swift
```diff
@@ -22,7 +22,7 @@ public class HTTPClientConnectionManager {
     }
 
     /// Acquires an `HTTPClientConnection` asynchronously.
-    public func acquireConnection() async throws -> HTTPClientConnection {
+    public func acquireConnection() async throws -> HTTPExchange {
```
graebm, line 25: acquireConnection() returns an HTTPExchange but releaseConnection() takes an HTTPClientConnection

stop using exchange for now, or make them both use exchange (my preference is stop using "exchange")

Source/AwsCommonRuntimeKit/http/HTTPRequest.swift
```diff
@@ -45,15 +45,38 @@ public class HTTPRequest: HTTPMessage {
...
-            let iStreamCore = IStreamCore(iStreamable: body, allocator: allocator)
-            aws_http_message_set_body_stream(self.rawValue, &iStreamCore.rawValue)
+public class HTTP2Request: HTTPRequestBase {
+    /// Creates an http2 request which can be passed to a connection.
+    /// - Parameters:
+    ///   - headers: (Optional) headers to send
+    ///   - body: (Optional) body stream to send as part of request
+    ///   - manualDataWrites: Set it to true to indicate body data will be provided over time.
+    ///                       The data can be be supplied via `HTTP2Stream.writeData`.
+    ///                       The last data should be sent with endOfStream as true to complete the stream.
```
graebm, line 67: `manualDataWrites` isn't supposed to be here, right?

Source/AwsCommonRuntimeKit/http/HTTP2StreamManager.swift
```diff
@@ -0,0 +1,81 @@
...
+        acquireStreamCore.continuation.resume(throwing: CommonRunTimeError.crtError(CRTError(code: errorCode)))
+        return
+    }
+
+    // SUCCESS
+    do {
+        let http2Stream = try HTTP2Stream(rawValue: stream!, callbackData: acquireStreamCore.callbackCore)
+        acquireStreamCore.continuation.resume(returning: http2Stream)
+    } catch {
+        acquireStreamCore.continuation.resume(throwing: error)
```
graebm, line 79: can HTTP2Stream's init really fail?

and if it does, do we really need to do fancy stuff to murder the stream, stop future callbacks, etc?

Source/AwsCommonRuntimeKit/http/HTTP2StreamManagerOptions.swift
```diff
@@ -0,0 +1,197 @@
...
+
+    public init(clientBootstrap: ClientBootstrap,
+                hostName: String,
+                port: UInt16,
+                proxyOptions: HTTPProxyOptions? = nil,
+                proxyEnvSettings: HTTPProxyEnvSettings? = nil,
+                socketOptions: SocketOptions = SocketOptions(),
+                tlsOptions: TLSConnectionOptions? = nil,
+                monitoringOptions: HTTPMonitoringOptions? = nil,
+                maxConnections: Int = 2,
```
graebm, line 113: 2 connections max 😳
that seems low
maybe just follow the lead of the C API and force the user to pass it in

Source/AwsCommonRuntimeKit/http/HTTPClientConnectionOptions.swift
```diff
@@ -42,6 +43,47 @@ public struct HTTPClientConnectionOptions: CStructWithShutdownOptions {
 
     public var monitoringOptions: HTTPMonitoringOptions?
 
+    /// (Optional) HTTP/2 specific configuration
+    /// Specify whether you have prior knowledge that cleartext (HTTP) connections are HTTP/2 (RFC-7540 3.4).
+    /// If false, then cleartext connections are treated as HTTP/1.1.
+    /// It is illegal to set this true when secure connections are being used.
+    /// Note that upgrading from HTTP/1.1 to HTTP/2 is not supported (RFC-7540 3.2).
+    public var http2PriorKnowledge: Bool
```
graebm, line 51: ask dengke if we need to add these options to the vanilla connection manager. I'm not sure anyone would use this to create HTTP/2 connections. I think we added these in C so the stream manager could use the connection manager under the hood.

if you can just leave these out, go for it. for all we know we'll be making yet-another connection manager that handles both HTTP/1 and 2 in the next few months

Source/AwsCommonRuntimeKit/http/HTTPStream.swift
```diff
@@ -43,14 +35,27 @@ public class HTTPStream {
...
+        guard !activated else {
+            return
+        }
+
+        lock.lock()
+        defer {
+            lock.unlock()
+        }
+        guard !activated else {
+            return
```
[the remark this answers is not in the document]
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/http/HTTPClientConnectionManager.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/http/HTTPRequest.swift
waahm7: Good catch, this is an outdated comment.

Source/AwsCommonRuntimeKit/http/HTTPClientConnection.swift
waahm7: We will need to create a protocol which have a makeRequest function so that we can call makeRequest on the `HTTPClientConnectionBase` which almost gives us the same solution as HTTPExchange previously. 

We can do it if there can be things that might only apply to HTTP1Connection but not HTTP2Connection given by ConnectionManager.

Source/AwsCommonRuntimeKit/http/HTTP1Stream.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/http/HTTPStream.swift
```diff
@@ -43,14 +35,27 @@ public class HTTPStream {
...
+        }
+
         callbackData.stream = self
```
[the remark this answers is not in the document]
ganeshnj: https://github.com/awslabs/smithy-swift/blob/1fbd5f7bebd024e29c577240a960cd011ce71ac4/Sources/ClientRuntime/Networking/Http/CRT/CRTClientEngine.swift#L114-L138

some logging and processing done on the callback stream.

Source/AwsCommonRuntimeKit/http/HTTPStream.swift
```diff
@@ -43,14 +35,22 @@ public class HTTPStream {
...
     /// Activates the client stream.
     public func activate() throws {
+        lock.lock()
+        defer {
+            lock.unlock()
+        }
+        guard !activated else {
+            return
+        }
+
```
ganeshnj, line 47: no need to lock until required critical section arrives
```suggestion
        guard !activated else {
            return
        }
        lock.lock()
        defer {
            lock.unlock()
        }

```

Source/AwsCommonRuntimeKit/http/HTTPVersion.swift
ganeshnj: mark it as deprecated with message
```suggestion
    @available(*, deprecated, message: "Not supported, use version_1_1")
    case version_1_0
```

ganeshnj: thinking more or can we just not have it, this never existed before right?

Test/AwsCommonRuntimeKitTests/http/HTTP2ClientConnectionTests.swift
```diff
@@ -0,0 +1,184 @@
...
+        let connection = try await connectionManager.acquireConnection()
+        XCTAssertEqual(connection.httpVersion, HTTPVersion.version_2)
+    }
+
+    // Test that the binding works not the actual functionality. C part has tests for functionality
+    func testHTTP2UpdateSetting() async throws {
+        let connectionManager = try await getHttpConnectionManager(endpoint: "httpbin.org", alpnList: ["h2","http/1.1"])
+        let connection = try await connectionManager.acquireConnection()
+        if let connection = connection as? HTTP2ClientConnection {
+            try await connection.updateSetting(setting: HTTP2Settings(enablePush: false))
```
ganeshnj, line 21: there should be an assertion after `updateSetting` call

in rest of the tests too.

Test/AwsCommonRuntimeKitTests/http/HTTP2ClientConnectionTests.swift
```diff
@@ -0,0 +1,184 @@
...
+            XCTFail("Connection is not HTTP2")
+        }
+    }
+
+    // Test that the binding works not the actual functionality. C part has tests for functionality
+    func testHTTP2SendGoAway() async throws {
+        let connectionManager = try await getHttpConnectionManager(endpoint: "httpbin.org", alpnList: ["h2","http/1.1"])
+        let connection = try await connectionManager.acquireConnection()
+        if let connection = connection as? HTTP2ClientConnection {
+          connection.sendGoAway(error: .internalError, allowMoreStreams: false)
```
ganeshnj, line 57: assert?

Test/AwsCommonRuntimeKitTests/http/HTTP2ClientConnectionTests.swift
```diff
@@ -0,0 +1,184 @@
...
+
+        // Parse json body
+        struct Response: Codable {
+            let data: String
+        }
+
+        let body: Response = try! JSONDecoder().decode(Response.self, from: httpResponse.body)
+        XCTAssertEqual(body.data, testBody + TEST_DOC_LINE)
+    }
+}
```
ganeshnj, line 184: could you add one test case for bidi-stream?

graebm (approved):

Source/AwsCommonRuntimeKit/http/HTTP2Settings.swift
```diff
@@ -0,0 +1,72 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+
+import AwsCHttp
+
+/// Predefined settings identifiers (RFC-7540 6.5.2)
+/// Nil means use default values
```
ganeshnj, line 7: strange that Nil will be interpreted as a default value which I think could be something other than `nil`

Source/AwsCommonRuntimeKit/http/HTTP2Stream.swift
```diff
@@ -0,0 +1,85 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+
+import AwsCHttp
+import Foundation
+
+public class HTTP2Stream: HTTPStream {
```
ganeshnj, line 7: /// docs

Source/AwsCommonRuntimeKit/http/HTTP2StreamManager.swift
```diff
@@ -0,0 +1,77 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+import AwsCHttp
+import Collections
+
+public class HTTP2StreamManager {
```
ganeshnj, line 6: /// docs

Source/AwsCommonRuntimeKit/http/HTTPClientConnectionManagerCallbackCore.swift
```diff
@@ -45,7 +45,16 @@ private func onConnectionSetup(connection: UnsafeMutablePointer<aws_http_connect
...
+        callbackDataCore.continuation.resume(
+            returning: HTTP2ClientConnection(
+                manager: callbackDataCore.connectionManager,
+                connection: connection!))
+    } else {
+        callbackDataCore.continuation.resume(
+            returning: HTTPClientConnection(
+                manager: callbackDataCore.connectionManager,
+                connection: connection!))
+    }
```
ganeshnj, line 59: can be simplified
```suggestion
    let connection = HTTP2ClientConnection(
                manager: callbackDataCore.connectionManager,
                connection: connection!)
    switch connection.version {
    ....
    }
```

Source/AwsCommonRuntimeKit/http/HTTPRequest.swift
ganeshnj: /// docs

Source/AwsCommonRuntimeKit/http/HTTPRequest.swift
```diff
@@ -45,15 +45,35 @@ public class HTTPRequest: HTTPMessage {
...
         self.path = path
+        self.body = body
+        addHeaders(headers: headers)
+    }
+}
 
-        if let body = body {
-            let iStreamCore = IStreamCore(iStreamable: body, allocator: allocator)
-            aws_http_message_set_body_stream(self.rawValue, iStreamCore.rawValue)
+public class HTTP2Request: HTTPRequestBase {
```
ganeshnj, line 60: /// docs

Source/AwsCommonRuntimeKit/http/HTTPRequestBase.swift
```diff
@@ -3,7 +3,7 @@
 import AwsCHttp
 import AwsCIo
 
-public class HTTPMessage {
+public class HTTPRequestBase {
```
ganeshnj, line 6: /// docs

Source/AwsCommonRuntimeKit/http/HTTPVersion.swift
```diff
@@ -0,0 +1,26 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+import AwsCHttp
+
+public enum HTTPVersion {
```
ganeshnj, line 5: /// docs

ganeshnj (changes requested):

Source/AwsCommonRuntimeKit/http/HTTPClientConnectionManagerCallbackCore.swift
ganeshnj: switch because when we add H3, we this breaks.

Source/AwsCommonRuntimeKit/http/HTTPStream.swift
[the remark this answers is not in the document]
waahm7: I think it is only used for `makeStatusCode`. We can pass the status code in one of the callback instead.

Source/AwsCommonRuntimeKit/http/HTTP1Stream.swift
```diff
@@ -0,0 +1,22 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+import AwsCHttp
+import Foundation
+
+public class HTTP1Stream: HTTPStream {
```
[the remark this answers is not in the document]
waahm7: In future, we will most probably bind Http 1.1 Chunked Encoding.

Source/AwsCommonRuntimeKit/http/HTTP1Stream.swift
```diff
@@ -0,0 +1,22 @@
...
+    /// Stream keeps a reference to HttpConnection to keep it alive
+    private let httpConnection: HTTPClientConnection
+
+    // Called by HTTPClientConnection
+    init(
+        httpConnection: HTTPClientConnection,
+        options: aws_http_make_request_options,
```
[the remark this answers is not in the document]
waahm7: I think you mean not expose C types. This is an internal function.

Source/AwsCommonRuntimeKit/http/HTTPStream.swift
waahm7: Thread sanitizer might think that this is a race condition. Activated is not supposed to be called multiple times, so this should not be in the hot path.
reference: https://github.com/awslabs/aws-crt-swift/pull/132#discussion_r1072972998

Test/AwsCommonRuntimeKitTests/http/HTTP2ClientConnectionTests.swift
waahm7: I am not sure what do you mean by bidi-stream. This is similar, where we send data using writeData and receive data in the onBody callback until endOfStream is true.

Source/AwsCommonRuntimeKit/http/HTTP2Settings.swift
waahm7: This is a common theme that if you don’t provide a value, the C APIs use the default. We decided not to replicate default values at multiple places because there is a high chance, we won’t update it everywhere if that value is updated in the future. In this case, the default values are mentioned in the RFC.

Test/AwsCommonRuntimeKitTests/http/HTTP2ClientConnectionTests.swift
waahm7: I am not sure what to assert. These tests just check that the bindings work and the C has tests for actual functionality. @TingDaoK Is there something we can assert for in these tests?

Source/AwsCommonRuntimeKit/http/HTTPVersion.swift
waahm7: Thanks, added.

Source/AwsCommonRuntimeKit/http/HTTPRequestBase.swift
waahm7: Thanks, added.

Source/AwsCommonRuntimeKit/http/HTTPRequest.swift
waahm7: Thanks, added.

Test/AwsCommonRuntimeKitTests/http/HTTP2ClientConnectionTests.swift
waahm7: same as above.

Source/AwsCommonRuntimeKit/http/HTTP2Stream.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/http/HTTP2StreamManager.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/http/HTTP1Stream.swift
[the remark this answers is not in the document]
ganeshnj: if internal, can this marked as internal instead public? I guess no because we expose it places which are consumed in other packages.

Source/AwsCommonRuntimeKit/http/HTTPClientConnectionManagerCallbackCore.swift
```diff
@@ -39,13 +39,28 @@ private func onConnectionSetup(connection: UnsafeMutablePointer<aws_http_connect
...
+    case AWS_HTTP_VERSION_2: continuation.resume(
+        returning: HTTP2ClientConnection(
+            manager: callbackDataCore.connectionManager,
+            connection: connection!))
+    case AWS_HTTP_VERSION_1_1:
+        continuation.resume(
+            returning: HTTPClientConnection(
+                manager: callbackDataCore.connectionManager,
+                connection: connection!))
+    default:
```
ganeshnj, line 60: default should be avoided per Swift best practices, you can add unknown here and do the right think. I suppose we want to throw an error or log something.

Source/AwsCommonRuntimeKit/http/HTTP1Stream.swift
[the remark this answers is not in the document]
waahm7: @ganeshnj We expose this class as needed, but the init is internal. So taking a C type directly in this init is not an issue.

Source/AwsCommonRuntimeKit/http/HTTPClientConnectionManagerCallbackCore.swift
waahm7: I don't think we can use unkown default in this case. As per https://docs.swift.org/swift-book/ReferenceManual/Attributes.html#ID620, this attribute is only allowed when compiling with library evolution mode. It is a C enum, even if I handle all the cases, it still says "switch must be exhaustive, remove '@unknown' to handle remaining values". We are already throwing an error.

Source/AwsCommonRuntimeKit/http/HTTPStream.swift
waahm7: Fixed, in #142

Source/AwsCommonRuntimeKit/http/HTTPClientConnectionManagerCallbackCore.swift
ganeshnj: ah my bad, it is C enum. Thanks for pointing that out, all good.

ganeshnj (approved):

Test/AwsCommonRuntimeKitTests/http/HTTP2ClientConnectionTests.swift
ganeshnj: I would have expected a test case where both writeData and onBody used asynchronously.

Test/AwsCommonRuntimeKitTests/http/HTTP2ClientConnectionTests.swift
ganeshnj: In that case your would want to use https://developer.apple.com/documentation/xctest/xctassertnothrow if no error is thrown in updateSetting

Test/AwsCommonRuntimeKitTests/http/HTTP2ClientConnectionTests.swift
waahm7: @ganeshnj They are used both in `testHTTP2StreamUpload`. `nghttp2.org/httpbin/put` is a server which just echoes back whatever we send. So we send data in chunks to `nghttp2.org/httpbin/put` and await stream to complete. We receive data in onBody in a different thread, and onComplete is triggered. At the end we compare the data.

Test/AwsCommonRuntimeKitTests/http/HTTP2ClientConnectionTests.swift
waahm7: It doesn't work with error `'async' call in an autoclosure that does not support concurrency`. The test will fail if anything throws anyhow.

Issue #135, opened 2023-01-06
Failed to convert TimeInterval to UInt64


ganeshnj:
### Describe the bug

```
import AWSS3

@main
public struct TestSdk {
    public private(set) var text = "Hello, World!"

    public static func main() async throws {
        print(TestSdk().text)
        
//         list buckets
        let client = try await S3Client()
        let output = try await client.listBuckets(input: ListBucketsInput())
        print(output)
    }
}

```
Logs
```
Hello, World!
2023-01-06T11:45:24-0800 info DefaultRegionProvider : [Logging] Attempting to resolve region with: BundleRegionProvider
2023-01-06T11:45:24-0800 info DefaultRegionProvider : [Logging] Attempting to resolve region with: EnvironmentRegionProvider
2023-01-06T11:45:24-0800 info DefaultRegionProvider : [Logging] Attempting to resolve region with: ProfileRegionProvider
2023-01-06T11:45:24-0800 info DefaultRegionProvider : [Logging] Resolved region with: ProfileRegionProvider
Swift/arm64e-apple-macos.swiftinterface:32169: Fatal error: Double value cannot be converted to UInt64 because the result would be greater than UInt64.max
2023-01-06 11:45:24.097253-0800 TestSdk[40406:2548657] Swift/arm64e-apple-macos.swiftinterface:32169: Fatal error: Double value cannot be converted to UInt64 because the result would be greater than UInt64.max
(lldb) 
```
Stacktrace
```
Thread 2 Queue : com.apple.root.user-initiated-qos.cooperative (concurrent)
#0	0x00000001916d432c in _swift_runtime_on_report ()
#1	0x00000001917698d8 in _swift_stdlib_reportFatalErrorInFile ()
#2	0x000000019138d1b0 in closure #1 in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#3	0x000000019138cef4 in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#4	0x000000019138ccec in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#5	0x000000019138c84c in _assertionFailure(_:_:file:line:flags:) ()
#6	0x00000001006a19b8 in Credentials.init(accessKey:secret:sessionToken:expiration:allocator:) at /Users/jangirg/Projects/Amplify/SwiftSDK/aws-crt-swift/Source/AwsCommonRuntimeKit/auth/credentials/Credentials.swift:34
[...]

PR #136, opened 2023-01-07
Fix: Credential's Expiration time

waahm7:
Issue #135

*Description of changes:*


graebm:
I agree with translating UINT64_MAX to mean "nil Date", but there are still corner cases here.

The error was:
> Double value cannot be converted to UInt64 because the result would be greater than UInt64.max

It appears Swift is more strict when casting a double to an int that normal C is. Normal C will gladly cast between double and uint64_t, even if the double value is INFINITY. But Swift [does this](https://github.com/apple/swift/blob/193348694fdb5a20e28f55e94ededf7ff40531e9/stdlib/public/core/Integers.swift#L3032-L3060):
>   /// If `source` is outside the bounds of this type after rounding toward
>   /// zero, a runtime error may occur.


So this code in Credentials.swift can still explode:
`expirationTimeout = UInt64(expiration.timeIntervalSince1970)`

graebm (approved):

PR #142, opened 2023-01-19
Remove HTTPStream from HTTP callbacks

waahm7:
*Description of changes:*
- Removes Stream pointer from HTTP callbacks which complicates overall logic due to cyclic dependency without providing much value.

### DIFF
--- a/Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
+++ b/Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
@@ -2,30 +2,36 @@
 //  SPDX-License-Identifier: Apache-2.0.
 import Foundation
 
+/// Definition for outgoing request and callbacks to receive response.
 public struct HTTPRequestOptions {
 
-    public typealias OnInterimResponse = (_ statusCode: Int32,
+    /// Callback to receive interim response
+    public typealias OnInterimResponse = (_ statusCode: UInt32,
                                           _ headers: [HTTPHeader]) -> Void
-    public typealias OnResponse = (_ statusCode: Int32,
+    /// Callback to receive main headers
+    public typealias OnResponse = (_ statusCode: UInt32,
                                    _ headers: [HTTPHeader]) -> Void
-    public typealias OnTrailer = (_ headers: [HTTPHeader]) -> Void
+    /// Callback to receive the incoming body
     public typealias OnIncomingBody = (_ bodyChunk: Data) -> Void
-    public typealias OnStreamComplete = (_ result: Result<Int32, CommonRunTimeError>) -> Void
+    /// Callback to receive trailer headers
+    public typealias OnTrailer = (_ headers: [HTTPHeader]) -> Void
+    /// Callback to know when request is completed, whether successful or unsuccessful
+    public typealias OnStreamComplete = (_ result: Result<UInt32, CommonRunTimeError>) -> Void
 
     /// Outgoing request.
     let request: HTTPRequestBase
 
-    /// Invoked when informational 1xx interim response is received
+    /// Invoked 0+ times if informational 1xx interim responses are received.
     public let onInterimResponse: OnInterimResponse?
 
     /// Invoked when main response headers are received.
     public let onResponse: OnResponse
 
-    /// Invoked when trailer response is received.
-    public let onTrailer: OnTrailer?
-
     /// Invoked repeatedly as body data is received.
     public let onIncomingBody: OnIncomingBody
+
+    /// Invoked when trailer headers are received.
+    public let onTrailer: OnTrailer?
 
     /// Invoked when request/response stream is complete, whether successful or unsuccessful
     public let onStreamComplete: OnStreamComplete
@@ -33,12 +39,12 @@ public struct HTTPRequestOptions {
     /// When using HTTP/2, set http2ManualDataWrites to true to specify that request body data will be provided over time.
     /// The stream will only be polled for writing when data has been supplied via `HTTP2Stream.writeData`
     public var http2ManualDataWrites: Bool = false
-
+    
     public init(request: HTTPRequestBase,
                 onInterimResponse: OnInterimResponse? = nil,
                 onResponse: @escaping OnResponse,
-                onTrailer: OnTrailer? = nil,
                 onIncomingBody: @escaping OnIncomingBody,
+                onTrailer: OnTrailer? = nil,
                 onStreamComplete: @escaping OnStreamComplete,
                 http2ManualDataWrites: Bool = false) {
         self.request = request
--- a/Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
+++ b/Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
@@ -3,6 +3,14 @@
 import Foundation
 
 public struct HTTPRequestOptions {
+
+    public typealias OnInterimResponse = (_ statusCode: Int32,
+                                          _ headers: [HTTPHeader]) -> Void
+
+    public typealias OnResponse = (_ statusCode: Int32,
+                                   _ headers: [HTTPHeader]) -> Void
+    public typealias OnTrailer = (_ headers: [HTTPHeader]) -> Void
+
     public typealias OnIncomingHeaders = (_ statusCode: Int32,
                                           _ headerBlock: HTTPHeaderBlock,
                                           _ headers: [HTTPHeader]) -> Void
@@ -13,11 +21,14 @@ public struct HTTPRequestOptions {
     /// Outgoing request.
     let request: HTTPRequestBase
 
-    /// Invoked repeatedly as headers are received.
-    public let onIncomingHeaders: OnIncomingHeaders
+    /// Invoked when informational 1xx interim response is received
+    public let onInterimResponse: OnInterimResponse?
 
-    /// Invoked when response header block has been completely read.
-    public let onIncomingHeadersBlockDone: OnIncomingHeadersBlockDone
+    /// Invoked when main response headers are received.
+    public let onResponse: OnResponse
+
+    /// Invoked when trailer response headers are received.
+    public let onTrailer: OnTrailer?
 
     /// Invoked repeatedly as body data is received.
     public let onIncomingBody: OnIncomingBody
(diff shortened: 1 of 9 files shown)


Source/AwsCommonRuntimeKit/http/HTTPStreamCallbackCore.swift
```diff
@@ -66,9 +72,7 @@ private func onResponseHeaderBlockDone(stream: UnsafeMutablePointer<aws_http_str
                                        headerBlock: aws_http_header_block,
                                        userData: UnsafeMutableRawPointer!) -> Int32 {
     let httpStreamCbData = Unmanaged<HTTPStreamCallbackCore>.fromOpaque(userData).takeUnretainedValue()
-    let stream = httpStreamCbData.stream!
-
-    httpStreamCbData.requestOptions.onIncomingHeadersBlockDone(stream, HTTPHeaderBlock(rawValue: headerBlock))
+    httpStreamCbData.requestOptions.onIncomingHeadersBlockDone(HTTPHeaderBlock(rawValue: headerBlock))
```
graebm, line 75: probably want to pass status in `onIncomingHeadersBlockDone` too. It's theoretically possible there are 0 headers, and so the onIncomingHeaders callback never fires

OR, while you're messing with the API, make it so there's only 1 `onIncomingHeaders` callback in Swift. Use the C on_incoming_headers callback to buffer up a list, then use the C on_incoming_headers_block_done callback to deliver those buffered up headers to Swift.

OR EVEN BETTER, still do the buffering thing, but have 3 separate callbacks: `onInterimResponse` (for 1xx INFORMATIONAL headers and status code), onResponse (for normal headers and status code), and onTrailer (for trailing headers (no status code)). This is how I wished I'd built the C APIs in retrospect. The scary thing with all different kinds of headers using the same callback is: if the user doesn't check the block type, their code will get real messed up if they ever get an informational 1xx interim response

That 3rd approach is *kind of* what we have In python ([python API](https://github.com/awslabs/aws-crt-python/blob/10d3900c2e88dfb491c9b93b75d47024685f9f66/awscrt/http.py#L177), [C code](https://github.com/awslabs/aws-crt-python/blob/10d3900c2e88dfb491c9b93b75d47024685f9f66/source/http_stream.c#L127-L137)). Except we never actually hooked up the `on_interim_response` and `on_trailer` callbacks, we only have the 1 `on_response` callback right now.

Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
```diff
@@ -3,22 +3,26 @@
...
+    public typealias OnInterimResponse = (_ statusCode: Int32,
                                           _ headers: [HTTPHeader]) -> Void
-    public typealias OnIncomingHeadersBlockDone = (_ stream: HTTPStream,
-                                                   _ headerBlock: HTTPHeaderBlock) -> Void
-    public typealias OnIncomingBody = (_ stream: HTTPStream, _ bodyChunk: Data) -> Void
-    public typealias OnStreamComplete = (_ stream: HTTPStream, _ error: CRTError?) -> Void
+    public typealias OnResponse = (_ statusCode: Int32,
+                                   _ headers: [HTTPHeader]) -> Void
+    public typealias OnTrailer = (_ headers: [HTTPHeader]) -> Void
+    public typealias OnIncomingBody = (_ bodyChunk: Data) -> Void
```
graebm, line 12: utterly trivial: declare OnIncomingBody before OnTrailer, since that's the body arrives before the trailer

Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
```diff
@@ -3,22 +3,26 @@
...
+    /// Invoked when main response headers are received.
+    public let onResponse: OnResponse
 
-    /// Invoked when response header block has been completely read.
-    public let onIncomingHeadersBlockDone: OnIncomingHeadersBlockDone
+    /// Invoked when trailer response is received.
+    public let onTrailer: OnTrailer?
 
     /// Invoked repeatedly as body data is received.
     public let onIncomingBody: OnIncomingBody
```
graebm, line 28: again, declare OnIncomingBody before OnTrailer. just because that's the order these events occur

Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
```diff
@@ -3,22 +3,26 @@
...
+    public typealias OnIncomingBody = (_ bodyChunk: Data) -> Void
+    public typealias OnStreamComplete = (_ result: Result<Int32, CommonRunTimeError>) -> Void
 
     /// Outgoing request.
     let request: HTTPRequestBase
 
-    /// Invoked repeatedly as headers are received.
-    public let onIncomingHeaders: OnIncomingHeaders
+    /// Invoked when informational 1xx interim response is received
+    public let onInterimResponse: OnInterimResponse?
```
graebm, line 19: /// Invoked 0+ times if informational 1xx interim responses are received

Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
```diff
@@ -3,22 +3,26 @@
...
+
+    /// Invoked when main response headers are received.
+    public let onResponse: OnResponse
 
-    /// Invoked when response header block has been completely read.
-    public let onIncomingHeadersBlockDone: OnIncomingHeadersBlockDone
+    /// Invoked when trailer response is received.
```
graebm, line 24: ~trailer response~ -> trailer
or "trailing headers"?

Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
```diff
@@ -31,14 +35,16 @@ public struct HTTPRequestOptions {
     public var http2ManualDataWrites: Bool = false
 
     public init(request: HTTPRequestBase,
-                onIncomingHeaders: @escaping OnIncomingHeaders,
-                onIncomingHeadersBlockDone: @escaping OnIncomingHeadersBlockDone,
+                onInterimResponse: OnInterimResponse? = nil,
+                onResponse: @escaping OnResponse,
+                onTrailer: OnTrailer? = nil,
                 onIncomingBody: @escaping OnIncomingBody,
```
graebm, line 41: again, body before trailer

graebm (approved):
beautiful

Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
```diff
@@ -3,22 +3,26 @@
 import Foundation
 
 public struct HTTPRequestOptions {
-    public typealias OnIncomingHeaders = (_ stream: HTTPStream,
-                                          _ headerBlock: HTTPHeaderBlock,
+
+    public typealias OnInterimResponse = (_ statusCode: Int32,
```
ganeshnj, line 7: should be UInt32
```suggestion
    public typealias OnInterimResponse = (_ statusCode: UInt32,
```

Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
```diff
@@ -3,22 +3,26 @@
...
                                           _ headers: [HTTPHeader]) -> Void
-    public typealias OnIncomingHeadersBlockDone = (_ stream: HTTPStream,
-                                                   _ headerBlock: HTTPHeaderBlock) -> Void
-    public typealias OnIncomingBody = (_ stream: HTTPStream, _ bodyChunk: Data) -> Void
-    public typealias OnStreamComplete = (_ stream: HTTPStream, _ error: CRTError?) -> Void
+    public typealias OnResponse = (_ statusCode: Int32,
```
ganeshnj, line 9:
```suggestion
    public typealias OnResponse = (_ statusCode: UInt32,
```

Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
ganeshnj: also /// docs

Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
ganeshnj: also /// docs

Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
```diff
@@ -3,22 +3,26 @@
...
+                                   _ headers: [HTTPHeader]) -> Void
+    public typealias OnTrailer = (_ headers: [HTTPHeader]) -> Void
```
ganeshnj, line 11: /// docs

Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
ganeshnj: /// docs

Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
```diff
@@ -3,22 +3,26 @@
...
+    public typealias OnIncomingBody = (_ bodyChunk: Data) -> Void
+    public typealias OnStreamComplete = (_ result: Result<Int32, CommonRunTimeError>) -> Void
```
ganeshnj, line 13: /// docs

ganeshnj (commented):
LGTM, minor comment on adding swift docs for public APIs and constructs.

Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/http/HTTPRequestOptions.swift
waahm7: Thanks, fixed.

Issue #127, opened 2022-12-21
Expose `generateRandom` as a non-c closure


epau:
### Describe the feature

Currently, `generateRandom` is exposed as a `@convention(c)` closure. Which requires this implementation detail to be exposed all the way to the consumer (due to the constraints of c closures).

Can we expose a regular swift closure here instead?

### Use Case

N/A

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #147, opened 2023-01-23
Fix generate random callback

waahm7:
Issue #127 

*Description of changes:*
- Expose generateRandom as a non-C closure after relevant changes in C
- A new pattern for to retain callbacks until shutdown callback has fired which doesn't require double retain/release.

### DIFF
--- a/Source/AwsCommonRuntimeKit/crt/Utilities.swift
+++ b/Source/AwsCommonRuntimeKit/crt/Utilities.swift
@@ -5,6 +5,17 @@ import struct Foundation.Data
 import struct Foundation.TimeInterval
 import AwsCCal
 
+class Box<T> {
+    let contents: T
+    init(_ contents: T) {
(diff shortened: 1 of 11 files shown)


Source/AwsCommonRuntimeKit/crt/Utilities.swift
```diff
@@ -5,6 +5,17 @@ import struct Foundation.Data
 import struct Foundation.TimeInterval
 import AwsCCal
 
+class Box<T> {
```
graebm, line 8: comment what this is used for

graebm (approved):

Source/AwsCommonRuntimeKit/crt/Utilities.swift
waahm7: Thanks, fixed.

Issue #150, opened 2023-01-26
Chunked signer


ganeshnj:
### Describe the feature

Event stream messages are signed one by one hence Swift SDK needs a chunked signer.

### Use Case

Event Streams

### Proposed Solution

There is already signRequest method in Signer
```swift
    public static func signRequest(
        request: HTTPRequestBase,
        config: SigningConfig,
        allocator: Allocator = defaultAllocator) async throws -> HTTPRequestBase {
```
The new API should be something like
```swift
    func signChunk(chunk: Data, previousSignature: Data, config: SigningConfig) -> Data {
        
    }
}
```

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #151, opened 2023-01-30
Feat: Chunk Signer

waahm7:
Issue #150 

*Description of changes:*
- Adds chunk signer

### DIFF
--- a/Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
+++ b/Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
@@ -30,9 +30,9 @@ public class Signer {
     /// - `Throws`: An error of type `AwsCommonRuntimeError` which will pull last error found in the CRT
     /// - `Returns`: Returns a signed http request `HttpRequest`
     public static func signRequest(
-        request: HTTPRequest,
+        request: HTTPRequestBase,
         config: SigningConfig,
-        allocator: Allocator = defaultAllocator) async throws -> HTTPRequest {
+        allocator: Allocator = defaultAllocator) async throws -> HTTPRequestBase {
 
         guard let signable = aws_signable_new_http_request(allocator.rawValue, request.rawValue) else {
             throw CommonRunTimeError.crtError(.makeFromLastError())
@@ -41,7 +41,9 @@ public class Signer {
             aws_signable_destroy(signable)
         }
 
-        return try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<HTTPRequest, Error>) in
+        return try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<
+                                                                HTTPRequestBase,
+                                                                Error>) in
             let signRequestCore = SignRequestCore(request: request,
                                                   continuation: continuation,
                                                   shouldSignHeader: config.shouldSignHeader,
@@ -74,11 +76,11 @@ public class Signer {
 
 class SignRequestCore {
     let allocator: Allocator
-    let request: HTTPRequest
-    var continuation: CheckedContinuation<HTTPRequest, Error>
+    let request: HTTPRequestBase
+    var continuation: CheckedContinuation<HTTPRequestBase, Error>
     let shouldSignHeader: ((String) -> Bool)?
-    init(request: HTTPRequest,
-         continuation: CheckedContinuation<HTTPRequest, Error>,
+    init(request: HTTPRequestBase,
+         continuation: CheckedContinuation<HTTPRequestBase, Error>,
          shouldSignHeader: ((String) -> Bool)? = nil,
          allocator: Allocator) {
         self.allocator = allocator
--- a/Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
+++ b/Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
@@ -64,7 +64,7 @@ public class Signer {
                         onSigningComplete,
                         signRequestCore.passRetained())
                         != AWS_OP_SUCCESS {
-                        
+
                         signRequestCore.release()
                         continuation.resume(throwing: CommonRunTimeError.crtError(.makeFromLastError()))
                     }
--- a/Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
+++ b/Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
@@ -54,16 +54,16 @@ public class Signer {
             }
             config.withCPointer(userData: shouldSignHeaderUserData) { configPointer in
                 configPointer.withMemoryRebound(
-                        to: aws_signing_config_base.self,
-                        capacity: 1) { configBasePointer in
+                    to: aws_signing_config_base.self,
+                    capacity: 1) { configBasePointer in
 
                     if aws_sign_request_aws(
-                            allocator.rawValue,
-                            signable,
-                            configBasePointer,
-                            onSigningComplete,
-                            signRequestCore.passRetained())
-                            
+                        allocator.rawValue,
+                        signable,
+                        configBasePointer,
+                        onSigningComplete,
+                        signRequestCore.passRetained())
+
                         != AWS_OP_SUCCESS {
                         signRequestCore.release()
                         continuation.resume(throwing: CommonRunTimeError.crtError(.makeFromLastError()))
--- a/Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
+++ b/Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
@@ -63,8 +63,8 @@ public class Signer {
                         configBasePointer,
                         onSigningComplete,
                         signRequestCore.passRetained())
-
                         != AWS_OP_SUCCESS {
+                        
                         signRequestCore.release()
                         continuation.resume(throwing: CommonRunTimeError.crtError(.makeFromLastError()))
                     }
--- a/Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
+++ b/Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
@@ -61,7 +61,7 @@ public class Signer {
                         allocator.rawValue,
                         signable,
                         configBasePointer,
-                        onSigningComplete,
+                        onRequestSigningComplete,
                         signRequestCore.passRetained())
                         != AWS_OP_SUCCESS {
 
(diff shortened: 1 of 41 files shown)


Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
```diff
@@ -72,6 +72,105 @@ public class Signer {
...
+    /// - Returns: Signing Result
+    /// - Throws: CommonRunTimeError.crtError
+    public static func signTrailerHeaders(headers: [HTTPHeader],
+                                          previousSignature: String,
+                                          config: SigningConfig,
+                                          allocator: Allocator = defaultAllocator) async throws -> String {
+
+        guard let signable = previousSignature.withByteCursorPointer({ previousSignatureCursor in
+            headers.withCHeaders(allocator: allocator) { cHeaders in
+                aws_signable_new_trailing_headers(allocator.rawValue, cHeaders, previousSignatureCursor.pointee)
```
graebm, line 118: I'm noticing that in C, it takes `struct aws_http_headers *trailing_headers`, but not `const`
isn't that usually an issue when binding to Swift?

Source/AwsCommonRuntimeKit/http/HTTPHeader.swift
```diff
@@ -28,3 +28,15 @@ public class HTTPHeader: CStruct {
         }
     }
 }
+
+extension Array where Element == HTTPHeader {
+    func withCHeaders<Result>(allocator: Allocator,
+                              _ body: (OpaquePointer) -> Result) -> Result {
+        var cHeaders: OpaquePointer = aws_http_headers_new(allocator.rawValue)
```
graebm, line 35: leak? don't we need an `aws_http_headers_release()` somewhere?

and if this is a leak, why didn't tests catch it?

Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
```diff
@@ -72,6 +72,105 @@ public class Signer {
...
+
+    private static func sign(config: SigningConfig,
+                             signable: UnsafePointer<aws_signable>,
+                             allocator: Allocator) async throws -> String {
+
+        try await withCheckedThrowingContinuation { continuation in
+            config.withCPointer { configPointer in
+                configPointer.withMemoryRebound(to: aws_signing_config_base.self,
+                                                capacity: 1) { configBasePointer in
+                    let chunkSignerCore = ChunkSignerCore(continuation: continuation,
```
graebm, line 137: is `chunkSignerCore` still a good name, given that we're using it for TrailerHeaders now too?

same for `onChunkSigningComplete` a few lines down

Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
waahm7: This is not an issue because we are getting a mutable opaque pointer from C using `aws_http_headers_new` and passing that to C. I have added const anyhow since it is a good practice.

Source/AwsCommonRuntimeKit/http/HTTPHeader.swift
waahm7: Thanks, that is a great catch. The tests didn't catch it because I forgot to pass tracing alocator to sign chunk call. I have fixed and created a corresponding fix in aws-c-auth.

Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
```diff
@@ -72,6 +72,105 @@ public class Signer {
...
+    /// Signs a body chunk according to the supplied signing configuration
+    /// - Parameters:
+    ///   - chunk: Chunk to sign
+    ///   - previousSignature: The signature of the previous component of the request: either the request itself for the first chunk,
+    ///                        or the previous chunk otherwise.
+    ///    - config: The `SigningConfig` to use when signing.
+    ///   - allocator: (Optional) allocator to override
+    /// - Returns: Signature of the chunk
+    /// - Throws: CommonRunTimeError.crtError
+    public static func signChunk(chunk: Data,
```
ganeshnj, line 85: I expect the signature to be a byte array here ie `Data`

same for return type.

Any reason choosing it as a String?

ganeshnj: Also, how would it work if we done't have previous signature. I think previous signature must be optional here.

ganeshnj (changes requested):

Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
ganeshnj: Yes, checked that.

With Transcribe Streaming also, request signature works are previous signature. Feel free to ignore https://github.com/awslabs/aws-crt-swift/pull/151#discussion_r1093602179.

waahm7: For signature, I choosed a string because it matched the C type. I had updated it to Data.

Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
waahm7: I have updated it and removed `chunkSignerCore` as I realized we don’t really need to do any allocation.

Source/AwsCommonRuntimeKit/http/HTTPHeader.swift
```diff
@@ -28,3 +28,18 @@ public class HTTPHeader: CStruct {
...
+    func withCHeaders<Result>(allocator: Allocator,
+                              _ body: (OpaquePointer) -> Result) -> Result {
+        let cHeaders: OpaquePointer = aws_http_headers_new(allocator.rawValue)
+        defer {
+            aws_http_headers_release(cHeaders)
+        }
+        forEach { $0.withCPointer { guard aws_http_headers_add_header(cHeaders, $0) == AWS_OP_SUCCESS else {
+            let error = CRTError.makeFromLastError()
+            fatalError("Unable to add header due to error code: \(error.code) message:\(error.message)")
+        }} }
```
ganeshnj, line 42: trivial: line breaks would make it readable
```suggestion
        forEach { 
            $0.withCPointer { 
                guard aws_http_headers_add_header(cHeaders, $0) == AWS_OP_SUCCESS else {
                    let error = CRTError.makeFromLastError()
                    fatalError("Unable to add header due to error code: \(error.code) message:\(error.message)")
                }
            }
        }
```

ganeshnj (approved):

Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
graebm: I disagree with this change. I'm pretty sure the returned signature should be `String`, and the previousSignature should also be `String`. If it were just `Data` then the user would be like ... am I supposed to base64 encode this or what?

Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
graebm: OK. related: I kinda wish our ref-counting functions lied and took `const` pointers, even though they do mutate the refcount, because of cases like this, where we're not REALLY mutating the headers, we just need to be sure they don't get deleted

Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
```diff
@@ -72,6 +72,82 @@ public class Signer {
...
+    ///   - previousSignature: The signature of the previous component of the request: either the request itself for the first chunk,
+    ///                        or the previous chunk otherwise.
+    ///   - config: The `SigningConfig` to use when signing.
+    ///   - allocator: (Optional) allocator to override
+    /// - Returns: Signing Result
+    /// - Throws: CommonRunTimeError.crtError
+    public static func signTrailerHeaders(headers: [HTTPHeader],
+                                          previousSignature: Data,
+                                          config: SigningConfig,
+                                          allocator: Allocator = defaultAllocator) async throws -> Data {
```
graebm, line 114:
```suggestion
                                          previousSignature: String,
                                          config: SigningConfig,
                                          allocator: Allocator = defaultAllocator) async throws -> String {
```

Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
```diff
@@ -72,6 +72,82 @@ public class Signer {
...
+    ///   - previousSignature: The signature of the previous component of the request: either the request itself for the first chunk,
+    ///                        or the previous chunk otherwise.
+    ///    - config: The `SigningConfig` to use when signing.
+    ///   - allocator: (Optional) allocator to override
+    /// - Returns: Signature of the chunk
+    /// - Throws: CommonRunTimeError.crtError
+    public static func signChunk(chunk: Data,
+                                 previousSignature: Data,
+                                 config: SigningConfig,
+                                 allocator: Allocator = defaultAllocator) async throws -> Data {
```
graebm, line 88:
```suggestion
                                 previousSignature: String,
                                 config: SigningConfig,
                                 allocator: Allocator = defaultAllocator) async throws -> String {
```

Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
ganeshnj: I expected it to be `Data` from API point of view.

@graebm are you suggesting this for consistency among other language bindings?

To some degree I don't like converting String to Data if the underlying implementation doesn't work on bytes array. We are wasting CPU cycles here (not significant but still).

Thoughts?

ganeshnj: Didn't see https://github.com/awslabs/aws-crt-swift/pull/151#discussion_r1093877914

Happy to go with String.

Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
ganeshnj: I believe, consumers don't have to do anything with String, we can treat it as immutable during the whole request time.

Source/AwsCommonRuntimeKit/auth/signing/Signer.swift
waahm7: Thanks, I have updated it to String.

graebm (approved):

PR #154, opened 2023-02-04
Override Default Allocator

waahm7:
*Description of changes:*
- Always override defaultAllocator with tracingAllocator in tests. I would have preferred to make it internal, but it needs to be public to override in tests `onSetup`.
- Add documentation for `CommonRuntimeKit` and warn users about the dangers of cleanup.

### DIFF
--- a/Source/AwsCommonRuntimeKit/CommonRuntimeKit.swift
+++ b/Source/AwsCommonRuntimeKit/CommonRuntimeKit.swift
@@ -2,7 +2,7 @@ import AwsCEventStream
 import AwsCAuth
 
 /**
- * Used to initialize the library.
+ * Initializes the library.
  * `CommonRuntimeKit.initialize` must be called before using any other functionality.
  */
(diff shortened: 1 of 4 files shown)


Source/AwsCommonRuntimeKit/CommonRuntimeKit.swift
```diff
@@ -1,13 +1,36 @@
...
+    /**
+     * This is an alternative constructor mainly for testing.
+     * Use this instead if you want to override defaultAllocator.
+     */
+    public static func initialize(customDefaultAllocator: Allocator) {
+        aws_auth_library_init(customDefaultAllocator.rawValue)
+        aws_event_stream_library_init(customDefaultAllocator.rawValue)
+        defaultAllocator = customDefaultAllocator.rawValue
+    }
+
```
graebm, line 28: have one initialize() function call the other to cut down on copy/paste

Issue #143, opened 2023-01-23
Host resolver bindings


ganeshnj:
### Describe the feature

We need following APIs for Smithy DNS ref arch
```
enum HostAddressType {
    // IPv6
    AAAA,
    // IPv4
    A,
}
```
```
struct HostAddress {
    // address type (ipv6, ipv4 etc...)
    HostAddressType addressType;
    // the resolved numerical address represented as a String
    // passing this string to pton(), for example, 
    // should correctly return the numerical representation
    String address;
    // name that address was resolved from
    String hostName;
    // service record. Currently unused largely because we use Http, but this may
    // change as we add more protocols.
    String? service;
}
```
```
struct HostResolverArguments {
    // Host name to resolve
    String hostName;
    // service record. Currently unused largely because we use Http, but this may
    // change as we add more protocols.
    String? service;
}
```
```
interface HostResolver {
    // Resolves the address(es) for HostResolverArguments and returns a list of 
    // addresses with (most likely) two addresses, one AAAA and one A. Calls to this
    // function will likely alter the cache so that if there's multiple addresses, 
    // a different set will be returned on the next call.
    // implemetations don't have to explictly call getaddrinfo(), they can use
    // high level abstractions provided in their language runtimes/libraries.
    Result<List<HostAddress>> resolveAddress(HostResolverArguments args) async;
    // Reports a failure on an address so that the background cache can accomodate
    // the failure and likely not return the address until it recovers.
    void reportFailureOnAddress(HostAddress addr);
    // Empties the cache for an address.
    void purgeCache(HostResolverArguments args);
    // Empties the cache for all addresses.
    void purgeCache();
}
```

### Use Case

Smithy Reference Architecture

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #162, opened 2023-02-17
Host Resovler Arch Update

waahm7:
Issue #143 

*Description of changes:*
- Updates host resolver architecture
- Adds recordConnectionOnFailure, purgeCache, purgeCacheAddress

### DIFF
--- a/Source/AwsCommonRuntimeKit/io/HostAddressType.swift
+++ b/Source/AwsCommonRuntimeKit/io/HostAddressType.swift
@@ -3,21 +3,21 @@
 
 import AwsCIo
 
-public enum AddressRecordType {
-    case typeA
-    case typeAAAA
+public enum HostAddressType {
+    case A
+    case AAAA
 }
 
-extension AddressRecordType: RawRepresentable, CaseIterable {
+extension HostAddressType: RawRepresentable, CaseIterable {
 
     public init(rawValue: aws_address_record_type) {
         let value = Self.allCases.first { $0.rawValue == rawValue }
-        self = value ?? .typeA
+        self = value ?? .A
     }
     public var rawValue: aws_address_record_type {
         switch self {
-        case .typeA:  return aws_address_record_type(rawValue: 0)
-        case .typeAAAA:  return aws_address_record_type(rawValue: 1)
+        case .A:  return aws_address_record_type(rawValue: 0)
+        case .AAAA:  return aws_address_record_type(rawValue: 1)
         }
     }
 }
--- a/Source/AwsCommonRuntimeKit/io/HostAddressType.swift
+++ b/Source/AwsCommonRuntimeKit/io/HostAddressType.swift
@@ -3,6 +3,7 @@
 
 import AwsCIo
 
+/// Type of Host Address (ipv4 or ipv6)
 public enum HostAddressType {
     case A
     case AAAA
--- a/Source/AwsCommonRuntimeKit/io/HostResolver.swift
+++ b/Source/AwsCommonRuntimeKit/io/HostResolver.swift
@@ -83,7 +83,7 @@ public class HostResolver: HostResolverProtocol {
     }
 
     public func reportFailureOnAddress(address: HostAddress) throws {
-        guard address.withCPointer ({ cAddress in
+        guard address.withCPointer({ cAddress in
             aws_host_resolver_record_connection_failure(rawValue, cAddress)
         }) == AWS_OP_SUCCESS else {
             throw CommonRunTimeError.crtError(.makeFromLastError())
--- a/Source/AwsCommonRuntimeKit/io/HostResolver.swift
+++ b/Source/AwsCommonRuntimeKit/io/HostResolver.swift
@@ -108,6 +108,7 @@ public class HostResolver: HostResolverProtocol {
             purgeCacheOptions.user_data = continuationCore.passRetained()
             guard aws_host_resolver_purge_host_cache(rawValue, &purgeCacheOptions) == AWS_OP_SUCCESS else {
                 continuationCore.release()
+                continuation.resume()
                 return
             }
         })
@@ -125,6 +126,7 @@ public class HostResolver: HostResolverProtocol {
                     continuationCore.passRetained()) == AWS_OP_SUCCESS
             else {
                 continuationCore.release()
+                continuation.resume()
                 return
             }
         })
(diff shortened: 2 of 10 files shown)


Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -4,11 +4,27 @@
...
+    /// Resolves the address(es) for HostResolverArguments and returns a list of
+    /// addresses with (most likely) two addresses, one AAAA and one A.
+    func resolveAddress(args: HostResolverArguments) async throws -> [HostAddress]
+    /// Reports a failure on an address so that the background cache can accommodate
+    /// the failure and likely not return the address until it recovers.
+    func reportFailureOnAddress(address: HostAddress) throws
+    /// Empties the cache for an address.
+    func purgeCache(args: HostResolverArguments) async throws
+    /// Empties the cache for all addresses.
+    func purgeCache() async throws
```
graebm, line 18: doesn't seem like these last 3 functions should be throwing.

like, how could the user meaningfully respond to any of these failures

Oh, the underlying C functions can report an error.
Maybe this is a good time to change that

waahm7: Thanks, I have updated the Swift API to ignore the error.

graebm (approved):

Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -59,15 +78,58 @@ public class HostResolver {
...
+        let host = AWSString(args.hostName, allocator: allocator)
+        return await withCheckedContinuation({ (continuation: CheckedContinuation<(), Never>) in
+            let continuationCore = Box(continuation)
+            var purgeCacheOptions = aws_host_resolver_purge_host_options()
+            purgeCacheOptions.host = UnsafePointer(host.rawValue)
+            purgeCacheOptions.on_host_purge_complete_callback = onPurgeCacheComplete
+            purgeCacheOptions.user_data = continuationCore.passRetained()
+            guard aws_host_resolver_purge_host_cache(rawValue, &purgeCacheOptions) == AWS_OP_SUCCESS else {
+                continuationCore.release()
+                return
```
graebm, line 111: forgot to resume continuation on failure

waahm7: Good catch, fixed.

Source/AwsCommonRuntimeKit/io/HostAddressType.swift
```diff
@@ -3,21 +3,22 @@
...
+    case AAAA
 }
 
-extension AddressRecordType: RawRepresentable, CaseIterable {
+extension HostAddressType: RawRepresentable, CaseIterable {
 
     public init(rawValue: aws_address_record_type) {
         let value = Self.allCases.first { $0.rawValue == rawValue }
-        self = value ?? .typeA
+        self = value ?? .A
```
ganeshnj, line 16: not related to this PR but we should make these type of init throwable.

Falling back on .A is bad for future work, for example if we have IPv8.

Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -4,11 +4,27 @@
 import AwsCCommon
 import AwsCIo
 
-public class HostResolver {
+/// All host resolver implementation must conform to this protocol.
+public protocol HostResolverProtocol {
+    /// Resolves the address(es) for HostResolverArguments and returns a list of
+    /// addresses with (most likely) two addresses, one AAAA and one A.
+    func resolveAddress(args: HostResolverArguments) async throws -> [HostAddress]
```
ganeshnj, line 11: use Swift Doc than just plain ///

https://github.com/apple/swift/blob/main/docs/DocumentationComments.md

Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -59,15 +78,60 @@ public class HostResolver {
...
+            let continuationCore = Box(continuation)
+            var purgeCacheOptions = aws_host_resolver_purge_host_options()
+            purgeCacheOptions.host = UnsafePointer(host.rawValue)
+            purgeCacheOptions.on_host_purge_complete_callback = onPurgeCacheComplete
+            purgeCacheOptions.user_data = continuationCore.passRetained()
+            guard aws_host_resolver_purge_host_cache(rawValue, &purgeCacheOptions) == AWS_OP_SUCCESS else {
+                continuationCore.release()
+                continuation.resume()
+                return
+            }
```
ganeshnj, line 113: if guard's else is not executed, we are not resuming continuation. There must be at least one resume.

Source/AwsCommonRuntimeKit/io/HostResolver.swift
```diff
@@ -59,15 +78,60 @@ public class HostResolver {
...
+            guard aws_host_resolver_purge_cache_with_callback(
+                    rawValue,
+                    onPurgeCacheComplete,
+                    continuationCore.passRetained()) == AWS_OP_SUCCESS
+            else {
+                continuationCore.release()
+                continuation.resume()
+                return
+            }
+        })
```
ganeshnj, line 132: Shouldn't there be a resume as well?

ganeshnj (changes requested):

Source/AwsCommonRuntimeKit/io/HostResolver.swift
waahm7: It is in the callback.

Source/AwsCommonRuntimeKit/io/HostResolver.swift
waahm7: It is in the callback.

Source/AwsCommonRuntimeKit/io/HostAddressType.swift
waahm7: Yeah, we should refactor all enums to a better pattern.

Source/AwsCommonRuntimeKit/io/HostResolver.swift
waahm7: Thanks, fixed.

ganeshnj (approved):

PR #165, opened 2023-02-23
refactor: streams

ganeshnj:
*Issue #, if available:*

*Description of changes:*

Companion for https://github.com/awslabs/smithy-swift/pull/530

### DIFF
--- a/Source/AwsCommonRuntimeKit/http/HTTPStreamCallbackCore.swift
+++ b/Source/AwsCommonRuntimeKit/http/HTTPStreamCallbackCore.swift
@@ -94,8 +94,13 @@ private func onResponseBody(stream: UnsafeMutablePointer<aws_http_stream>?,
 
     let incomingBodyFn = httpStreamCbData.requestOptions.onIncomingBody
     let callbackBytes = Data(bytesNoCopy: bufPtr, count: bufLen, deallocator: .none)
-    incomingBodyFn(callbackBytes)
-    return AWS_OP_SUCCESS
+
+    do {
+        try incomingBodyFn(callbackBytes)
+        return AWS_OP_SUCCESS
+    } catch {
+        return AWS_OP_ERR
+    }
 }
 
 private func onComplete(stream: UnsafeMutablePointer<aws_http_stream>?,
@@ -105,7 +110,11 @@ private func onComplete(stream: UnsafeMutablePointer<aws_http_stream>?,
     let httpStreamCbData = Unmanaged<HTTPStreamCallbackCore>.fromOpaque(userData).takeUnretainedValue()
     let onStreamCompleteFn = httpStreamCbData.requestOptions.onStreamComplete
     guard errorCode == AWS_OP_SUCCESS else {
-        onStreamCompleteFn(.failure(CommonRunTimeError.crtError(CRTError(code: errorCode))))
+        do {
+            try onStreamCompleteFn(.failure(CommonRunTimeError.crtError(CRTError(code: errorCode))))
+        } catch {
+            fatalError("C callback needs to updated.")
+        }
         return
     }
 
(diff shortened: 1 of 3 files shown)


waahm7:
C callbacks are not designed to be throwable. If something goes wrong with the stream/connection, you can send a reset stream or close the connection.

Source/AwsCommonRuntimeKit/http/HTTPStreamCallbackCore.swift
```diff
@@ -105,7 +110,11 @@ private func onComplete(stream: UnsafeMutablePointer<aws_http_stream>?,
     let httpStreamCbData = Unmanaged<HTTPStreamCallbackCore>.fromOpaque(userData).takeUnretainedValue()
     let onStreamCompleteFn = httpStreamCbData.requestOptions.onStreamComplete
     guard errorCode == AWS_OP_SUCCESS else {
-        onStreamCompleteFn(.failure(CommonRunTimeError.crtError(CRTError(code: errorCode))))
+        do {
+            try onStreamCompleteFn(.failure(CommonRunTimeError.crtError(CRTError(code: errorCode))))
+        } catch {
+            fatalError("C callback needs to updated.")
```
ganeshnj, line 116: @waahm7  is there a way we cal close the connection here?

Source/AwsCommonRuntimeKit/http/HTTPStreamCallbackCore.swift
```diff
@@ -118,7 +127,11 @@ private func onComplete(stream: UnsafeMutablePointer<aws_http_stream>?,
             """
         )
     }
-    onStreamCompleteFn(.success(UInt32(status)))
+    do {
+        try onStreamCompleteFn(.success(UInt32(status)))
+    } catch {
+        fatalError("C callback needs to updated.")
```
ganeshnj, line 133: @waahm7  is there a way we cal close the connection here?

Source/AwsCommonRuntimeKit/http/HTTPStreamCallbackCore.swift
waahm7: While we might be able to handle this error at this location, it is not the best place to do so. Ideally, the SDK should have a pointer to the stream in this function and should send the appropriate [RST_STREAM](https://datatracker.ietf.org/doc/html/rfc9113#section-6.4) or [GO_AWAY](https://datatracker.ietf.org/doc/html/rfc9113#name-goaway) frame for an HTTP/2 connection.

Source/AwsCommonRuntimeKit/http/HTTPStreamCallbackCore.swift
```diff
@@ -94,8 +94,13 @@ private func onResponseBody(stream: UnsafeMutablePointer<aws_http_stream>?,
...
     let incomingBodyFn = httpStreamCbData.requestOptions.onIncomingBody
     let callbackBytes = Data(bytesNoCopy: bufPtr, count: bufLen, deallocator: .none)
-    incomingBodyFn(callbackBytes)
-    return AWS_OP_SUCCESS
+
+    do {
+        try incomingBodyFn(callbackBytes)
+        return AWS_OP_SUCCESS
+    } catch {
+        return AWS_OP_ERR
```
waahm7, line 102: We should raise an error, something like `return aws_raise_error(Int32(AWS_ERROR_HTTP_CALLBACK_FAILURE.rawValue))`

waahm7: We should also make other callbacks throwable apart from onComplete.

Source/AwsCommonRuntimeKit/http/HTTPStreamCallbackCore.swift
waahm7: Discussed offline that we will not be making onComplete throwable.

waahm7 (approved):

epau (approved):

Issue #163, opened 2023-02-18
S2N prevents building on Linux


ugocottin:
### Describe the bug

Building on Linux distro throw an error caused by `aws-common-runtime/s2n` submodule.
This error does not occur on macOS.

### Expected Behavior

Building should be successful on Linux distro

### Current Behavior

```sh
error: 'src': target 'S2N' has invalid header layout: umbrella header found at '/src/aws-common-runtime/s2n/api/S2N.h', but directories exist next to it: /src/aws-common-runtime/s2n/api/unstable; consider removing them
```

### Reproduction Steps

On Linux distro, clone project with submodule:
```sh
git clone --recurse-submodules https://github.com/awslabs/aws-crt-swift.git
```

Build project:
```
swift build
```

### Additional Information/Context

S2N submodule version: v1.3.35

### AWS CRT SWIFT version used

0.6.1

### Compiler and Version used

Swift 5.7

### Operating System and version

Amazon Linux 2 (4.14), Ubuntu 22.04

waahm7:
Hi, 

Thank you for creating the issue. I have tried it on both Amazon Linux 2 and Ubuntu 22.04 but could not reproduce the issue. It works fine for both. Could you please try cloning it again and building it? Also, you will need to install OpenSSL, please do the following.

For Ubuntu:
```
apt-get install openssl
apt-get install libssl-dev
```
For Amazon Linux 2:
```
yum install openssl
yum install openssl-devel
```

waahm7:
Hi, 

Thank you for your answer. I have tried using both podman and docker to mount the source code, and I am able to do so without encountering any build issues. Could you please try cloning the repository again to see if that resolves the problem? Thank you.

ugocottin:
Hi,

I cloned again the repository (now on commit `ccec2f7`).

I've tried again on my machine (macOS 13.2.1 on Apple Silicon M1 Pro), and on another Mac (macOS 13.0.1, Intel), both with podman version 4.4.1 with vm Fedora CoreOS 37.X, unsuccessfully.

I've also tried with various versions of the container image `swift:5.X-amazonlinux2`, unsuccessfully.

waahm7:
Thank you for the response. Can you please explain a bit more about the steps that you are taking to encounter this issue, or what am I missing below?

1. Create an AL2 or Ubuntu EC2 instance.
[...]

PR #179, opened 2023-04-12
Fix S2N for Linux

waahm7:
*Issue #, if available:*
#163 

*Description of changes:*
`s2n.h` is treated as an umbrella header because it has the same name as the target on a case-insensitive file system. I tried updating the target name to `AwsCS2N`, but that didn't work because S2N has a [module map](https://github.com/aws/s2n-tls/blob/main/module.modulemap) which uses the module name `S2N`. After reviewing the Clang modules [documentation](https://clang.llvm.org/docs/Modules.html), I added `-fno-implicit-modules`, which ignores the module map file provided by S2N and uses the file generated by the Swift compiler.


JonathanHenson (approved):

Package.swift
```diff
@@ -37,18 +37,19 @@ let s2nExcludes = ["bin", "codebuild", "coverage", "docker-images",
...
     dependencies: ["LibCrypto"],
     path: "aws-common-runtime/s2n",
     exclude: s2nExcludes,
     publicHeadersPath: "api",
     cSettings: [
         .headerSearchPath("./"),
         .define("POSIX_C_SOURCE=200809L"),
-        .define("S2N_NO_PQ")
+        .define("S2N_NO_PQ"),
+        .unsafeFlags(["-fno-implicit-module-maps"])
```
jbelkins, line 49: I think using `unsafeFlags()` in a target definition prohibits the package from being published as a version (it may still be published as a Git SHA, branch, etc.): https://developer.apple.com/documentation/packagedescription/swiftsetting/unsafeflags(_:_:) (see the Discussion on this page.)

Issue #159, opened 2023-02-10
Fix Package.Swift for iOS


waahm7:
### Describe the bug

Currently, there are checks like https://github.com/awslabs/aws-crt-swift/blob/main/Package.swift#L117 which do not take care of what should happen for iOS. We should add explicit checks for iOS after researching.

epau:
As part of this work, we must refactor our handling of platform and architecture specific dependencies. Currently, we use conditionals to specify specific files to use for a particular platform or architecture. Because applications may be cross compiled, this can result in an incorrect configuration. 

The package description provides a way to customize dependencies by platform https://github.com/apple/swift-evolution/blob/main/proposals/0273-swiftpm-conditional-target-dependencies.md

We must investigate if this also supports architecture based conditions and if not we must investigate a different way to handle customizing dependencies based on the target platform's architecture.

waahm7:
As part of this work, we will also look at #163.

PR #184, opened 2023-05-04
Remove arch based conditionals

waahm7:
*Issue #, if available:*
https://github.com/awslabs/aws-crt-swift/issues/159
*Description of changes:*

The Package.swift file is compiled at compile-time, and SwiftPM is not yet mature enough to handle complex cross-compilations. Remove the #if directive for the architecture. This particular #if directive wouldn't have created an issue because:
- We always excluded the intel folder.
- For ARM, in the C code, we only have a runtime check for Linux and BSD.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -63,10 +63,10 @@ var awsCCalPlatformExcludes = [
     "ecdsa-fuzz-corpus/windows/p256_sig_corpus.txt",
     "ecdsa-fuzz-corpus/darwin/p256_sig_corpus.txt"] + excludesFromAll
 
-#if os(macOS) || os(iOS)
+#if os(macOS)
 awsCCalPlatformExcludes.append("source/windows")
 awsCCalPlatformExcludes.append("source/unix")
-#elseif(Windows)
+#elseif os(Windows)
 awsCCalPlatformExcludes.append("source/darwin")
 awsCCalPlatformExcludes.append("source/unix")
 #else
--- a/Package.swift
+++ b/Package.swift
@@ -31,7 +31,6 @@ var awsCCommonPlatformExcludes = ["source/android",
                                   "scripts/appverifier_ctest.py",
                                   "scripts/appverifier_xml.py"] + excludesFromAll
 
-
 // includes arch/generic
 awsCCommonPlatformExcludes.append("source/arch/intel")
 awsCCommonPlatformExcludes.append("source/arch/arm")
--- a/Package.swift
+++ b/Package.swift
@@ -22,6 +22,7 @@ let cSettings: [CSetting] = [
 //////////////////////////////////////////////////////////////////////
 /// Configure C targets.
 /// Note: We can not use unsafe flags because SwiftPM makes the target ineligible for use by other packages.
+///       We are also not using any architecture based conditionals due to lack of proper cross compilation support.
 /// Configure aws-c-common
 //////////////////////////////////////////////////////////////////////
 var awsCCommonPlatformExcludes = ["source/android",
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -78,7 +65,7 @@ var awsCCalPlatformExcludes = [
 #if os(macOS)
 awsCCalPlatformExcludes.append("source/windows")
 awsCCalPlatformExcludes.append("source/unix")
-#elseif(Windows)
+#elseif os(Windows)
 awsCCalPlatformExcludes.append("source/darwin")
 awsCCalPlatformExcludes.append("source/unix")
 #else
```
graebm, line 71: trivial: Sorry, every time I see this block of code in a PR it just makes me itchy

Looks like the only valid values are (from [docs](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements/#Conditional-Compilation-Block)):
> macOS, iOS, watchOS, tvOS, Linux, Windows

maybe write this as:
```
#if !os(Windows)
awsCCalPlatformExcludes.append("source/windows")
#endif

#if !os(Linux)
awsCCalPlatformExcludes.append("source/unix")
#endif

#if !(os(macOS) || os(iOS) || os(watchOS) || os(tvOS))
awsCCalPlatformExcludes.append("source/darwin")
#endif
```

or
```
#if os(Windows)
awsCCalPlatformExcludes.append("source/darwin")
awsCCalPlatformExcludes.append("source/unix")
#elif os(Linux)
awsCCalPlatformExcludes.append("source/darwin")
awsCCalPlatformExcludes.append("source/windows")
#else // macOS, iOS, watchOS, tvOS
awsCCalPlatformExcludes.append("source/windows")
awsCCalPlatformExcludes.append("source/unix")
#endif
```

or

change from an "include everything then carefully exclude things" model, to an "explicitly include things, use a helper function to gather up the majority of files from where we expect them to be" model

or

tell me to go away, we'll revamp this when we truly revisit supporting iOS

Package.swift
```diff
@@ -22,33 +22,20 @@ let cSettings: [CSetting] = [
 //////////////////////////////////////////////////////////////////////
 /// Configure C targets.
 /// Note: We can not use unsafe flags because SwiftPM makes the target ineligible for use by other packages.
+///       We are also not using any architecture based conditionals due to lack of proper cross compilation support.
 /// Configure aws-c-common
 //////////////////////////////////////////////////////////////////////
-var awsCCommonPlatformExcludes = ["source/windows", "source/android",
+var awsCCommonPlatformExcludes = ["source/android",
                                   "AWSCRTAndroidTestRunner", "verification",
                                   "include/aws/common/", "sanitizer-blacklist.txt",
```
DmitriyMusatkin, line 30: sanitizer blacklists no longer exist

Package.swift
DmitriyMusatkin: i might be missing some context and previous discussions on this, but i agree this code block looks off to me as well. is there a major reason why we cant specify base of things to include and then add platform specific things to it instead of going exclusion route

waahm7: SwiftPM doesn't support something like "source/*.c" and recursively including everything. We can have a hacky solution where we can create a custom function that returns a list of files and then include platform-specific content. This is a hack, and we will have to make sure it does not have any unintended consequences.

In the long term, we should eliminate OS-specific checks as well because they don't work for cross-compiling use cases like MacOS -> Linux. However, the current SwiftPM approach is to have separate dependencies for different OSs and conditionally link with them, which is not feasible in our case. Another solution I can think of is linking to C repositories as binary targets and generating platform-specific binaries, but I am not sure how much effort this involves.
https://forums.swift.org/t/platform-specific-resources/43612/3

waahm7: For now, I have updated it to the second approach.

Package.swift
waahm7: Thanks, fixed.

epau (approved):

Issue #198, opened 2023-09-21
Add Swift support for flexible checksums


smivij:
### Describe the feature

As per Flexible Checksums SEP, we need to support below checksum algorithms and need support from CRT for it.

CRC32C	
CRC32	
SHA256
SHA1	

I believe all of these are already implemented in CRT, we need Swift bindings for these. I believe SHA256 already has Swift bindings, need work for the others.

### Use Case

please refer to flexible checksum SEP

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

waahm7:
Released in `0.18.0`.

PR #209, opened 2023-11-06
Bind Sha1, crc32 and crc32c 

waahm7:
Issue: https://github.com/awslabs/aws-crt-swift/issues/198

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -248,7 +248,7 @@ packageTargets.append(contentsOf: [
         cSettings: cSettings
     ),
     .target(
-        name: "AwsChecksums",
+        name: "AwsCChecksums",
         dependencies: ["AwsCCommon"],
         path: "aws-common-runtime/aws-checksums",
         exclude: awsCChecksumsExcludes,
@@ -256,7 +256,7 @@ packageTargets.append(contentsOf: [
     ),
     .target(
         name: "AwsCEventStream",
-        dependencies: ["AwsChecksums", "AwsCCommon", "AwsCIo", "AwsCCal"],
+        dependencies: ["AwsCChecksums", "AwsCCommon", "AwsCIo", "AwsCCal"],
         path: "aws-common-runtime/aws-c-event-stream",
         exclude: awsCEventStreamExcludes,
         cSettings: cSettings
@@ -269,7 +269,7 @@ packageTargets.append(contentsOf: [
                         "AwsCCompression",
                         "AwsCIo",
                         "AwsCCommon",
-                        "AwsChecksums",
+                        "AwsCChecksums",
                         "AwsCEventStream",
                         .product(name: "Collections", package: "swift-collections")],
         path: "Source/AwsCommonRuntimeKit"
--- a/Source/AwsCommonRuntimeKit/crt/Checksums.swift
+++ b/Source/AwsCommonRuntimeKit/crt/Checksums.swift
@@ -2,7 +2,6 @@
 //  SPDX-License-Identifier: Apache-2.0.
 
 import struct Foundation.Data
-import struct Foundation.TimeInterval
 import AwsCChecksums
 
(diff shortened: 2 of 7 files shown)


Package.swift
```diff
@@ -248,15 +248,15 @@ packageTargets.append(contentsOf: [
         cSettings: cSettings
     ),
     .target(
-        name: "AwsChecksums",
+        name: "AwsCChecksums",
```
waahm7, line 251: Debatable naming: Should we stay consistent with C module naming or with how C modules are named in Swift?

graebm: This looks good. Name it like the others

Source/AwsCommonRuntimeKit/crt/Checksums.swift
```diff
@@ -0,0 +1,31 @@
...
+import AwsCChecksums
+
+extension Data {
+    
+    /// Computes the CRC32 over data.
+    /// - Parameter previousCrc32:  Pass 0 in the previousCrc32 parameter as an initial value unless continuing to update a running crc in a subsequent call.
+    public func crc32(previousCrc32: UInt32 = 0) -> UInt32 {
+        self.withUnsafeBytes { bufferPointer in
+            guard let ptr = bufferPointer.baseAddress?.assumingMemoryBound(to: UInt8.self) else {
+                return 0
```
graebm, line 14: debatable: I'd prefer we pass `ptr=NULL, len=0` down to the C function. Just use the one canonical implementation, rather than let the language binding handle 1 specific case.

Source/AwsCommonRuntimeKit/crt/Checksums.swift
```diff
@@ -0,0 +1,31 @@
...
+            return aws_checksums_crc32(ptr, Int32(count), previousCrc32)
+        }
+    }
+    
+    /// Computes the crc32c over data.
+    /// - Parameter previousCrc32c:  Pass 0 in the previousCrc32c parameter as an initial value unless continuing to update a running crc in a subsequent call.
+    public func crc32c(previousCrc32c: UInt32 = 0) -> UInt32 {
+        self.withUnsafeBytes { bufferPointer in
+            guard let ptr = bufferPointer.baseAddress?.assumingMemoryBound(to: UInt8.self) else {
+                return 0
```
graebm, line 25: same

graebm (approved):
fix & ship

Source/AwsCommonRuntimeKit/crt/Checksums.swift
```diff
@@ -0,0 +1,29 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+
+import struct Foundation.Data
+import AwsCChecksums
+
+extension Data {
```
jbelkins, line 7: Why are these two checksums in `Checksums.swift` while the other two are in `Utilities.swift`?

jbelkins (commented):
Just one question

Source/AwsCommonRuntimeKit/crt/Checksums.swift
waahm7: SHA1 and SHA256 are cryptographic hash functions, while CRC32 and CRC32C are checksum algorithms with hardware and software based implementation in C. SHA1 and SHA256 can be used as checksums, so I was a bit confused about their placement and initially put them in Checksums.swift. However, considering that they are conceptually different, belong to different modules in C, and have different interfaces, I kept them in Utilities.swift.

waahm7: Thanks, I have moved the three hash functions to a Hash.swift file. Now our Utilities.swift contains only internal utility functions, which is great.

PR #216, opened 2023-11-22
Fix EventStream to properly handle utf-8

waahm7:
*Issue #, if available:*

*Description of changes:*
- There was a bug where the length of the String was not properly passed down to the C layer, since .count returns the number of characters in the String, but one character can take up multiple bytes in UTF-8.


Source/AwsCommonRuntimeKit/event-stream/EventStreamMessage.swift
```diff
@@ -111,16 +112,16 @@ extension EventStreamMessage {
                         aws_event_stream_add_string_header(
                             rawHeaders,
                             headerName,
-                            UInt8(header.name.count),
+                            headerNameLength,
                             $0,
-                            UInt16(value.count),
+                            UInt16(value.utf8.count),
```
graebm, line 117: above this diff, `if value.count > EventStreamHeader.maxValueLength` needs to also use utf8.count

Source/AwsCommonRuntimeKit/event-stream/EventStreamMessage.swift
```diff
@@ -54,36 +54,37 @@ extension EventStreamMessage {
...
         let addCHeader: () throws -> Int32 = {
             return try header.name.withCString { headerName in
+                let headerNameLength = UInt8(header.name.utf8.count)
                 switch header.value {
                 case .bool(let value):
                     return aws_event_stream_add_bool_header(
                         rawHeaders,
                         headerName,
-                        UInt8(header.name.count),
+                        headerNameLength,
```
graebm, line 63: This is spooky. Please search the codebase for all other uses of `withCString` or `.count` to see if we're making this mistake anywhere else

Source/AwsCommonRuntimeKit/event-stream/EventStreamMessage.swift
```diff
@@ -54,36 +54,37 @@ extension EventStreamMessage {
         }
         let addCHeader: () throws -> Int32 = {
             return try header.name.withCString { headerName in
+                let headerNameLength = UInt8(header.name.utf8.count)
```
graebm, line 57: I wonder if `.utf8.count` is really just O(N) under the hood...

If it is, I'd be tempted to just call `strlen(cstring)` because it seems harder to mess up, compared to checking that we always do `swiftstring.utf8.count` and never accidentally do `swiftstring.count`

really wish `withCString()` passed both pointer and length to the closure

Source/AwsCommonRuntimeKit/event-stream/EventStreamMessage.swift
```diff
@@ -54,36 +54,37 @@ extension EventStreamMessage {
         }
```
graebm, line 54: above this diff, using `header.name.count` again

Source/AwsCommonRuntimeKit/event-stream/EventStreamMessage.swift
waahm7: Thanks, I have removed `withCString` since it is not needed and Swift will automatically do it for you under the hood. I have also verified the usage of `.count`.

Source/AwsCommonRuntimeKit/event-stream/EventStreamMessage.swift
waahm7: Updated and removed withCString. Yeah, I agree it's really easy to mess up, given that we need to pass `.utf8.count` to C APIs instead of `.count`.

Source/AwsCommonRuntimeKit/event-stream/EventStreamMessage.swift
waahm7: Thanks, fixed.

Source/AwsCommonRuntimeKit/event-stream/EventStreamMessage.swift
```diff
@@ -47,90 +47,89 @@ public struct EventStreamMessage {
...
-                        value)
-                case .int16(let value):
-                    return aws_event_stream_add_int16_header(
-                        rawHeaders,
-                        headerName,
-                        UInt8(header.name.count),
-                        value)
-                case .int32(let value):
-                    return aws_event_stream_add_int32_header(
+            let headerNameLength = UInt8(header.name.utf8.count)
```
graebm, line 56: trivial: move this up, so we can do `if headerNameLength > EventStreamHeader.maxNameLength`
instead of needing to type the easy-to-mess-up `.utf8.count` multiple times

Source/AwsCommonRuntimeKit/event-stream/EventStreamMessage.swift
```diff
@@ -47,90 +47,89 @@ public struct EventStreamMessage {
...
                         rawHeaders,
-                        headerName,
-                        UInt8(header.name.count),
-                        value)
-                case .int64(let value):
-                    return aws_event_stream_add_int64_header(
+                        header.name,
+                        headerNameLength,
+                        bytes,
+                        UInt16($0.count),
```
graebm, line 102: trivial: similar, we get length in 2 mildly different ways here: `value.count` up at the length check, then `$0.count` down here. Can we use a temp variable, or do `value.count` in both places, or move the length-check down so we can do`$0.count` in both places?

graebm (approved):
fix & ship

graebm (approved):

Issue #218, opened 2024-01-03
Add support for writeChunk


dayaffe:
### Describe the feature

Add support for [writeChunk](https://github.com/awslabs/aws-crt-java/blob/236940324212d93015f50668247314415922377a/src/main/java/software/amazon/awssdk/crt/http/HttpStream.java#L44) similar to the Java's CRT. This method should allow us to use it for either chunked encoding or aws-chunked encoding which adds the ability to specify a checksum per block.

### Use Case

Used for chunked encoding and for flexible checksums

### Proposed Solution

Bind out the implementation that already exists in C to the Swift CRT.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #220, opened 2024-01-08
WriteChunk API for HTTP1

waahm7:
*Issue #, if available:*
#218 
*Description of changes:*
- Adds support for writeChunk API in HTTP1.
- Adds unified writeChunk in HTTStream base class. 
- Update tests to use postman-echo instead of httpbin.

### DIFF
--- a/Source/AwsCommonRuntimeKit/http/HTTPStream.swift
+++ b/Source/AwsCommonRuntimeKit/http/HTTPStream.swift
@@ -42,7 +42,8 @@ public class HTTPStream {
         }
     }
     
-    public func writeChunk(chunk: Data) async throws {
+    /// This method must be overriden in each subclass.
+    public func writeChunk(chunk: Data, endOfStream: Bool) async throws {
         fatalError("writeChunk is not implemented for HTTPStream base")
     }
     
--- a/Source/AwsCommonRuntimeKit/http/HTTPStream.swift
+++ b/Source/AwsCommonRuntimeKit/http/HTTPStream.swift
@@ -42,7 +42,9 @@ public class HTTPStream {
         }
     }
     
-    /// This method must be overriden in each subclass.
+    /// This method must be overridden in each subclass because this function is specific to each subclass.
+    /// For HTTP/1.1 see ``HTTP1Stream/writeChunk(chunk:endOfStream:)``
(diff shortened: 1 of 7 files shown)


Source/AwsCommonRuntimeKit/http/HTTPStream.swift
```diff
@@ -3,7 +3,7 @@
 import AwsCHttp
 import Foundation
 
-/// An base class that represents a single Http Request/Response for both HTTP/1.1 and HTTP/2.
+/// An base abstract class that represents a single Http Request/Response for both HTTP/1.1 and HTTP/2.
 /// Can be used to update the Window size, and get status code.
 public class HTTPStream {
```
waahm7, line 8: I thought about turning this class into a protocol to avoid the abstract class workaround with `fatalAssert`, but that would require making `rawValue` and `callbackData` variables public.

Source/AwsCommonRuntimeKit/http/HTTPStream.swift
```diff
@@ -41,7 +41,12 @@ public class HTTPStream {
             throw CommonRunTimeError.crtError(.makeFromLastError())
         }
     }
-
+    
+    /// This method must be overriden in each subclass.
+    public func writeChunk(chunk: Data, endOfStream: Bool) async throws {
```
graebm, line 46: mildly annoyed by calling this "chunk"
"chunk" is an HTTP/1-specific term

Source/AwsCommonRuntimeKit/http/HTTPStream.swift
```diff
@@ -41,7 +41,12 @@ public class HTTPStream {
             throw CommonRunTimeError.crtError(.makeFromLastError())
         }
     }
-
+    
+    /// This method must be overriden in each subclass.
```
graebm, line 45: is this how interface functions are supposed to be documented? doesn't seem very useful to the user

Should this document how you have to set up the connection with "use manual data writes" or "transfer-encoding: chunked" depending on whether it's HTTP/1 or HTTP/2?

graebm (approved):

Source/AwsCommonRuntimeKit/http/HTTPStream.swift
waahm7: I have updated the documentation to link to the H1 and H2 documentation.

PR #221, opened 2024-01-09
Fix MD5 to operate over data

waahm7:
*Description of changes:*
- Change MD5 function to operate over data instead of String to avoid any encoding/decoding issues and just operate on arbitrary bytes.

### DIFF
--- a/Source/AwsCommonRuntimeKit/crt/Hash.swift
+++ b/Source/AwsCommonRuntimeKit/crt/Hash.swift
@@ -7,7 +7,7 @@ import AwsCCal
 
 extension Data {
     
-    /// Computes the base64 encoded md5 hash over data. Use this if you don't need to stream the data you're hashing and you can load
+    /// Computes the md5 hash over data. Use this if you don't need to stream the data you're hashing and you can load
     /// the entire input to hash into memory.
     /// - Parameter truncate: If you specify truncate something other than 0, the output will be truncated to that number of bytes.
     public func computeMD5(truncate: Int = 0) throws -> Data {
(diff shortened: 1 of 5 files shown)


Source/AwsCommonRuntimeKit/crt/Hash.swift
```diff
@@ -5,27 +5,26 @@ import struct Foundation.Data
...
+
+            let bufferSize = 16
+            var bufferData = Data(count: bufferSize)
+            try bufferData.withUnsafeMutableBytes { bufferPointer in
+                var buffer = aws_byte_buf_from_empty_array(bufferPointer.baseAddress, bufferSize)
+                guard aws_md5_compute(allocator.rawValue, &byteCursor, &buffer, truncate) == AWS_OP_SUCCESS else {
+                    throw CommonRunTimeError.crtError(.makeFromLastError())
+                }
             }
+            return bufferData.base64EncodedString()
```
dayaffe, line 25: SDK can call base64EncodedString() its probably better to just return the Data here directly and rename function `MD5(...)`

dayaffe (approved):

Source/AwsCommonRuntimeKit/crt/Hash.swift
```diff
@@ -48,7 +47,7 @@ extension Data {
     /// Computes the sha1 hash over data.
     /// - Parameter truncate: If you specify truncate something other than 0, the output will be truncated to that number of bytes. For
     /// example, if you want a SHA1 digest as the first 10 bytes, set truncate to 10.
-    public func sha1(truncate: Int = 0) throws -> Data {
+    public func compteSHA1(truncate: Int = 0) throws -> Data {
```
dayaffe, line 50: spelling error -- should be `computeSHA1`

waahm7: Thanks, fixed.

Issue #212, opened 2023-11-15
Crash occurring when trying to force unwrap possible nil response header


Jeffrey-Chau-Leo:
### Describe the bug

A crash is occurring when trying to verify the users email with a verification code with amplify.

The `toString()` function below in `Utilities.swift` crashes due to possible force unwrapping when nil

```
extension aws_byte_cursor {
    func toString() -> String {
        if self.len == 0 {
            return ""
        }

        let data = Data(bytesNoCopy: self.ptr, count: self.len, deallocator: .none)
        return String(data: data, encoding: .utf8)!
    }

    func toOptionalString() -> String? {
        if self.len == 0 { return nil }
        let data = Data(bytesNoCopy: self.ptr, count: self.len, deallocator: .none)
        return String(data: data, encoding: .utf8)!
    }
}
```

### Expected Behavior

No crash

### Current Behavior

```
Thread 8 Crashed:
0   App                             0x101570abc         [inlined] value
1   App                             0x101570abc         [inlined] aws_byte_cursor.toString (Utilities.swift:160)
2   App                             0x101570abc         [inlined] HTTPHeader.init (HTTPHeader.swift:18)
3   App                             0x101570abc         [inlined] HTTPHeader.__allocating_init (HTTPHeader.swift:19)
4   App                             0x101570abc         [inlined] onResponseHeaders (HTTPStreamCallbackCore.swift:56)
5   App                             0x101570abc         [inlined] Sequence.forEach
6   App                             0x101570abc         onResponseHeaders (HTTPStreamCallbackCore.swift:56)
7   App                             0x101570960         [inlined] aws_byte_cursor.toString (Utilities.swift:160)
8   App                             0x101570960         [inlined] HTTPHeader.init (HTTPHeader.swift:18)
9   App                             0x101570960         [inlined] HTTPHeader.__allocating_init (HTTPHeader.swift:19)
10  App                             0x101570960         [inlined] onResponseHeaders (HTTPStreamCallbackCore.swift:56)
11  App                             0x101570960         [inlined] Sequence.forEach
12  App                             0x101570960         onResponseHeaders (HTTPStreamCallbackCore.swift:56)
13  App                             0x1015bb048         s_decoder_on_header (h1_connection.c:1156)
14  App                             0x1015bc8a4         s_linestate_header (h1_decoder.c:534)
15  App                             0x1015bc1f0         s_state_getline (h1_decoder.c:126)
[...]

PR #223, opened 2024-01-17
Fix String Decode Crash

waahm7:
*Issue #, if available:*
#212 

*Description of changes:*
- If the raw bytes encoding of headers is not UTF-8, the force unwrapping will crash. Change it to try to decode it as utf-8 replace invalid characters with unknown characters (�).

### DIFF
--- a/Source/AwsCommonRuntimeKit/crt/Utilities.swift
+++ b/Source/AwsCommonRuntimeKit/crt/Utilities.swift
@@ -136,7 +136,7 @@ extension aws_byte_cursor {
     func toOptionalString() -> String? {
         if self.len == 0 { return nil }
         let data = Data(bytesNoCopy: self.ptr, count: self.len, deallocator: .none)
-        return String(decoding: data, as: UTF8.self)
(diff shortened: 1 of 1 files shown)


Source/AwsCommonRuntimeKit/crt/Utilities.swift
```diff
@@ -130,13 +130,13 @@ extension aws_byte_cursor {
...
         let data = Data(bytesNoCopy: self.ptr, count: self.len, deallocator: .none)
-        return String(data: data, encoding: .utf8)!
+        return String(decoding: data, as: UTF8.self)
     }
 
     func toOptionalString() -> String? {
         if self.len == 0 { return nil }
         let data = Data(bytesNoCopy: self.ptr, count: self.len, deallocator: .none)
-        return String(data: data, encoding: .utf8)!
+        return String(data: data, encoding: .utf8)
```
waahm7, line 139: This is used for the credential's access key, secret access key, and session token. We don't want partial characters in this case.

DmitriyMusatkin (approved):

Issue #210, opened 2023-11-07
Enable HW Based Implementation for Checksum Algorithms


waahm7:
### Describe the feature

Re-enable hardware-based acceleration for checksum algorithms on at least macOS, which we disabled at https://github.com/awslabs/aws-crt-swift/blob/main/Package.swift#L144.

### Use Case

HW algorithms will have a better performance.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

waahm7:
We need advanced features in SwiftPM before enabling hardware-accelerated checksums. Specifically, we need the ability to specify CPU flags such as `-march`, and an equivalent of `check_c_source_compiles` or the ability to use CMake to build the C dependencies, as discussed in [this issue](https://github.com/apple/swift-package-manager/issues/4555).

PR #225, opened 2024-01-22
Improves documentation in Package.swift regarding lack of CPU flags.

waahm7:
*Issue #, if available:*
#210 

*Description of changes:*
- Improves documentation around why we have disabled the hardware-accelerated checksum in Swift.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -144,11 +144,11 @@ var awsCChecksumsExcludes = [
 // swift never uses Microsoft Visual C++ compiler
 awsCChecksumsExcludes.append("source/intel/visualc")
 
-// Hardware accellarated checksums are disabled because Swift PM doesn't like the necessary compiler flag.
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -33,7 +33,7 @@ var awsCCommonPlatformExcludes = ["source/android",
                                   "source/linux/system_info.c",
                                   "bin/"] + excludesFromAll
 
-// includes arch/generic
+// includes arch/generic because the Swift PM doesn't like the necessary compiler flags.
```
DmitriyMusatkin, line 36: nit: Maybe refer to it as SwiftPM (which is how their docs call it). standalone PM has a lot of other connatations

waahm7: Thanks, updated.

DmitriyMusatkin (approved):

PR #233, opened 2024-02-16
Extend TLS Option

xiazhvera:
*Issue #, if available:*

*Description of changes:*


Source/AwsCommonRuntimeKit/io/TLSContextOptions.swift
```diff
@@ -15,6 +15,18 @@ public class TLSContextOptions: CStruct {
         try TLSContextOptions(mtlsPkcs12FromPath: path, password: password)
     }
 
+    public static func makeMtlsFromRawData(
+        certificateData: String,
+        privateKeyData: String) throws -> TLSContextOptions {
+        try TLSContextOptions(certificateData: privateKeyData, privateKeyData: privateKeyData)
```
sbSteveK, line 21: It looks like certificateData is being assigned privateKeyData

Source/AwsCommonRuntimeKit/io/TLSContextOptions.swift
```diff
@@ -33,6 +45,32 @@ public class TLSContextOptions: CStruct {
...
+        } != AWS_OP_SUCCESS) {
+            throw CommonRunTimeError.crtError(CRTError.makeFromLastError())
+        }
+    }
+
+    init(certificatePath cert_path: String,
+         privateKeyPath private_path: String) throws {
+        self.rawValue = allocator.allocate(capacity: 1)
+        if aws_tls_ctx_options_init_client_mtls_from_path(rawValue,
+                                                           allocator.rawValue,
```
sbSteveK, line 67: Trivial: misaligned arguments

sbSteveK (approved):

Source/AwsCommonRuntimeKit/io/TLSContextOptions.swift
```diff
@@ -33,6 +45,32 @@ public class TLSContextOptions: CStruct {
         }
     }
 
+    init(certificateData cert_data: String,
+         privateKeyData private_key_data: String) throws {
+        self.rawValue = allocator.allocate(capacity: 1)
+        if( cert_data.withByteCursorPointer { certificateByteCursor in
```
waahm7, line 51: - guard let is better for this case because it ensures that we don't fall through the error branch.
- You can avoid nesting withByteCursorPointer by using the helper function like https://github.com/awslabs/aws-crt-swift/blob/main/Source/AwsCommonRuntimeKit/http/HTTPProxyOptions.swift#L37

xiazhvera: Updated to use an extra help function to pass the bytecursor.

Test/AwsCommonRuntimeKitTests/io/TLSContextTests.swift
```diff
@@ -10,4 +10,18 @@ class TLSContextTests: XCBaseTestCase {
     let context = try TLSContext(options: options, mode: .client)
     _ = TLSConnectionOptions(context: context)
   }
+
+  // TODO: The test is disabled as the github CI failed on it.
```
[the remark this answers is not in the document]
xiazhvera: The error was incorrectly raised in aws-c-io. It should be fixed in this PR: https://github.com/awslabs/aws-c-io/pull/622. 

The CI failed with the proper error after we updated aws-c-io: 
https://github.com/awslabs/aws-crt-swift/actions/runs/8510585488/job/23308473095?pr=233

Source/AwsCommonRuntimeKit/io/TLSContextOptions.swift
```diff
@@ -33,6 +45,32 @@ public class TLSContextOptions: CStruct {
...
+            throw CommonRunTimeError.crtError(CRTError.makeFromLastError())
+        }
+        self.rawValue = rawValue
+    }
+
+    init(certificatePath cert_path: String,
+         privateKeyPath private_path: String) throws {
+        self.rawValue = allocator.allocate(capacity: 1)
+        if aws_tls_ctx_options_init_client_mtls_from_path(rawValue,
+                                                           allocator.rawValue,
```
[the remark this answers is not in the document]
xiazhvera: I added the `vertical_parameter_alignment_on_call` in .swiftlint.yml so swiftlint would check for the vertical alignment. However, run swiftlint would not do auto-correct for it. We would need manually fix it if the misalignment is reported.

Test/AwsCommonRuntimeKitTests/io/TLSContextTests.swift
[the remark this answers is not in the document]
xiazhvera: I will revert the aws-c-io, we probably should update submodules in a different PR?

Source/AwsCommonRuntimeKit/io/TLSContextOptions.swift
```diff
@@ -9,10 +9,46 @@ public class TLSContextOptions: CStruct {
...
-        path: String,
+    /// Initializes TlsContextOptions for mutual TLS (mTLS), with client certificate and private key in the PKCS#12 format.
+    ///
+    /// NOTE: This only works on Apple devices.
+    ///
+    /// - Parameters:
+    ///     - pkcs12Path: Path to PKCS #12 file. The file is loaded from disk and stored internally. It must remain in
+    ///     memory for the lifetime of the returned object.
+    ///     - password: Password to PKCS #12 file. It must remain in memory for the lifetime of the returned object.
+    public static func makeMtls(
```
waahm7, line 20: You should make this change to the main the branch since this is a breaking change and we would like the SDK to update to this change ASAP.

Source/AwsCommonRuntimeKit/io/TLSContextOptions.swift
```diff
@@ -9,10 +9,46 @@ public class TLSContextOptions: CStruct {
         TLSContextOptions()
     }
 
-    public static func makeMtlsPkcs12FromPath(
-        path: String,
+    /// Initializes TlsContextOptions for mutual TLS (mTLS), with client certificate and private key in the PKCS#12 format.
```
waahm7, line 12:
```suggestion
    /// Initializes TLSContextOptions for mutual TLS (mTLS), with client certificate and private key in the PKCS#12 format.
```

Source/AwsCommonRuntimeKit/io/TLSContextOptions.swift
```diff
@@ -9,10 +9,46 @@ public class TLSContextOptions: CStruct {
...
+    ///
+    /// NOTE: This only works on Apple devices.
+    ///
+    /// - Parameters:
+    ///     - pkcs12Path: Path to PKCS #12 file. The file is loaded from disk and stored internally. It must remain in
+    ///     memory for the lifetime of the returned object.
+    ///     - password: Password to PKCS #12 file. It must remain in memory for the lifetime of the returned object.
```
waahm7, line 19: trivial: you also need an entry for throws for proper Swift style documentation. See: https://github.com/awslabs/aws-crt-swift/blob/main/Source/AwsCommonRuntimeKit/auth/credentials/CredentialsProvider.swift#L25

Source/AwsCommonRuntimeKit/io/TLSContextOptions.swift
waahm7, line 20: Trivial/Debatable/Suggestion: 
```suggestion
    public static func makeMTLS(
```

Source/AwsCommonRuntimeKit/io/TLSContextOptions.swift
```diff
@@ -9,10 +9,46 @@ public class TLSContextOptions: CStruct {
         TLSContextOptions()
     }
 
-    public static func makeMtlsPkcs12FromPath(
-        path: String,
+    /// Initializes TlsContextOptions for mutual TLS (mTLS), with client certificate and private key in the PKCS#12 format.
+    ///
+    /// NOTE: This only works on Apple devices.
```
waahm7, line 14: `@available` directive? 
Pros:
- Compile time safety instead of runtime error

Cons:
- Consumers of this API will have to do IF/ELSE at compile time, although they should probably do that anyhow if it is a cross-platform application.

Source/AwsCommonRuntimeKit/io/TLSContextOptions.swift
```diff
@@ -9,10 +9,46 @@ public class TLSContextOptions: CStruct {
...
+        pkcs12Path: String,
         password: String) throws -> TLSContextOptions {
-        try TLSContextOptions(mtlsPkcs12FromPath: path, password: password)
+        try TLSContextOptions(mtlsPkcs12FromPath: pkcs12Path, password: password)
+    }
+
+    /// Initializes TlsContextOptions for mutual TLS (mTLS), with client certificate and private key. These are in memory
+    /// buffers. These buffers must be in the PEM format.
+    ///
+    /// NOTE: This is unsupported on iOS.
```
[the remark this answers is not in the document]
xiazhvera: Thinking about it, it's better to set restrictions early on instead of adding them later. I added the available here for the path and data API. 
```
    @available(iOS, unavailable)
    @available(tvOS, unavailable)
    @available(watchOS, unavailable)
    public static func makeMTLS(
        certificateData: String,
        privateKeyData: String) throws -> TLSContextOptions {
        try TLSContextOptions(certificateData: certificateData, privateKeyData: privateKeyData)
    }
```

For the pkcs12 one, it seems that `@available` only works for Apple platforms, we dont have a way to disable the API for linux/windows. I will leave it there for now.

xiazhvera:
The changes are brought in new PR: https://github.com/awslabs/aws-crt-swift/pull/246

PR #238, opened 2024-02-29
Setup CI for Apple Platforms

xiazhvera:
*Issue #, if available:*

*Description of changes:*


builder.json
```diff
@@ -38,8 +40,13 @@
...
+      ]
+    },
+    "linux":{
+      "!test_steps":[
+        "swift test"
       ]
     }
   }
-}
+}
```
graebm, line 52: utterly trivial: https://stackoverflow.com/questions/44704968/visual-studio-code-insert-newline-at-the-end-of-files

integ_test.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/bash
 curl -L -o /tmp/http_client_test.py https://raw.githubusercontent.com/awslabs/aws-c-http/main/integration-testing/http_client_test.py
-python3 /tmp/http_client_test.py .build/x86_64-apple-macosx/release/Elasticurl
+ARCH=$(uname -a | sed 's/.* \([^ ]*\)$/\1/')
+python3 /tmp/http_client_test.py .build/$ARCH-apple-macosx/release/Elasticurl
```
graebm, line 4: utterly trivial: same

Test/AwsCommonRuntimeKitTests/XCBaseTestCase.swift
```diff
@@ -10,7 +10,7 @@ class XCBaseTestCase: XCTestCase {
 
     override func setUp() {
         super.setUp()
-        Logger.initialize(pipe: stdout, level: .trace)
+        Logger.initialize(pipe: stdout, level: .error)
```
graebm, line 13: Did you intend to commit this? I genuinely have no idea what's best, I don't know what it looks like to run tests in this repo

.github/workflows/ci.yml
```diff
@@ -51,22 +61,58 @@ jobs:
 
   # TODO: Run tests for devices
```
graebm, line 62: can this TODO be removed now?

graebm (approved):
great work!

Test/AwsCommonRuntimeKitTests/XCBaseTestCase.swift
xiazhvera: I didnt find a good way to enable logs only on failed tests for XCode. Using "trace" mode would spam the outputs here. Therefore, I set it to error. Let me add comments about this, so we could possibly update it in the future.

Test/AwsCommonRuntimeKitTests/XCBaseTestCase.swift
```diff
@@ -35,20 +38,44 @@ extension XCTestCase {
...
+extension XCTestCase {
+    /// Return the environment variable value, or Skip the test if env var is not set.
+    func GetEnvironmentVarOrSkip(environmentVarName name: String) throws -> String? {
+        let result = ProcessInfo.processInfo.environment[name]
+        guard result != nil else {
+            try skipTest(message: "Skipping test because environment is not configured properly.")
+            return nil
+        }
+        return result
+    }
```
waahm7, line 80: * Why create a separate extension block? We can make this part of the same extension block as the `skipIf*` functions. 
* function name should start with lowercase.
* We should avoid nullable types as much as possible.
* I think it would be better to just get rid of `skipTest` function and use `throw XCTSkip(message)` directly.

```suggestion
    func getEnvironmentVarOrSkipTest(environmentVarName name: String) throws -> String {
        guard let result = ProcessInfo.processInfo.environment[name] else {
            throw XCTSkip(message: "Skipping test because environment is not configured properly.")
        }
        return result
    }
```

Test/AwsCommonRuntimeKitTests/XCBaseTestCase.swift
```diff
@@ -35,20 +38,44 @@ extension XCTestCase {
     }
 
     func skipIfiOS() throws {
```
waahm7, line 40: Debatable: These functions are unused, apart from `skipIfLinux`. We can just get rid of them until needed.

Test/AwsCommonRuntimeKitTests/XCBaseTestCase.swift
```diff
@@ -35,20 +38,44 @@ extension XCTestCase {
...
+    func skipIftvOS() throws {
+        #if os(tvOS)
+            throw XCTSkip("Skipping test on tvOS")
+        #endif
+    }
+}
+
+extension XCTestCase {
+    /// Return the environment variable value, or Skip the test if env var is not set.
+    func GetEnvironmentVarOrSkip(environmentVarName name: String) throws -> String? {
```
waahm7, line 73: Debatable: similar name to other functions `skipIfEnvironmentNotSetup`

.github/workflows/ci.yml
```diff
@@ -39,34 +39,79 @@ jobs:
...
+    runs-on: ${{ matrix.runner }}
     env:
       DEVELOPER_DIR: /Applications/Xcode.app
+      XCODE_DESTINATION: 'OS X'
+      NSUnbufferedIO: YES
+    strategy:
+      fail-fast: false
+      matrix:
+        # This matrix runs tests on iOS sim & Mac, on oldest & newest supported Xcodes
+        runner:
```
waahm7, line 51: trivial: We should have `macos-13` & `macos-14` as well if available, since `xlarge` runners are M1-based.

.github/workflows/ci.yml
```diff
@@ -39,34 +39,79 @@ jobs:
...
-    runs-on: macos-12
+    runs-on: ${{ matrix.runner }}
     env:
       DEVELOPER_DIR: /Applications/Xcode.app
+      XCODE_DESTINATION: 'OS X'
+      NSUnbufferedIO: YES
+    strategy:
+      fail-fast: false
+      matrix:
+        # This matrix runs tests on iOS sim & Mac, on oldest & newest supported Xcodes
```
waahm7, line 50: Wrong comment? This matrix is only for Mac.

Test/AwsCommonRuntimeKitTests/XCBaseTestCase.swift
xiazhvera: We will use them soon for IoT related tests. I will keep them here for now.

waahm7 (approved):

PR #240, opened 2024-03-11
Mqtt5 Client Creation 

xiazhvera:
*Issue #, if available:*

*Description of changes:*


Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -0,0 +1,65 @@
...
+        options.connectOptions?.username?.withByteCursorPointer{ usernameByteCursor in
+            raw_connect_options.username = usernameByteCursor;
+        }
+        options.connectOptions?.password?.withByteCursorPointer{ passwordByteCursor in
+            raw_connect_options.password = passwordByteCursor;
+        }
+                                                                
+        raw_options.connect_options = UnsafePointer(&raw_connect_options);
+
+        raw_options.client_termination_handler = MqttClientTerminationCallback;
```
waahm7, line 55: Probably want to use https://github.com/awslabs/aws-crt-swift/blob/main/Source/AwsCommonRuntimeKit/crt/ShutdownCallbackCore.swift#L11. You will need to retain this callback.

waahm7: Never mind, probably not. If the user passes a closure as a callback, we need to retain it.

Source/AwsCommonRuntimeKit/crt/Utilities.swift
```diff
@@ -121,6 +132,26 @@ extension TimeInterval {
     var millisecond: UInt64 {
         UInt64((self*1000).rounded())
     }
+
+    var millisecondUInt32: UInt32 {
+        UInt32((self*1000).rounded())
```
sbSteveK, line 137: I think we need a bounds check? I'm wondering if we need to do the same for the millisecond() as well since that doesn't check for negatives.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -0,0 +1,31 @@
+///  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+///  SPDX-License-Identifier: Apache-2.0.
+
+import Foundation
+import AwsCMqtt
+import AwsCIo
+
+public class Mqtt5Client {
+     private var rawValue: UnsafeMutablePointer<aws_mqtt5_client>
+     private let clientOptions: MqttClientOptions
```
sbSteveK, line 10: I don't think we need this to be a property within the Mqtt5Client. Can we use it in the init and then let it fall out of scope?

PR #243, opened 2024-03-26
Mqtt5 Connection Bindings

sbSteveK:
Bindings for connect, connack, and disconnect packets


Source/AwsCommonRuntimeKit/mqtt/Mqtt5Types.swift
```diff
@@ -597,6 +597,29 @@ public class PublishReceivedData {
 /// Defines signature of the Publish callback
 public typealias OnPublishReceived = (PublishReceivedData) -> Void
 
+/// Type of a client lifecycle event
+private enum Mqtt5LifecycleEventType: Int {
+
```
xiazhvera, line 602: I'm not sure if we need bind this, as it is only used internally.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -403,6 +423,65 @@ public class DisconnectPacket {
...
+    }
+
+    static func convertFromNative(_ from: UnsafePointer<aws_mqtt5_packet_disconnect_view>?) -> DisconnectPacket? {
+        if let from = from {
+            guard let reasonCode = DisconnectReasonCode(rawValue: Int(from.pointee.reason_code.rawValue))
+            else { fatalError("DisconnectPacket from native missing a reason code.") }
+
+            let sessionExpiryInterval = convertOptionalUInt32(from.pointee.session_expiry_interval_seconds)
+            let sessionExpiryIntervalSeconds: TimeInterval? = sessionExpiryInterval.map { TimeInterval($0) }
+            let reasonString = convertAwsByteCursorToOptionalString(from.pointee.reason_string)
```
xiazhvera, line 468: As `from.pointee` is used multiple places, we could simplify it a little. 
```suggestion
            let view = from.pointee
            let reasonString = convertAwsByteCursorToOptionalString(view.reason_string)
```

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -495,4 +574,59 @@ public class ConnackPacket {
         self.responseInformation = responseInformation
         self.serverReference = serverReference
     }
+
+    static func convertFromNative(_ from: UnsafePointer<aws_mqtt5_packet_connack_view>?) -> ConnackPacket? {
+
+        if let from = from {
+
+            let sessionPresent = from.pointee.session_present
+            let reasonCode = ConnectReasonCode(rawValue: Int(from.pointee.reason_code.rawValue)) ?? .unspecifiedError
```
xiazhvera, line 583: We use `fatalError` in `DisconnectPacket` when the error code is invalid. Would we like to do the same thing here?

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Types.swift
```diff
@@ -625,26 +623,25 @@ public class LifecycleConnectSuccessData {
...
 /// Dataclass containing results of a Connect Failure Lifecycle Event.
-public class LifecycleConnectFailureData {
+public class LifecycleConnectionFailureData {
 
     /// Error which caused connection failure.
     public let crtError: CRTError
 
     /// Data model of an `MQTT5 CONNACK <https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901074>`_ packet.
-    public let connackPacket: ConnackPacket
+    public let connackPacket: ConnackPacket?
```
xiazhvera, line 635: Was it an invalid case that connackPacket is null?

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Types.swift
```diff
@@ -653,9 +650,9 @@ public class LifecycleDisconnectData {
     public let crtError: CRTError
 
     /// Data model of an `MQTT5 DISCONNECT <https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901205>`_ packet.
-    public let disconnectPacket: DisconnectPacket
+    public let disconnectPacket: DisconnectPacket?
 
-    public init (crtError: CRTError, disconnectPacket: DisconnectPacket) {
+    public init (crtError: CRTError, disconnectPacket: DisconnectPacket? = nil) {
```
xiazhvera, line 655: The same questions as above: Could DisconnectPacket be nil?

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Types.swift
```diff
@@ -712,35 +709,72 @@ public class NegotiatedSettings {
...
+        self.subscriptionIdentifiersAvailable = subscriptionIdentifiersAvailable
+        self.sharedSubscriptionsAvailable = sharedSubscriptionsAvailable
+        self.rejoinedSession = rejoinedSession
+        self.clientId = clientId
+    }
+
+    static func convertFromNative(_ from: UnsafePointer<aws_mqtt5_negotiated_settings>?) -> NegotiatedSettings? {
+
+        if let from = from {
+            guard let negotiatedMaximumQos = QoS(rawValue: Int(from.pointee.maximum_qos.rawValue))
```
xiazhvera, line 743: Similar to above: we could improve by 
```
let view = from.pointee
```

xiazhvera (approved):
Some minor suggestions. 
I do have some concern about the fatalErrors and the nil packets. However, I dont have a clear idea about how we handle the errors and logs right now. We probably wanna review the fatalErrors and error handling when we introduce the Logger.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -21,6 +33,29 @@ public class Mqtt5Client {
...
     }
 
+    public func start() throws {
+        let errorCode = aws_mqtt5_client_start(rawValue)
+        if errorCode != 0 {
+            throw CommonRunTimeError.crtError(CRTError(code: errorCode))
+        }
+    }
+
+    public func stop(disconnectPacket: DisconnectPacket? = nil) throws {
```
xiazhvera, line 43: Think about it, as we only have one parameter, we could use '_' so the user does not need explicit call out the parameter name here. 
```suggestion
    public func stop(_ disconnectPacket: DisconnectPacket? = nil) throws {
```

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Types.swift
xiazhvera: Waqar mentioned we could use a private init instead of static function `convertFromNative`. I think that is a good idea.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Types.swift
sbSteveK: Not sure what you mean by invalid case. 

ConnackPacket should be optional here since it's possible for the native client to emit a connection failure lifecycle event that both contains a failed connack (connect was sent and a negative connack received from iot broker) and a nil connack (e.g. connect timed out).

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Types.swift
sbSteveK: DisconnectPacket is only available in scenario where the IoT broker sends a disconnect packet. Other disconnections (e.g. user is shutting down client while connected) results in a disconnect lifecycle event without a disconnect packet.

PR #244, opened 2024-03-26
Mqtt5 Termination Callbacks

xiazhvera:
*Issue #, if available:*

*Description of changes:*


Source/AwsCommonRuntimeKit/mqtt/Mqtt5Types.swift
```diff
@@ -1029,7 +1038,7 @@ public class MqttClientOptions: CStruct {
         }
 
     typealias RawType = aws_mqtt5_client_options
-    func withCStruct<Result>( _ body: (aws_mqtt5_client_options) -> Result) -> Result {
+    func withCStruct<Result>(userData: UnsafeMutableRawPointer?, _ body: (aws_mqtt5_client_options) -> Result) -> Result {
```
sbSteveK, line 1041: While this `userData` will be passed down as individual callback's `userData` property, at this point it will only ever be the callbackCore. It should NOT be anything other than the callbackCore pointer. Should we rename this to be more representative of what is expected at this point?

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Types.swift
```diff
@@ -1104,10 +1109,98 @@ public class MqttClientOptions: CStruct {
...
                 return body(raw_options)
             }
         }
     }
 }
+
+/// Internal Classes
+/// Callback core for event loop callbacks
+class MqttShutdownCallbackCore {
+    let onPublishReceivedCallback: OnPublishReceived?
```
sbSteveK, line 1138: If we are always assigning something to these on init, we can probably make them not optional.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Types.swift
xiazhvera: The function is override the base function for all CStruct requires userData. We could not rename the parameter. I can definitely add more comments to indicate the userData here should be the callbackCore.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Types.swift
sbSteveK: In the init() it looks like if the user does not set a callback, we set it to an empty callback rather than leaving it nil. In that case we're sure that these will never be nil and will always be set removing the need to treat these as optional.

sbSteveK: Separate note: It might be cleaner if we want to always assign a callback to use ?? and a direct assignment. e.g. 
instead of an "if let" with an else block:
if let onPublishReceivedCallback = onPublishReceivedCallback {
            self.onPublishReceivedCallback = onPublishReceivedCallback
        } else {
            /// Pass an empty callback to make manual reference counting easier and avoid null checks.
            self.onPublishReceivedCallback = { (_) -> Void in return }
        }

we use:
self.onPublishReceivedCallback = onPublishReceivedCallback ?? { (_) in return }

PR #248, opened 2024-04-05
Mqtt5 connect tests

sbSteveK:
Test for Mqtt5 Connections

### DIFF
--- a/Test/AwsCommonRuntimeKitTests/mqtt/Mqtt5ClientTests.swift
+++ b/Test/AwsCommonRuntimeKitTests/mqtt/Mqtt5ClientTests.swift
@@ -32,7 +32,9 @@ func onLifecycleEventDisconnectionMinimal(_ : LifecycleDisconnectData){
 
 
 class Mqtt5ClientTests: XCBaseTestCase {
-
+    /*===============================================================
+                     CREATION TEST CASES
+    =================================================================*/
     /*
      * [New-UC1] Happy path. Minimal creation and cleanup
      */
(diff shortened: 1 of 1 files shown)


Test/AwsCommonRuntimeKitTests/mqtt/Mqtt5ClientTests.swift
```diff
@@ -114,4 +118,399 @@ class Mqtt5ClientTests: XCBaseTestCase {
...
+     * [ConnDC-UC4] Direct Connection with mutual TLS
+     */
+
+    func testMqtt5DirectConnectWithMutualTLS() throws {
+
+        let inputHost = try getEnvironmentVarOrSkipTest(environmentVarName: "AWS_TEST_MQTT5_IOT_CORE_HOST")
+        let inputCert = try getEnvironmentVarOrSkipTest(environmentVarName: "AWS_TEST_MQTT5_IOT_CORE_RSA_CERT")
+        let inputKey = try getEnvironmentVarOrSkipTest(environmentVarName: "AWS_TEST_MQTT5_IOT_CORE_RSA_KEY")
+
+        let tlsOptions = try TLSContextOptions.makeMTLS(
```
xiazhvera, line 367: The makeMTLS not working on ios/tvos/watchos right now. Unfortunately, we would need add `#if` to only enable the test on macos and linux. 
Example: https://github.com/awslabs/aws-crt-swift/blob/main/Test/AwsCommonRuntimeKitTests/io/TLSContextTests.swift#L27

Test/AwsCommonRuntimeKitTests/mqtt/Mqtt5ClientTests.swift
```diff
@@ -114,4 +252,666 @@ class Mqtt5ClientTests: XCBaseTestCase {
...
+        let client = try createClient(clientOptions: clientOptions, testContext: testContext)
+        try client.start()
+
+        if testContext.semaphoreConnectionFailure.wait(timeout: .now() + 5) == .timedOut {
+            print("Connection Failure Timed out after 5 seconds")
+            XCTFail("Connection Timed Out")
+        }
+
+        if let failureData = testContext.lifecycleConnectionFailureData {
+            if failureData.crtError.code != Int32(AWS_IO_SOCKET_CONNECTION_REFUSED.rawValue) &&
```
xiazhvera, line 587: Checkout the test doc, the error seems to be the following: 
```
        * error_code == AWS_ERROR_MQTT5_DECODE_PROTOCOL_ERROR (444)
        * error_code == AWS_ERROR_SYS_CALL_FAILURE (direct connect port)
```
However, when I tested it locally, I also get `AWS_IO_SOCKET_CONNECTION_REFUSED` instead. Was the doc outdated?

xiazhvera (approved):

PR #252, opened 2024-04-10
Mqtt5 Subscribe Operation

xiazhvera:
*Issue #, if available:*

*Description of changes:*
- Add subscribe operation
- Update MqttClientPublishRecievedEvents


Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -236,6 +240,55 @@ public class PublishPacket: CStruct {
...
+                identifier?.append(subscription_identifier)
+            }
+
+            let userProperties = convertOptionalUserProperties(
+                count: publishView.user_property_count,
+                userPropertiesPointer: publishView.user_properties)
+
+            guard let qos = QoS(rawValue: Int(publishView.qos.rawValue)) else {
+                fatalError("PublishPacket Received has an invalid qos")
+            }
```
waahm7, line 271: Can we use nativeValue here? Having both is confusing.

waahm7 (commented):
I haven't reviewed everything yet.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -349,6 +484,34 @@ public class SubackPacket {
         self.reasonString = reasonString
         self.userProperties = userProperties
     }
+
+    static func convertFromNative(_ from: UnsafePointer<aws_mqtt5_packet_suback_view>?) -> SubackPacket? {
```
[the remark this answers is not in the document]
waahm7: Same for all `convertFromNative` function.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -58,6 +58,47 @@ public class Mqtt5Client {
...
+            }
+
+            subscribePacket.withCPointer { subscribePacketPointer in
+                var callbackOptions = aws_mqtt5_subscribe_completion_options()
+                callbackOptions.completion_callback = subscribeCompletionCallback
+                callbackOptions.completion_user_data = ContinuationCore(continuation: continuation).passRetained()
+                let result = aws_mqtt5_client_subscribe(rawValue, subscribePacketPointer, &callbackOptions)
+                if result != 0 {
+                    continuation.resume(throwing: CommonRunTimeError.crtError(CRTError(code: -1)))
+                }
```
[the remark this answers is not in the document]
xiazhvera: What's the benefit using guard instead of if?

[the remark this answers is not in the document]
waahm7: It gives a compile time guarantee of not falling through in the case of error. Continuation.resume is not a returning statement but we want to make sure that we don't do anything after resuming the continuation.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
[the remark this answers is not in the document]
xiazhvera: Thining about, it is easier to handle the nil case with the static function instead of the init function. With the init function, we need either explicitly check the nil value or handle exception for the nil case.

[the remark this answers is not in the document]
waahm7: It should be the same. You will need to init?

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -220,14 +220,18 @@ public class PublishPacket: CStruct {
...
                                 raw_publish_view.content_type = cContentType
-                                raw_publish_view.correlation_data = cCorrelationData
                                 raw_publish_view.response_topic = cResponseTopic
 
+                                if let _correlationData = self.correlationData {
+                                    return _correlationData.withAWSByteCursorPointer { cCorrelationData in
+                                        raw_publish_view.correlation_data = UnsafePointer(cCorrelationData)
+                                        return body(raw_publish_view)
+                                    }
+                                }
```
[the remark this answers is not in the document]
xiazhvera: Added a `withOptionalByteCursorPointerFromData` as correlation_data is a `Data` type.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -65,6 +65,49 @@ public class Mqtt5Client {
...
+                continuationCore.continuation.resume(returning: suback)
+
+            }
+
+            subscribePacket.withCPointer { subscribePacketPointer in
+                var callbackOptions = aws_mqtt5_subscribe_completion_options()
+                callbackOptions.completion_callback = subscribeCompletionCallback
+                callbackOptions.completion_user_data = ContinuationCore(continuation: continuation).passRetained()
+                let result = aws_mqtt5_client_subscribe(rawValue, subscribePacketPointer, &callbackOptions)
+                guard result == 0 else {
```
sbSteveK, line 103: 0 -> AWS_OP_SUCCESS

sbSteveK (approved):

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -65,6 +65,49 @@ public class Mqtt5Client {
...
+
+            }
+
+            subscribePacket.withCPointer { subscribePacketPointer in
+                var callbackOptions = aws_mqtt5_subscribe_completion_options()
+                callbackOptions.completion_callback = subscribeCompletionCallback
+                callbackOptions.completion_user_data = ContinuationCore(continuation: continuation).passRetained()
+                let result = aws_mqtt5_client_subscribe(rawValue, subscribePacketPointer, &callbackOptions)
+                guard result == AWS_OP_SUCCESS else {
+                    return continuation.resume(throwing: CommonRunTimeError.crtError(CRTError.makeFromLastError()))
```
waahm7, line 104: Memory leak: Need to release the ContinuationCore here.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -220,22 +220,74 @@ public class PublishPacket: CStruct {
...
+                            contentType) { cResponseTopic, cContentType in
                                 raw_publish_view.content_type = cContentType
-                                raw_publish_view.correlation_data = cCorrelationData
                                 raw_publish_view.response_topic = cResponseTopic
 
-                                return body(raw_publish_view)
+                                return withOptionalByteCursorPointerFromData(to: self.correlationData) {  cCorrelationData in
+                                    if let _cCorrelationData = cCorrelationData {
+                                        raw_publish_view.correlation_data = _cCorrelationData
+                                    }
```
waahm7, line 232: You don't need the if. We will just assign `raw_publish_view.correlation_data` to nil if cCorrelationData is nil.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -287,11 +339,67 @@ public class Subscription {
...
+
+    private var topicFilterBuffer = aws_byte_buf()
+
+    typealias RawType = aws_mqtt5_subscription_view
+    func withCStruct<Result>(_ body: (RawType) -> Result) -> Result {
+        var view = aws_mqtt5_subscription_view()
+        view.qos = self.qos.nativeValue
+        view.no_local = self.noLocal ?? false
+        view.retain_as_published = self.retainAsPublished ?? false
+        if let _retainType = self.retainHandlingType {
```
waahm7, line 357: What does _variableName naming scheme represent? Maybe just get rid of _?
```suggestion
        if let retainType = self.retainHandlingType {
```

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -287,11 +339,67 @@ public class Subscription {
...
+            view.retain_handling_type = _retainType.natvieValue
+        } else {
+            view.retain_handling_type = aws_mqtt5_retain_handling_type(0)
+        }
+
+        return withByteCursorFromStrings(self.topicFilter) { _ in
+            view.topic_filter = aws_byte_cursor_from_buf(&topicFilterBuffer)
+            return body(view)
+        }
```
waahm7, line 366: Why call the with* function if you don't need the result?

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -328,6 +436,32 @@ public class SubscribePacket {
                   subscriptionIdentifier: subscriptionIdentifier,
                   userProperties: userProperties)
     }
+
+    typealias RawType = aws_mqtt5_packet_subscribe_view
+    func withCStruct<Result>(_ body: (RawType) -> Result) -> Result {
+        var raw_subscrbe_view = aws_mqtt5_packet_subscribe_view()
+        return self.subscriptions.withCSubscriptions { subscriptionPointer in
+            raw_subscrbe_view.subscriptions =
+            UnsafePointer<aws_mqtt5_subscription_view>(subscriptionPointer)
```
waahm7, line 445: nit: bad formatting.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -328,6 +436,32 @@ public class SubscribePacket {
...
+            raw_subscrbe_view.subscription_count = self.subscriptions.count
+
+            return withOptionalUserPropertyArray(
+                of: userProperties) { userPropertyPointer in
+
+                    if let _userPropertyPointer = userPropertyPointer {
+                        raw_subscrbe_view.user_property_count = userProperties!.count
+                        raw_subscrbe_view.user_properties =
+                        UnsafePointer<aws_mqtt5_user_property>(_userPropertyPointer)
```
waahm7, line 454: nit: bad format.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -328,6 +436,32 @@ public class SubscribePacket {
...
+        var raw_subscrbe_view = aws_mqtt5_packet_subscribe_view()
+        return self.subscriptions.withCSubscriptions { subscriptionPointer in
+            raw_subscrbe_view.subscriptions =
+            UnsafePointer<aws_mqtt5_subscription_view>(subscriptionPointer)
+            raw_subscrbe_view.subscription_count = self.subscriptions.count
+
+            return withOptionalUserPropertyArray(
+                of: userProperties) { userPropertyPointer in
+
+                    if let _userPropertyPointer = userPropertyPointer {
```
waahm7, line 451: If it works
```suggestion
                    if let userPropertyPointer = userPropertyPointer {
```

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -349,6 +483,34 @@ public class SubackPacket {
         self.reasonString = reasonString
         self.userProperties = userProperties
     }
+
+    public static func convertFromNative(_ from: UnsafePointer<aws_mqtt5_packet_suback_view>?) -> SubackPacket? {
+        if let _from = from {
```
waahm7, line 488: A guard will be better here instead of nesting 20 lines of code plus again, makes more sense to make it a failable init?.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -349,6 +483,34 @@ public class SubackPacket {
...
+            let subackPointer = _from.pointee
+
+            var subackReasonCodes: [SubackReasonCode] = []
+            for i in 0..<subackPointer.reason_code_count {
+                let reasonCodePointer = subackPointer.reason_codes.advanced(by: Int(i)).pointee
+                guard let reasonCode = SubackReasonCode(rawValue: Int(reasonCodePointer.rawValue))
+                else {fatalError("SubackPacket from native has an invalid reason code.")}
```
waahm7, line 495: nit: bad format. Is there a way to enforce/check this?

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -220,22 +220,74 @@ public class PublishPacket: CStruct {
...
                             }
                         }
                     }
                 }
             }
         }
     }
+
+    static func convertFromNative(_ from: UnsafePointer<aws_mqtt5_packet_publish_view>?) -> PublishPacket? {
+        if let _from = from {
```
waahm7, line 244: Same comments as the other one. 
1. Guard will be better
2. If you make this a failable init?, you won't need to make different variables to pass to the init function, you can assign it directly like s`elf.publishView = `.

Test/AwsCommonRuntimeKitTests/mqtt/Mqtt5ClientTests.swift
```diff
@@ -47,6 +47,7 @@ class Mqtt5ClientTests: XCBaseTestCase {
         public var onLifecycleEventConnectionFailure: OnLifecycleEventConnectionFailure?
         public var onLifecycleEventDisconnection: OnLifecycleEventDisconnection?
 
+        public let semaphorePublishReceived:DispatchSemaphore
```
waahm7, line 50: nit: bad format. Can we add more rules to our format checker?

Source/AwsCommonRuntimeKit/crt/Utilities.swift
```diff
@@ -90,6 +90,16 @@ func withOptionalAWSByteCursorFromData<Result>(
     }
 }
 
+func withOptionalByteCursorPointerFromData<Result>(
+    to data: Data?, _ body: (UnsafePointer<aws_byte_cursor>?) throws -> Result) rethrows -> Result {
+    guard let _data = data else {
+        return try body(nil)
+    }
```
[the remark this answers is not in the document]
xiazhvera: Debatable: I feel more safe to have a different name for a different variable type to avoid confusion. If it is not a strong against, I'd prefer keep it like this.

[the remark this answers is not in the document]
waahm7: Plus, I think it's better to shadow because you are not going to use data after the unwrap, so it should be unwrapped throughout the function, and it is not ideal to have both an optional and an unwrapped value in the scope if only one is needed and used.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -65,6 +65,49 @@ public class Mqtt5Client {
...
 
+    /// Tells the client to attempt to subscribe to one or more topic filters.
+    ///
+    /// - Parameters:
+    ///     - subscribePacket: SUBSCRIBE packet to send to the server
+    /// - Returns:
+    ///     - @return Suback packet if the subscription operation succeed otherwise errorCode
+    ///
+    /// - Throws: CommonRuntimeError.crtError
+    public func subscribe(subscribePacket: SubscribePacket) async throws -> SubackPacket {
```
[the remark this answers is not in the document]
xiazhvera: I will leave it for now, as we plan to remove the parameter variable name in a future PR: 
`public func subscribe(_ packet: SubscribePacket) async throws -> SubackPacket {`

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
xiazhvera: Failable init is a good idea. As we have a list of PRs chained after this one, such big update would affect the following PRs. I will have a separate PR to make all of the "convertFromNative" to failable init.

Source/AwsCommonRuntimeKit/crt/Utilities.swift
```diff
@@ -81,15 +81,25 @@ extension Data {
 
 func withOptionalAWSByteCursorFromData<Result>(
     to data: Data?, _ body: (aws_byte_cursor) throws -> Result) rethrows -> Result {
-    guard let _data = data else {
+    guard let data = data else {
```
waahm7, line 84: nit: same for others if it's the same name
```suggestion
    guard let data else {
```

waahm7, line 84: The function name doesn't represent what it does because it will always return a byte cursor. Maybe rename it to 
`withAWSByteCursorFromOptionalData`

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -148,7 +148,7 @@ public class PublishPacket: CStruct {
                 messageExpiryInterval: TimeInterval? = nil,
                 topicAlias: UInt16? = nil,
                 responseTopic: String? = nil,
```
waahm7, line 150: Not part of this PR but we should double check that we are not making anything a String if it can be non-utf8.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -179,63 +179,105 @@ public class PublishPacket: CStruct {
...
+        raw_publish_view.qos = self.qos.nativeValue
         raw_publish_view.retain = retain
         return topic.withByteCursor { topicCustor in
             raw_publish_view.topic =  topicCustor
             return withOptionalAWSByteCursorFromData(to: payload) { cByteCursor in
                 raw_publish_view.payload = cByteCursor
 
                 let _payloadFormatIndicatorInt: aws_mqtt5_payload_format_indicator? =
-                                                                    payloadFormatIndicator?.rawValue ?? nil
+                                                                    payloadFormatIndicator?.nativeValue ?? nil
```
waahm7, line 190: You don't need the ??. `?.` will return nill if it's nil. See https://docs.swift.org/swift-book/documentation/the-swift-programming-language/optionalchaining/

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -179,63 +179,105 @@ public class PublishPacket: CStruct {
...
-
+                    raw_publish_view.payload_format = payloadPointer
+                    raw_publish_view.message_expiry_interval_seconds = messageExpiryIntervalPointer
+                    raw_publish_view.topic_alias = topicAliasPointer
+                        
                     return withOptionalArrayRawPointer(of: subscriptionIdentifiers) { subscriptionPointer in
 
-                        if let _subscriptionPointer = subscriptionPointer {
-                            raw_publish_view.subscription_identifiers = _subscriptionPointer
+                        if let subscriptionPointer = subscriptionPointer {
```
waahm7, line 203: Better to just do `if let subscriptionPointer, subscriptionCount {
}` instead of force unwrap.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -179,63 +179,105 @@ public class PublishPacket: CStruct {
...
+                            raw_publish_view.subscription_identifiers = subscriptionPointer
                             raw_publish_view.subscription_identifier_count = subscriptionIdentifiers!.count
                         }
 
                         return withOptionalUserPropertyArray(
                             of: userProperties) { userPropertyPointer in
-                            if let _userPropertyPointer = userPropertyPointer {
+                            if let userPropertyPointer = userPropertyPointer {
```
waahm7, line 210: Same as above

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -179,63 +179,105 @@ public class PublishPacket: CStruct {
...
                                 raw_publish_view.user_properties =
-                                    UnsafePointer<aws_mqtt5_user_property>(_userPropertyPointer)
+                                    UnsafePointer<aws_mqtt5_user_property>(userPropertyPointer)
                             }
-                            return withOptionalByteCursorPointerFromString(
+                            return withOptionalByteCursorPointerFromStrings(
                             responseTopic,
-                            correlationData,
-                            contentType) { cResponseTopic, cCorrelationData, cContentType in
+                            contentType) { cResponseTopic, cContentType in
```
waahm7, line 217: nit weird format
responseTopic should not align with return and should have at least one tab.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -179,63 +179,105 @@ public class PublishPacket: CStruct {
...
                     }
                 }
             }
         }
     }
+
+    static func convertFromNative(_ from: UnsafePointer<aws_mqtt5_packet_publish_view>?) -> PublishPacket? {
+        guard let from = from else {
+            return nil
+        }
```
waahm7, line 236: if it works
```suggestion
        guard let from  else {
            return nil
        }
```

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -179,63 +179,105 @@ public class PublishPacket: CStruct {
...
+        let publishView = from.pointee
+
+        let payload: Data = Data(bytes: publishView.payload.ptr, count: publishView.payload.len)
```
waahm7, line 239: nit / debatable: If the type is really obvious, I think it makes it cleaner to just omit it or at least keep it consistent.
Same comment for others.
```suggestion
        let payload = Data(bytes: publishView.payload.ptr, count: publishView.payload.len)
```

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -179,63 +179,105 @@ public class PublishPacket: CStruct {
...
+
+        let payloadFormatIndicator: PayloadFormatIndicator? = publishView.payload_format != nil ?
+        PayloadFormatIndicator(rawValue: Int(publishView.payload_format.pointee.rawValue)) : nil
+
+        let messageExpiryInterval = convertOptionalUInt32(publishView.message_expiry_interval_seconds)
```
waahm7, line 244: Not relevant to this PR, but I would prefer that we get rid of all the convert* (either removing them or refactoring them) by the final merge. We should avoid introducing new terminologies. For example, if we need a function like this, it should be an extension over Optional, where Element == UInt32 instead of a separate function.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -328,6 +424,32 @@ public class SubscribePacket {
...
+        var raw_subscrbe_view = aws_mqtt5_packet_subscribe_view()
+        return self.subscriptions.withCSubscriptions { subscriptionPointer in
+            raw_subscrbe_view.subscriptions =
+                UnsafePointer<aws_mqtt5_subscription_view>(subscriptionPointer)
+            raw_subscrbe_view.subscription_count = self.subscriptions.count
+
+            return withOptionalUserPropertyArray(
+                of: userProperties) { userPropertyPointer in
+
+                    if let userPropertyPointer = userPropertyPointer {
```
waahm7, line 439:
```suggestion
                    if let userPropertyPointer, userProperties {
```

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -349,6 +471,36 @@ public class SubackPacket {
         self.reasonString = reasonString
         self.userProperties = userProperties
     }
+
+    public static func convertFromNative(_ from: UnsafePointer<aws_mqtt5_packet_suback_view>?) -> SubackPacket? {
+        
+        guard let from = from else {
+            return nil
+        }
```
waahm7, line 479:
```suggestion
        guard let from else {
            return nil
        }
```

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -435,17 +587,15 @@ public class DisconnectPacket: CStruct {
...
-                raw_disconnect_view.session_expiry_interval_seconds = _sessionExpiryIntervalPointer
-            }
-
+            raw_disconnect_view.session_expiry_interval_seconds = sessionExpiryIntervalPointer
+            
             return withOptionalUserPropertyArray(
                 of: userProperties) { userPropertyPointer in
 
-                if let _userPropertyPointer = userPropertyPointer {
+                if let userPropertyPointer = userPropertyPointer {
```
waahm7, line 595:
```suggestion
                if let userPropertyPointer {
```

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -460,29 +610,30 @@ public class DisconnectPacket: CStruct {
...
-
-            let disconnectPacket = DisconnectPacket(
-                reasonCode: reasonCode,
-                sessionExpiryInterval: sessionExpiryIntervalSeconds,
-                reasonString: reasonString,
-                serverReference: serverReference,
-                userProperties: userProperties
-            )
-            return disconnectPacket
+        guard let from = from else {
```
waahm7, line 613:
```suggestion
        guard let from else {
```

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -460,29 +610,30 @@ public class DisconnectPacket: CStruct {
...
+            return nil
         }
-        return nil
+        let disconnectView = from.pointee
+        guard let reasonCode = DisconnectReasonCode(rawValue: Int(disconnectView.reason_code.rawValue))
+        else { fatalError("aws_mqtt5_packet_disconnect_view from native missing a reason code.") }
```
waahm7, line 618: nit: bad format

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -65,9 +65,53 @@ public class Mqtt5Client {
...
+
     public func close() {
         self.callbackCore.close()
         aws_mqtt5_client_release(rawValue)
         rawValue = nil
     }
 }
+
+// The completion callback to invoke when an ack is received in native
+private func subscribeCompletionCallback(
```
sbSteveK, line 102: Does this work? This callback doesn't seem to fit the native callback signature. (packet_type, packet, error_code, context)

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -460,29 +610,30 @@ public class DisconnectPacket: CStruct {
...
+
+        let sessionExpiryInterval = convertOptionalUInt32(disconnectView.session_expiry_interval_seconds)
+        let sessionExpiryIntervalSeconds: TimeInterval? = sessionExpiryInterval.map { TimeInterval($0) }
+        let reasonString = convertAwsByteCursorToOptionalString(disconnectView.reason_string)
```
waahm7, line 622: Don't need this function, Utilities.swift already has a helper extension. You can just do `disconnectView.reason_string.toOptionalString()`.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -578,59 +729,59 @@ public class ConnackPacket {
...
 
     static func convertFromNative(_ from: UnsafePointer<aws_mqtt5_packet_connack_view>?) -> ConnackPacket? {
+        guard let from = from else {
+            return nil
+        }
+        let connackView = from.pointee
+
+        let sessionPresent = connackView.session_present
+        guard let reasonCode = ConnectReasonCode(rawValue: Int(connackView.reason_code.rawValue))
+        else { fatalError("aws_mqtt5_packet_connack_view from native missing a reason code.") }
```
waahm7, line 739: bad formart, else is at the same level of the other code.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -578,59 +729,59 @@ public class ConnackPacket {
...
+        let sessionExpiryInterval = (connackView.session_expiry_interval?.pointee).map { TimeInterval($0) }
```
waahm7, line 740: This can result in nil.map, does that work?

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Types.swift
```diff
@@ -442,21 +449,18 @@ extension ClientOperationQueueBehaviorType {
 
 /// Optional property describing a PUBLISH payload's format.
 /// Enum values match `MQTT5 spec <https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901111>`__ encoding values.
-public enum PayloadFormatIndicator {
+public enum PayloadFormatIndicator: Int {
```
waahm7, line 452: Might want to re-discuss the API with Bret after the discussion in the Ops meeting if this is a good idea.

waahm7: If it is, we should make a struct which contains data and format for the Payload instead of having two separate connected variables.

sbSteveK: I think this enum will be extremely rarely used or set in either direction and is fully separate from payload in all other language implementations. Generally, it'll be ignored by the customer. The payload is a Data type and that feels sufficient.

waahm7: Do we even need this indicator? We should just pass the bytes as it is.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
xiazhvera: As userProperties is a class member, I dont think it worked, while I tried with following:
```
if let userPropertyPointer,
   let userPropertyCount = self.userProperties?.count {

}
```

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
xiazhvera: This change will be addressed in a separate PR.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Types.swift
sbSteveK: I don't think we need it since like you said, it'll always be bytes that can be converted to utf8 strings... but it's also part of the mqtt5 spec and expected to be there. It's possible someone may want to check its value for... something?

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
xiazhvera: `map` works for Optional: https://developer.apple.com/documentation/swift/optional/map(_:)

Issue #251, opened 2024-04-10
AwsCommonRuntimeKit potentially needs a privacy manifest


phantumcode:
### Describe the bug

Issue originally created in the amplify-swift repo by @rowanlsnapper https://github.com/aws-amplify/amplify-swift/issues/3603

I'm working through some privacy manifest issues with an app that I'm working on, and the last remaining required reason API warning that I'm still dealing with is "file timestamps" (https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278393).

I'm confident that the app itself is not directly using any of the mentioned file timestamp APIs, and I've worked through the dependencies to see which of them could be using these without proper privacy manifest declarations. I notice that AwsCommonRuntimeKit is using fstat(_:,_:) in a couple of C source files, and it looks like the module doesn't declare this in a privacy manifest.

Could this please be investigated and actioned if necessary.

### Expected Behavior

App Store Connect does not send an email advising undeclared required reason API usage, app passes validation.

### Current Behavior

Unknown

### Reproduction Steps

1. Create an app that includes `amplify-swift` using SPM (in my case am linking `Amplify`, `AWSPluginsCore`, and `AWSCognitoAuthPlugin` to my app target).
2. Submit app to the App Store for RR API validation (can be triggered by assigning a build to a TestFlight external tester group).
3. Observe App Store Connect sends an email advising undeclared required reason API usage.

### aws-crt-swift version used

0.26.1

### Compiler and Version used

Swift 5.7

### Operating System and version

iOS

sichanyoo:
Thank you for transferring the ticket @phantumcode. And thank you for submitting the original ticket @rowanlsnapper. 

We already have a ticket that tracks this in aws-sdk-swift: https://github.com/awslabs/aws-sdk-swift/issues/1191. 

I'll pick up this task now; all updates / PRs will be posted to the ticket in our aws-sdk-swift repo.

Closing the ticket as it's tracked in https://github.com/awslabs/aws-sdk-swift/issues/1191.

PR #254, opened 2024-04-18
chore: Add a privacy manifest

sichanyoo:
*Issue #, if available:*
https://github.com/awslabs/aws-crt-swift/issues/251

*Description of changes:*
Adds a privacy manifest file for `AwsCommonRuntimeKit`.


Source/AwsCommonRuntimeKit/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
waahm7, line 1: trivial/debatable: Would it be possible to place this in root of the project instead of Source/*?

waahm7 (approved):

Source/AwsCommonRuntimeKit/PrivacyInfo.xcprivacy
waahm7: I guess different modules might require different privacy manifest so it might make more sense to keep it here.

PR #255, opened 2024-04-18
Mqtt5 unsubscribe

sbSteveK:
Unsubscribe op and tests

*Description of changes:*

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -7,8 +7,6 @@ analyzer_rules:
 
 disabled_rules:
   - trailing_whitespace
-  - todo
-  - identifier_name
   - compiler_protocol_init
   - function_parameter_count
   - multiple_closures_with_trailing_closure
@@ -46,7 +44,7 @@ opening_brace: error
 return_arrow_whitespace: error
 statement_position:
   severity: error
-todo: warning
 trailing_semicolon: error
 vertical_parameter_alignment: error
-vertical_parameter_alignment_on_call: true
+vertical_parameter_alignment_on_call:
+  severity: error
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -19,11 +19,11 @@ disabled_rules:
 
 opt_in_rules:
   - vertical_parameter_alignment_on_call
-  - opening_brace
   - empty_count
 
 # configurable rules can be customized from this configuration file
 force_cast: warning
+opening_brace: error
 closing_brace: error
 colon:
   severity: error
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -15,6 +15,7 @@ disabled_rules:
   - syntactic_sugar
   - unused_capture_list
   - blanket_disable_command
+  - identifier_name
 
 opt_in_rules:
   - vertical_parameter_alignment_on_call
@@ -31,11 +32,6 @@ empty_enum_arguments: error
 function_body_length:
   warning: 100
   error: 150
-identifier_name:
-  excluded:
-    - id
-    - of
-    - or
 line_length:
   warning: 140
   error: 160
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -17,8 +17,9 @@ disabled_rules:
   - blanket_disable_command
 
 opt_in_rules:
-  - empty_count
   - vertical_parameter_alignment_on_call
+  - opening_brace
+  - empty_count
 
 # configurable rules can be customized from this configuration file
 force_cast: warning
@@ -26,7 +27,6 @@ closing_brace: error
 colon:
   severity: error
 comma: error
-empty_count: warning
 empty_enum_arguments: error
 function_body_length:
   warning: 100
@@ -40,7 +40,6 @@ line_length:
   warning: 140
   error: 160
   ignores_comments: true
-opening_brace: error
 return_arrow_whitespace: error
 statement_position:
   severity: error
--- a/Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
+++ b/Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
@@ -122,8 +122,9 @@ public class Mqtt5Client {
                     return continuationCore.continuation.resume(throwing: CommonRunTimeError.crtError(CRTError(code: errorCode)))
                 }
 
-                guard let unsuback = UnsubackPacket.convertFromNative(unsubackPacket)
-                else { fatalError("Unsuback missing in the Unsubscribe completion callback.") }
+                guard let unsuback = UnsubackPacket.convertFromNative(unsubackPacket) else {
+                    fatalError("Unsuback missing in the Unsubscribe completion callback.")
+                }
 
                 continuationCore.continuation.resume(returning: unsuback)
             }
@@ -133,7 +134,7 @@ public class Mqtt5Client {
                 callbackOptions.completion_callback = unsubscribeCompletionCallback
                 callbackOptions.completion_user_data = ContinuationCore(continuation: continuation).passRetained()
                 let result = aws_mqtt5_client_unsubscribe(rawValue, unsubscribePacketPointer, &callbackOptions)
-                guard result == 0 else {
+                guard result == AWS_OP_SUCCESS else {
                     return continuation.resume(throwing: CommonRunTimeError.crtError(CRTError.makeFromLastError()))
                 }
             }
--- a/Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
+++ b/Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
@@ -63,7 +63,8 @@ public class Mqtt5Client {
         return try await withCheckedThrowingContinuation { continuation in
 
             // The completion callback to invoke when an ack is received in native
-            func subscribeCompletionCallback(subackPacket: UnsafePointer<aws_mqtt5_packet_suback_view>?, errorCode: Int32, userData: UnsafeMutableRawPointer?) {
+            func subscribeCompletionCallback(
+                subackPacket: UnsafePointer<aws_mqtt5_packet_suback_view>?, errorCode: Int32, userData: UnsafeMutableRawPointer?) {
                 let continuationCore = Unmanaged<ContinuationCore<SubackPacket>>.fromOpaque(userData!).takeRetainedValue()
                 if errorCode == 0 {
                     guard let suback = SubackPacket.convertFromNative(subackPacket)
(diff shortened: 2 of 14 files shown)


Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -107,7 +106,37 @@ public class Mqtt5Client {
...
+            func unsubscribeCompletionCallback(unsubackPacket: UnsafePointer<aws_mqtt5_packet_unsuback_view>?,
+                                               errorCode: Int32,
+                                               userData: UnsafeMutableRawPointer?) {
+                let continuationCore = Unmanaged<ContinuationCore<UnsubackPacket>>.fromOpaque(userData!).takeRetainedValue()
+
+                guard errorCode == AWS_OP_SUCCESS else {
+                    return continuationCore.continuation.resume(throwing: CommonRunTimeError.crtError(CRTError(code: errorCode)))
+                }
+
+                guard let unsuback = UnsubackPacket.convertFromNative(unsubackPacket)
```
bretambrose, line 125: I like the other format style where it's

```
guard <condition> else {
   <return failure>
}
```

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -107,7 +106,37 @@ public class Mqtt5Client {
...
+                else { fatalError("Unsuback missing in the Unsubscribe completion callback.") }
```
bretambrose, line 126: Adding on, can we look into using a formatter like swift-format?

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -556,35 +557,97 @@ public class UnsubscribePacket {
...
+            }
+            return body(raw_unsubscribe_view)
+        }
+    }
+
+    func convertTopicFilters(_ topicFilters: [String]) -> UnsafeMutablePointer<aws_byte_cursor>? {
+        let cArray = UnsafeMutablePointer<aws_byte_cursor>.allocate(capacity: topicFilters.count)
+
+        for (index, string) in topicFilters.enumerated() {
+            let data = string.data(using: .utf8)!
```
bretambrose, line 589: Concerns about naked unwraps.  We never want to crash on checkable validation errors.

Test/AwsCommonRuntimeKitTests/mqtt/Mqtt5ClientTests.swift
```diff
@@ -6,6 +6,10 @@ import Foundation
 import AwsCMqtt
 @testable import AwsCommonRuntimeKit
 
+enum TimeoutError: Error {
```
bretambrose, line 9: Per discussion, feel like there's hackiness/weirdness related to timeouts and async that shouldn't be necessary in a modern language and test framework

Test/AwsCommonRuntimeKitTests/mqtt/Mqtt5ClientTests.swift
```diff
@@ -843,17 +930,20 @@ class Mqtt5ClientTests: XCBaseTestCase {
             XCTAssertEqual(negotiatedSettings.maximumQos, QoS.atLeastOnce)
         } else {
             XCTFail("Missing negotiated settings")
+            return
```
bretambrose, line 933: Also per discussion, feels like there's a tremendous amount of boilerplate repeated across different sets of tests.  Much broader that existing review obviously.

Test/AwsCommonRuntimeKitTests/mqtt/Mqtt5ClientTests.swift
```diff
@@ -904,27 +995,267 @@ class Mqtt5ClientTests: XCBaseTestCase {
...
+        let tlsContext = try TLSContext(options: tlsOptions, mode: .client)
+
+        let clientIDPublisher = createClientId() + "Publisher"
+        let topic = "test/MQTT5_Binding_Swift_" + UUID().uuidString
+        let willPacket = PublishPacket(qos: .atLeastOnce, topic: topic, payload: "TEST WILL".data(using: .utf8))
+
+        let connectOptionsPublisher = MqttConnectOptions(clientId: clientIDPublisher, will: willPacket)
+        let clientOptions = MqttClientOptions(
+            hostName: inputHost,
+            port: UInt32(8883),
```
bretambrose, line 1119: Can the port be an environment variable too?

Test/AwsCommonRuntimeKitTests/mqtt/Mqtt5ClientTests.swift
```diff
@@ -904,27 +995,267 @@ class Mqtt5ClientTests: XCBaseTestCase {
...
+        }
+
+        let topic = "test/MQTT5_Binding_Swift_" + UUID().uuidString
+        let subscribePacket = SubscribePacket(topicFilter: topic, qos: QoS.atLeastOnce, noLocal: false)
+
+        let subackPacket: SubackPacket =
+            try await withTimeout(client: client, seconds: 2, operation: {
+                try await client.subscribe(subscribePacket: subscribePacket)
+            })
+        print("SubackPacket received with result \(subackPacket.reasonCodes[0])")
```
bretambrose, line 1055: Are prints normal in tests?

sbSteveK: Only if you're an eventstreamrpc test >,<

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -107,7 +106,38 @@ public class Mqtt5Client {
...
+                continuationCore.continuation.resume(returning: unsuback)
+            }
+
+            unsubscribePacket.withCPointer { unsubscribePacketPointer in
+                var callbackOptions = aws_mqtt5_unsubscribe_completion_options()
+                callbackOptions.completion_callback = unsubscribeCompletionCallback
+                callbackOptions.completion_user_data = ContinuationCore(continuation: continuation).passRetained()
+                let result = aws_mqtt5_client_unsubscribe(rawValue, unsubscribePacketPointer, &callbackOptions)
+                guard result == AWS_OP_SUCCESS else {
+                    return continuation.resume(throwing: CommonRunTimeError.crtError(CRTError.makeFromLastError()))
```
xiazhvera, line 138: Just realize if the operation failed, we should release the continuationCore as its reference was acquired by `passRetained()`. I updated the subscribe and the publish operation to release it: 
https://github.com/awslabs/aws-crt-swift/blob/27cd454c3a72ecaaeed4513da19a1a2be18fad03/Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift#L104

Test/AwsCommonRuntimeKitTests/mqtt/Mqtt5ClientTests.swift
```diff
@@ -550,20 +550,17 @@ class Mqtt5ClientTests: XCBaseTestCase {
         if testContext.semaphoreConnectionFailure.wait(timeout: .now() + 5) == .timedOut {
             print("Connection Failure Timed out after 5 seconds")
             XCTFail("Connection Timed Out")
+            return
```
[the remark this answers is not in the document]
sbSteveK: XCTFail sets the test as failed but continues to run. Calling return exits the test as soon as we register it as failed.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
sbSteveK: We are planning to address these as well as any other Swift -> Native validation checks per operation in a separate branch.

Test/AwsCommonRuntimeKitTests/mqtt/Mqtt5ClientTests.swift
sbSteveK: I've dug some more into this and we could use an `XCTestExpectation` and wrap the async functions within a task that calls `fulfill` on the expectation when it completes and times out when it doesn't but the underlying problem that's being hacked around is that you can't cancel the task once it's awaiting something from native to complete since native doesn't understand the idea of an async task being cancelled. If we wait for the expectation and it times out, it will fail the test but until native completes the continuation, the individual test will continue to run until the test-wide timeout closes everything down. 

I'm open to changing to a hacky way of using expectations in which a timed out expectation results in us manually closing the client but that's not really different from what we're doing now. (and might result in something more complicated to use across all operation calls) At this point, the current implementation is running the test and failing appropriately when it's breaking in the way we expect it to. Unsure it's worth the time to figure out a cleaner way atm but we could revisit this as a later task.

xiazhvera (approved):

xiazhvera:
The swiftlint change looks good to me, while I feel we should call it out and have Waqar have a quick look on it. 
Other than that, the PR looks good to me.

.swiftlint.yml
```diff
@@ -17,36 +15,31 @@ disabled_rules:
   - syntactic_sugar
   - unused_capture_list
   - blanket_disable_command
+  - identifier_name
 
 opt_in_rules:
-  - empty_count
   - vertical_parameter_alignment_on_call
+  - opening_brace
```
waahm7, line 22: I think error severity is correct for this rule, instead of the default `warning` severity.

.swiftlint.yml
```diff
@@ -7,8 +7,6 @@ analyzer_rules:
 
 disabled_rules:
   - trailing_whitespace
-  - todo
```
waahm7, line 10: Why enable the todo rule? I think we will never get rid of all the todos in our repo.

.swiftlint.yml
sbSteveK: There were a number of places where certain options were listed in two separate places resulting in a weird override.

.swiftlint.yml
sbSteveK: We can re-disable this later. I'm leaving it in for now because we've got a lot of todos that we actually want to tackle before merging mqtt to main.

.swiftlint.yml
waahm7: Yeah, but I think this change unintentionally lowered the severity of `opening_brace` from `error` to `warning`.

.swiftlint.yml
waahm7: I am fine with that, however, this will just create noise in every PR. We should just look for todos in the final PR when reviewing.
image

xiazhvera (approved):

PR #258, opened 2024-04-29
remove Int from enums that don't have underlying int values

sbSteveK:
Remove Int inheritance in mqtt enums that don't have underlying int values

### DIFF
--- a/Source/AwsCommonRuntimeKit/mqtt/Mqtt5Types.swift
+++ b/Source/AwsCommonRuntimeKit/mqtt/Mqtt5Types.swift
@@ -20,6 +20,13 @@ public enum QoS: Int {
 
 }
 
+internal extension QoS {
+    /// Returns the native representation of the Swift enum
+    var nativeValue: aws_mqtt5_qos {
+        return aws_mqtt5_qos(rawValue: UInt32(self.rawValue))
+    }
+}
+
 /// Server return code for connect attempts.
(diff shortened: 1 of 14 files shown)


Source/AwsCommonRuntimeKit/mqtt/Mqtt5Types.swift
```diff
@@ -442,91 +466,135 @@ public enum ClientOperationQueueBehaviorType {
...
 }
 
 /// An enumeration that controls how the client applies topic aliasing to outbound publish packets.
 /// Topic alias behavior is described in `MQTT5 Topic Aliasing <https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901113>`_
-public enum OutboundTopicAliasBehaviorType: Int {
+public enum OutboundTopicAliasBehaviorType {
     /// Maps to Disabled.  This keeps the client from being broken (by default) if the broker
     /// topic aliasing implementation has a problem.
-    case `default` = 0
+    case `default`
```
xiazhvera, line 544: Would we consider to rename `default` to other name to avoid confusing between swift default keyword?

xiazhvera (approved):
I have some concern on `default`. Other than that, looks good to me.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Types.swift
sbSteveK: changed it to `defaultBehavior`

PR #260, opened 2024-05-02
Mqtt5 Websocket

xiazhvera:
*Issue #, if available:*

*Description of changes:*


Source/AwsCommonRuntimeKit/auth/credentials/CredentialsProvider.swift
```diff
@@ -294,6 +295,9 @@ extension CredentialsProvider.Source {
             chainDefaultOptions.bootstrap = bootstrap.rawValue
             chainDefaultOptions.profile_collection_cached = fileBasedConfiguration.rawValue
             chainDefaultOptions.shutdown_options = shutdownCallbackCore.getRetainedCredentialProviderShutdownOptions()
+            if let tlsContext {
```
sbSteveK, line 298: Trivial: can we skip the let and set it like this
chainDefaultOptional.tls_ctx = tlsContext?.rawValue

I think that checks whether tlsContext is nil or not before attempting to use rawValue.

sbSteveK (approved):

PR #262, opened 2024-05-03
Mqtt5 Init From Pointer

sbSteveK:
Initialize packets using a convenience init() instead of a static function.

### DIFF
--- a/Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
+++ b/Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
@@ -349,41 +349,41 @@ internal func MqttClientHandleLifecycleEvent(_ lifecycleEvent: UnsafePointer<aws
 }
 
 internal func MqttClientHandlePublishRecieved(
-    _ publishPacketView: UnsafePointer<aws_mqtt5_packet_publish_view>?,
-    _ userData: UnsafeMutableRawPointer?) {
-    let callbackCore = Unmanaged<MqttCallbackCore>.fromOpaque(userData!).takeUnretainedValue()
+    _ publish: UnsafePointer<aws_mqtt5_packet_publish_view>?,
+    _ user_data: UnsafeMutableRawPointer?) {
+    let callbackCore = Unmanaged<MqttCallbackCore>.fromOpaque(user_data!).takeUnretainedValue()
 
     // validate the callback flag, if flag is false, return
     callbackCore.rwlock.read {
         if callbackCore.callbackFlag == false { return }
-        if let publishPacketView {
-            let publishPacket = PublishPacket(publishPacketView)
+        if let publish {
+            let publishPacket = PublishPacket(publish)
             let publishReceivedData = PublishReceivedData(publishPacket: publishPacket)
             callbackCore.onPublishReceivedCallback(publishReceivedData)
         } else {
-            fatalError("MqttClientHandlePublishRecieved called with null publishPacketView")
+            fatalError("MqttClientHandlePublishRecieved called with null publish")
         }
     }
 }
 
 internal func MqttClientWebsocketTransform(
-    _ rawHttpMessage: OpaquePointer?,
-    _ userData: UnsafeMutableRawPointer?,
-    _ completeFn: (@convention(c) (OpaquePointer?, Int32, UnsafeMutableRawPointer?) -> Void)?,
-    _ completeCtx: UnsafeMutableRawPointer?) {
+    _ request: OpaquePointer?,
+    _ user_data: UnsafeMutableRawPointer?,
+    _ complete_fn: (@convention(c) (OpaquePointer?, Int32, UnsafeMutableRawPointer?) -> Void)?,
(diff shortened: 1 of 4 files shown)


Source/AwsCommonRuntimeKit/mqtt/Mqtt5Options.swift
```diff
@@ -547,42 +547,34 @@ public class NegotiatedSettings {
...
-            serverKeepAlive: negotiatedServerKeepAlive,
-            retainAvailable: negotiatedRetainAvailable,
-            wildcardSubscriptionsAvailable: negotiatedWildcardSubscriptionsAvailable,
-            subscriptionIdentifiersAvailable: negotiatedSubscriptionIdentifiersAvailable,
-            sharedSubscriptionsAvailable: negotiatedSharedSubscriptionsAvailable,
-            rejoinedSession: negotiatedRejoinedSession,
-            clientId: negotiatedClientId)
-
-        return negotiatedSettings
+    internal convenience init(_ from: UnsafePointer<aws_mqtt5_negotiated_settings>){
```
waahm7, line 550: If you make it an actual init instead of a convenience init, you won't need 20 different variables and can just do self.negotiatedSettings = etc, plus it is internal by default.

sbSteveK: I felt like it might be clearer intent to explicitly call it internal. I'm also calling the existing init in case we ever change anything there so we know to change it in both as a precaution. I'd rather do the assigning in a singular place but am open to doing a manual assign and not make it call itself if that makes more sense.

waahm7, line 550: I think from is too generic, and since Swift allows us to have the same functions with different parameter names, it makes more sense to keep the parameter names explicit.

```suggestion
    internal convenience init(_ nativePointer: UnsafePointer<aws_mqtt5_negotiated_settings>){
```

sbSteveK: Maybe if it's making a copy and then using the copy to call init which will then make another copy, that's bad and we should just assign it instead of it being convenience. Was planning on looking into how memory is handled across all these as next task.

sbSteveK: That's a good point. I was copy pasting these arguments but yea, we should definitely give them unique names.

waahm7: The init function is just 1 on 1 assignment, so I would not expect it to change in the future. If we ever add new properties to this class, Swift provides a compile time grantee that we will initialize it in both places. I think keeping it separate makes it more cleaner.

waahm7: Also it will make sure that we don't forget setting any new properties in second init function. With current approach, if we add a new property the convenience init will just start assigning that to NULL but if they are separate we will get a compile time error to handle it in the second init function as well.

waahm7: Just saw that the public init requires each member to be passed in so we won't run into the problem of forgetting to set it in the init function. However, we should probably assign some default to these members; otherwise, using it will be very cumbersome to have to pass each parameter individually.

sbSteveK: Looking into it further, I think I'm going to separate the inits and not make the native->swift one a convenience init.

Negotiated Settings is only ever created by the CRT and all fields must be present in the negotiated settings resulting from a connack to the server. There should be no default values and everything is to be explicitly set by the time it reaches here.

waahm7: > Negotiated Settings is only ever created by the CRT 
Then, we should make the other init internal as well.

xiazhvera (approved):
Trivial formatting note.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -278,111 +278,111 @@ public class Mqtt5Client {
...
     }
 }
 
 internal func MqttClientWebsocketTransform(
-    _ rawHttpMessage: OpaquePointer?,
-    _ userData: UnsafeMutableRawPointer?,
-    _ completeFn: (@convention(c) (OpaquePointer?, Int32, UnsafeMutableRawPointer?) -> Void)?,
-    _ completeCtx: UnsafeMutableRawPointer?) {
+    _ request: OpaquePointer?,
+    _ user_data: UnsafeMutableRawPointer?,
```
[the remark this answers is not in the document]
sbSteveK: I changed all calls that are coming from native to have arguments that reflect the ones expected in the native function signatures for clarity when working with callbacks. If we change it to camelcase here we should probably do it across all callback functions. Open to going either way as long as we're consistent.

PR #263, opened 2024-05-13
Mqtt5 Client Close() and ReadWrite Lock

xiazhvera:
*Issue #, if available:*

*Description of changes:*
- Added `Close()` function. The client could now call "close()" to release the native client.
- Cleanup the read write lock. 


Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -259,43 +313,69 @@ public class Mqtt5Client {
...
             }
         }
     }
 
+    /// Discard all operations and cleanup the client. It is MANDATORY function to call to release the client.
     public func close() {
-        self.callbackCore.close()
-        aws_mqtt5_client_release(rawValue)
-        rawValue = nil
+        self.rwlock.write {
```
sfod, line 334: Is it safe to call `close()` from a callback or it'll block indefinitely?

xiazhvera: No worry. The callbacks will be scheduled in an async task, therefore, the lock would not block the thread in that case. 

PublishReceived Example: 
https://github.com/awslabs/aws-crt-swift/blob/64afcfe69b658661cae87a599f0a09d6f20db01b/Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift#L444

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -121,44 +121,69 @@ public typealias OnWebSocketHandshakeInterceptComplete = (HTTPRequestBase, Int32
...
+    internal let onLifecycleEventAttemptingConnect: OnLifecycleEventAttemptingConnect
+    internal let onLifecycleEventConnectionSuccess: OnLifecycleEventConnectionSuccess
+    internal let onLifecycleEventConnectionFailure: OnLifecycleEventConnectionFailure
+    internal let onLifecycleEventDisconnection: OnLifecycleEventDisconnection
+    // The websocket interceptor could be nil if the websocket is not in use
+    internal let onWebsocketInterceptor: OnWebSocketHandshakeIntercept?
+    internal let rwlock = ReadWriteLock()
+    internal var callbackFlag = true
+
+    /// Creates a Mqtt5Client instance using the provided Mqtt5ClientOptions. Once the Mqtt5Client is created,
```
sbSteveK, line 144: Trivial: "Creates a Mqtt5Client instance using the provided Mqtt5ClientOptions." We can probably remove the first sentence about creation as it's just how an init() is expected to work.

"Once the Mqtt5Client is created, changing the settings will not cause a change in already created Mqtt5Client's." We can also remove this sentence as all options in Mqtt5ClientOptions are const. 

"Once created, it is MANDATORY to call `close()` to clean up the Mqtt5Client resource" Remove "Once created, " and just state that it is mandatory to call `close()`

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -259,43 +313,69 @@ public class Mqtt5Client {
...
+        }
+    }
+
+    /////////////////////////////////////////////////////////
+    // helper functions for self retained reference
+    private func callbackUserData() -> UnsafeMutableRawPointer {
+        return Unmanaged<Mqtt5Client>.passRetained(self).toOpaque()
+    }
+
+    private func release() {
```
sbSteveK, line 347: This is only ever called in one place if we fail to create a new native client. Let's just execute this code there instead of having it as a separate function.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -259,43 +313,69 @@ public class Mqtt5Client {
...
+        self.rwlock.write {
+            self.callbackFlag = false
+            aws_mqtt5_client_release(rawValue)
+            self.rawValue = nil
+        }
+    }
+
+    /////////////////////////////////////////////////////////
+    // helper functions for self retained reference
+    private func callbackUserData() -> UnsafeMutableRawPointer {
```
sbSteveK, line 343: This is only ever called in one place at the creation of the native client to be used as user_data. Let's not make this a function that needs to be looked up for a singular use case.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -167,8 +192,21 @@ public class Mqtt5Client {
...
+    ///
+    /// - Parameters:
+    ///     - disconnectPacket: (optional) Properties of a DISCONNECT packet to send as part of the shutdown
+    ///     process. When disconnectPacket is null, no DISCONNECT packets will be sent.
+    ///
+    /// - Throws: CommonRuntimeError.crtError
     public func stop(disconnectPacket: DisconnectPacket? = nil) throws {
-        if rawValue != nil {
+        try self.rwlock.read {
+            // validate the client in case close() is called.
```
sbSteveK, line 205: trivial: "Validate `close()` has not been called on client."

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -259,43 +313,69 @@ public class Mqtt5Client {
...
         fatalError("MqttClientLifecycleEvents was called from native without an aws_mqtt5_client_lifecycle_event.")
     }
-    let callbackCore = Unmanaged<MqttCallbackCore>.fromOpaque(lifecycleEvent.pointee.user_data).takeUnretainedValue()
+    let client = Unmanaged<Mqtt5Client>.fromOpaque(userData).takeUnretainedValue()
     let crtError = CRTError(code: lifecycleEvent.pointee.error_code)
 
     // validate the callback flag, if flag is false, return
-    callbackCore.rwlock.read {
-        if callbackCore.callbackFlag == false { return }
+    client.rwlock.read {
```
sbSteveK, line 369: Trivial: Should we move `let client` and `let crtError` into the client lock because if close has been called, I think trying to check the callbackFlag might be accessing potentially bad memory? Why are we using the flag check here instead of doing a nil check on assigning client?

sbSteveK (approved):

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
xiazhvera: The lock is used to protect the flag and native client. It is safe to access the swift client here. Meanwhile, the swift client would always available until the termination callback get called.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -259,43 +313,56 @@ public class Mqtt5Client {
...
+                        continuationCore.release()
+                        return continuation.resume(throwing: CommonRunTimeError.crtError(CRTError.makeFromLastError()))
+                    }
                 }
             }
         }
     }
 
+    /// Discard all operations and cleanup the client. It is MANDATORY function to call to release the client.
     public func close() {
```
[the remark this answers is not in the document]
waahm7: In other words, making it mandatory to call close just pushes the problem to customers to figure out when it is safe to deinit the client. We should try to solve this problem internally because it's a hard problem. I imagine something along the following lines.
MQTTClient has an external ref and an internal ref.

1. External ref goes to zero with ARC, the MQTT client is in the closing state and the customer cannot do anything new with it like publish, etc.
2. All the pending stuff will either complete successfully or with an error (if the client was stopped, etc.)
3. Internal ref goes to zero, and the client is cleaned up.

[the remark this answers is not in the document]
xiazhvera: The internal/external reference approach does not serve as a substitute for the close() function in this context. Instead, the close() function targets to resolve the issue where the external reference count never reaches zero.

If the client is offline or has any network issue, the operations would never completes and be left in the queue. As the operations use continuation to follow the async API design, consequentially, the ARC never goes to 0 in such case (external ref never goes to 0). The close() is used to release the external ref in such case. 
And it is not something we can decide internally, as the user might choose to wait forever on the server responds or they could timeout in certain case. They should be able to decide which way to go.

Another possible approach is that we can provide a Client Timeout concept which will release the client after a certain time. However, I think it will cause confusion between mqtt actual operation timeout (which failed the operation after certain time), and the binding client timeout (which release the client if the operation not completes). I don't feel like to introduce a new concept here.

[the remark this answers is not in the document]
waahm7: > The internal/external reference approach does not serve as a substitute for the close() function in this context. Instead, the close() function aims to resolve the issue where the external reference count never reaches zero.

@xiazhvera Yeah, I agree and am not saying that we should get rid of close. It should be an optional function, which means that it cancels anything pending, etc. It should not always be required. If the customer lets go of the last reference, it should still clean up everything instead of leaking memory. I can think of the following choices when the last reference is dropped:
1. Make it call the close function, i.e., cancel everything pending.
2. Wait for the operations to complete (maybe with a timeout value?) and clean up everything after the timer is up or all the pending operations are completed.

> If the client is offline or has any network issues, the operations would never complete and be left in the queue. As the operations use continuation to follow the async API design, consequently, the ARC never goes to 0 in such cases (external reference never goes to 0).

The problem is by introducing a mandatory close function, we are just pushing the problem to the customer. Should they keep waiting for the pending operations or timeout, etc.? Instead of us handling this problem for them, every customer will have to develop a way to keep track of the pending operations, implement a timeout function, and then close the client after the timeout.

> Another possible approach is that we can provide a Client Timeout concept, which will release the client after a certain time. However, I think it will cause confusion between the MQTT actual operation timeout (which fails the operation after a certain time) and the binding client timeout (which releases the client if the operation does not complete). I don't feel like introducing a new concept here.

Honestly, I don't mind this. If the MQTT actual operation has a default timeout, we don't even need to introduce a client timeout. The client will get cleaned up when the MQTT operation times out. If there is no default timeout, we can introduce one or just document that customers should set it if they don't want to wait forever.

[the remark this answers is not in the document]
waahm7: >  the ARC never goes to 0 in such cases (external reference never goes to 0).
The ARC should go to zero unless the customer is keeping an explicit reference to the client or capturing it in the callback, etc. (which is a tricky case). From our side, pending callbacks should not acquire a reference to the client or an external ref if we are using internal/external reference.

Test/AwsCommonRuntimeKitTests/mqtt/Mqtt5ClientTests.swift
```diff
@@ -1858,4 +1914,49 @@ class Mqtt5ClientTests: XCBaseTestCase {
...
+
+
+        let testContext = MqttTestContext()
+        let client = try createClient(clientOptions: clientOptions, testContext: testContext)
+        defer { client.close() }
+        try connectClient(client: client, testContext: testContext)
+        try stopClient(client: client, testContext: testContext)
+
+        let topic = "test/MQTT5_Binding_Swift_" + UUID().uuidString
+        let subscribePacket = PublishPacket(qos: QoS.atLeastOnce, topic: topic)
```
sbSteveK, line 1957: Trivial: publishPacket

xiazhvera:
Added back the callback core, now the `close` is no longer required.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -121,44 +121,154 @@ public typealias OnWebSocketHandshakeInterceptComplete = (HTTPRequestBase, Int32
...
+public typealias OnWebSocketHandshakeIntercept = (HTTPRequest, @escaping OnWebSocketHandshakeInterceptComplete) async -> Void
 
 // MARK: - Mqtt5 Client
 public class Mqtt5Client {
-    private var rawValue: UnsafeMutablePointer<aws_mqtt5_client>?
-    private var callbackCore: MqttCallbackCore
+    internal var clientCore: Mqtt5ClientCore
 
+    /// Creates a Mqtt5Client instance using the provided MqttClientOptions. It is MANDATORY to call `close()`
+    /// to clean up the Mqtt5Client resource
```
waahm7, line 131: old doc comment?

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -121,44 +121,154 @@ public typealias OnWebSocketHandshakeInterceptComplete = (HTTPRequestBase, Int32
...
+    ///
+    /// - Parameters:
+    ///     clientOptions: The MqttClientOptions class to use to configure the new Mqtt5Client.
+    ///
+    /// - Throws: CommonRuntimeError.crtError If the system is unable to allocate space for a native MQTT5 client structure
+    init(clientOptions options: MqttClientOptions) throws {
```
waahm7, line 137: Should this init be public? Also, I find it cleaner to just keep the same name as parameter unless needed.
```suggestion
    public init(clientOptions :MqttClientOptions) throws {
```

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -121,44 +121,154 @@ public typealias OnWebSocketHandshakeInterceptComplete = (HTTPRequestBase, Int32
...
 /// handshake since some work flows may be asynchronous. To accommodate that, onComplete must be invoked upon
 /// completion of the signing process.
-public typealias OnWebSocketHandshakeIntercept = (HTTPRequest, @escaping OnWebSocketHandshakeInterceptComplete) -> Void
+public typealias OnWebSocketHandshakeIntercept = (HTTPRequest, @escaping OnWebSocketHandshakeInterceptComplete) async -> Void
 
 // MARK: - Mqtt5 Client
 public class Mqtt5Client {
-    private var rawValue: UnsafeMutablePointer<aws_mqtt5_client>?
-    private var callbackCore: MqttCallbackCore
+    internal var clientCore: Mqtt5ClientCore
```
waahm7, line 128: Not sure, but should this be a weak ref? Otherwise the Mqtt5Clint will stay alive until core is alive.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -121,44 +121,154 @@ public typealias OnWebSocketHandshakeInterceptComplete = (HTTPRequestBase, Int32
...
+    // user callbacks
+    ///////////////////////////////////////
+    internal let onPublishReceivedCallback: OnPublishReceived
+    internal let onLifecycleEventStoppedCallback: OnLifecycleEventStopped
+    internal let onLifecycleEventAttemptingConnect: OnLifecycleEventAttemptingConnect
+    internal let onLifecycleEventConnectionSuccess: OnLifecycleEventConnectionSuccess
+    internal let onLifecycleEventConnectionFailure: OnLifecycleEventConnectionFailure
+    internal let onLifecycleEventDisconnection: OnLifecycleEventDisconnection
+    // The websocket interceptor could be nil if the websocket is not in use
+    internal let onWebsocketInterceptor: OnWebSocketHandshakeIntercept?
```
waahm7, line 228: Debatable: I feel like internal is redundant, I get the idea that it is more explicit. However, marking stuff internal explicitly does beg the question of why internal and not private?

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -121,44 +121,154 @@ public typealias OnWebSocketHandshakeInterceptComplete = (HTTPRequestBase, Int32
...
+
+    /// Creates a Mqtt5Client instance using the provided MqttClientOptions. It is MANDATORY to call `close()`
```
waahm7, line 230: outdated doc.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -121,44 +121,154 @@ public typealias OnWebSocketHandshakeInterceptComplete = (HTTPRequestBase, Int32
...
+    deinit {
+        clientCore.close()
+    }
+
+}
+
+// MARK: - Internal/Private
+
+/// Mqtt5 Client Core, internal class to handle Mqtt5 Client operations
+public class Mqtt5ClientCore {
```
waahm7, line 214: From what I understand the idea of Core class is to retain stuff that is needed while waiting for the callback from C. So that the actual class can go away, and when the callback happens, we don't crash with invalid references. Having a Core class which is just a thin wrapper over the actual class is not useful IMO and just makes the code harder to read.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -121,44 +121,154 @@ public typealias OnWebSocketHandshakeInterceptComplete = (HTTPRequestBase, Int32
...
 
+    /// Creates a Mqtt5Client instance using the provided MqttClientOptions. It is MANDATORY to call `close()`
+    /// to clean up the Mqtt5Client resource
+    ///
+    /// - Parameters:
+    ///     clientOptions: The MqttClientOptions class to use to configure the new Mqtt5Client.
+    ///
+    /// - Throws: CommonRuntimeError.crtError If the system is unable to allocate space for a native MQTT5 client structure
+    init(clientOptions options: MqttClientOptions) throws {
+        try options.validateConversionToNative()
```
waahm7, line 138: Maybe not part of this PR but this is redundant, the core also validates it. We should just update CStruct to update or add more functions which do the validation. I think we should just get rid of all the validateConversion* functions. It should be part of CStruct interface, so that we can have a consistent way of doing this.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -259,43 +397,55 @@ public class Mqtt5Client {
...
+                    guard result == AWS_OP_SUCCESS else {
+                        continuationCore.release()
+                        return continuation.resume(throwing: CommonRunTimeError.crtError(CRTError.makeFromLastError()))
+                    }
                 }
             }
         }
     }
 
+    /// Discard all operations and cleanup the client. It is MANDATORY function to call to release the client core.
```
waahm7, line 416: outdated doc?

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
waahm7: See my other comment on the CoreClass.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -121,44 +121,154 @@ public typealias OnWebSocketHandshakeInterceptComplete = (HTTPRequestBase, Int32
...
+    ///
+    /// - Throws: CommonRuntimeError.crtError
     public func start() throws {
-        if rawValue != nil {
+        try self.rwlock.read {
+            // Validate close() has not been called on client.
+            guard let rawValue = self.rawValue else {
+                // TODO add new error type for client closed
+                throw CommonRunTimeError.crtError(CRTError.makeFromLastError())
+            }
```
waahm7, line 271: We can pull this into a helper closure, since every function will need to do this.

waahm7 (commented):
I have reviewed the PR from a Swift perspective and added some comments, mainly I think the use of the core class is not right in this context. I think it would be helpful to add an FAQ or update the README to outline the different use cases, best practices, etc., like what happens if there are pending operations and the client is never closed, etc.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
xiazhvera: I set the member functions to internal so that we can directly access the member function in the callbacks function:
For example: `internal func MqttClientHandleLifecycleEvent(_ lifecycleEvent: UnsafePointer<aws_mqtt5_client_lifecycle_event>?) `
https://github.com/awslabs/aws-crt-swift/blob/8867080300bf91f2306070dd26b71c273990aa04/Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift#L447

waahm7: Should this be fileprivate then? I am fine with internal, it's just if you explicitly say it, it makes me wonder why internal. If you don't state it, it's more like I don't want it to be public, just give me default internal scope.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
xiazhvera: I make the native client and related members live in internal core instead of the actual class, so that we could avoid any accessing to the actual client class during callback or any native operations to avoid invalid references issue,

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
xiazhvera: The `rawValue` null validation would be removed in a later [PR](https://github.com/awslabs/aws-crt-swift/pull/264/files#diff-56edcc58b3d774bc93486d6f0f62fd6ddb0bac7bdcb18b268cdbd02bc6e54532L182). The final version would be just one line code:
```
 try self.rwlock.read {

}
```
I will keep it as it for now.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
xiazhvera: The change might introduce changes in CStruct interface. I will think about it and update in a later PR.

PR #268, opened 2024-06-11
Update Errors in Mqtt Client 

xiazhvera:
*Issue #, if available:*

*Description of changes:*
1. Added error space for aws-crt-swift
2. Added a way to allow customized error message for Swift
3. Added `unwrap` function to unwrap Optional


Source/AwsCommonRuntimeKit/crt/CommonRuntimeError.swift
```diff
@@ -23,6 +23,20 @@ public struct CRTError: Equatable {
         self.name = String(cString: aws_error_name(self.code))
     }
 
+    public init<T: BinaryInteger>(code: T, context: String?) {
+        if code > INT32_MAX || code <= 0 {
+            self.code = Int32(AWS_ERROR_UNKNOWN.rawValue)
```
sfod, line 28: Is it intentional that `code == 0` is treated as unknown error? It doesn't feel right, because `AWS_ERROR_SUCCESS` is valid code and can be passed to this function, even if by accident (e.g. by `aws_last_error ()` call).

Source/AwsCommonRuntimeKit/crt/CommonRuntimeError.swift
```diff
@@ -23,6 +23,20 @@ public struct CRTError: Equatable {
         self.name = String(cString: aws_error_name(self.code))
     }
 
+    public init<T: BinaryInteger>(code: T, context: String?) {
```
sbSteveK, line 26: Do we still need the context string if we are generating errors from crt swift?

In the areas that are currently using context, it seems like error logging would be more appropriate. Unless we are not going to generate logs from aws-crt-swift.

sbSteveK (approved):
Unsure whether we need the context string in CommonRuntimeError. That can be changed later if we decide against it. Looks solid otherwise.

On a side note, once we do this, we need to update common to reflect the error range that aws-crt-swift is taking.

Source/AwsCommonRuntimeKit/crt/CommonRuntimeError.swift
waahm7, line 26: We should just update the existing function. It won't break anything.

Source/AwsCommonRuntimeKit/crt/Utilities.swift
```diff
@@ -147,7 +148,7 @@ extension TimeInterval {
         let _millisecond = (self * 1_000).rounded()
         guard _millisecond >= 0 && _millisecond <= Double(UInt32.max) else {
             // todo convert the millisecond conversion errors into aws-crt-swift errors
-            throw CommonRunTimeError.crtError(CRTError(code: AWS_ERROR_INVALID_ARGUMENT.rawValue))
+            throw CommonRunTimeError.crtError(CRTError(code: AWS_CRT_SWIFT_INVALID_ARGUMENT.rawValue))
```
waahm7, line 151: We should avoid duplicating errors because they are "common". The distinction between AWS_ERROR_INVALID_ARGUMENT and AWS_CRT_SWIFT_INVALID_ARGUMENT might not be clear to customers, leading to confusion. Having two different errors for the same issue can be unnecessarily complicated.

Source/AwsCommonRuntimeKit/crt/Utilities.swift
```diff
@@ -223,36 +224,14 @@ extension aws_array_list {
...
-func convertOptionalBool(_ pointer: UnsafePointer<Bool>?) -> Bool? {
-    guard let validPointer = pointer else {
-        return nil
+extension Optional {
+    /// unwrap an optional unsafepointer to its underlying type
+    func unwrap<T>() -> T? where Wrapped == (UnsafePointer<T>) {
+        guard let validPointer = self else {
+            return nil
+        }
+        return validPointer.pointee
```
waahm7, line 233: This feels like something Swift standard API should already provide. How is this different from `.?`?

Source/LibNative/module.modulemap
```diff
@@ -0,0 +1,6 @@
+
+module LibNative {
+    header "CommonRuntimeError.h"
+    export *
+}
+
```
waahm7, line 6: Why not put it in https://github.com/awslabs/aws-crt-swift/tree/main/aws-common-runtime/config?

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -286,7 +286,7 @@ public class Mqtt5ClientCore {
         try self.rwlock.read {
             // Validate close() has not been called on client.
             guard let rawValue = self.rawValue else {
-                throw CommonRunTimeError.crtError(CRTError.makeFromLastError())
+                throw CommonRunTimeError.crtError(CRTError(code: AWS_CRT_SWIFT_MQTT_CLIENT_CLOSED.rawValue))
```
waahm7, line 289: Debatable: Probably better to just use `AWS_ERROR_INVALID_STATE` with context?

waahm7 (commented):
Added comments. I am willing to agree to disagree here. I still feel that this PR doesn't have a strong enough use case for Swift errors. The added complexity might not justify the benefits, which seem limited to a slightly better error name that might not be significant for users. Our error handling and Swift error handling is terrible, and I understand the desire to make it better for the customers. However, this just feels like a bandage which doesn't solve the issue. I would suggest holding off on merging this until we identify a more concrete use case, but I'll defer to your judgment.

Source/AwsCommonRuntimeKit/crt/CommonRuntimeError.swift
```diff
@@ -14,7 +14,7 @@ public struct CRTError: Equatable {
     public let name: String
 
     public init<T: BinaryInteger>(code: T) {
-        if code > INT32_MAX || code <= 0 {
+        if code > INT32_MAX || code < 0 {
```
waahm7, line 17: I think this was intentional. This can lead a bad call to last error becoming AWS_ERROR_SUCCESS, which will probably make the program fail at some later stage, making it harder to debug. I think it is easier to debug where the unknown error is coming from and fix that.

Source/AwsCommonRuntimeKit/crt/CommonRuntimeError.swift
waahm7: Yes, that was intentional. This can lead a bad call to last error becoming AWS_ERROR_SUCCESS, which will probably make the program fail at some later stage, making it harder to debug. I think it is easier to debug where the unknown error is coming from and fix that.

Source/AwsCommonRuntimeKit/crt/CommonRuntimeError.swift
waahm7: Or customers will get exceptions with AWS_ERROR_SUCCESS instead of ERROR_UNKNOWN

Source/LibNative/module.modulemap
xiazhvera: The `config` module is required to build before `aws-c-common` because it is used for build configuration. Meanwhile, the `error` file uses error struct from `aws-c-common`, which introduces a dependency on `aws-c-common`. If we add the error file into config module, it would be a recursive dependency which is not allowed by the swift package.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
xiazhvera: I decided to keep the Swift error space, as it makes sense to maintain it whether we use it or not.  Consequently, I am also keeping the `AWS_CRT_SWIFT_MQTT_CLIENT_CLOSED` error code. 
Additionally, I removed `AWS_CRT_SWIFT_INVALID_ARGUMENT ` to avoid confusion with `AWS_ERROR_INVALID_ARGUMENT `

Source/AwsCommonRuntimeKit/crt/CommonRuntimeError.swift
xiazhvera: I eventually decide to keep the context, as the logging for Swift is not there yet.

PR #276, opened 2024-08-02
Add string array as an option for request context params for endpoint resolution

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Source/AwsCommonRuntimeKit/crt/Utilities.swift
+++ b/Source/AwsCommonRuntimeKit/crt/Utilities.swift
@@ -273,36 +273,11 @@ func withByteCursorFromStrings<Result>(
     }
 }
 
-public func scan<S: Sequence, U>(_ seq: S, _ initial: U, _ combine: (U, S.Element) -> U) -> [U] {
-    var result: [U] = []
-    result.reserveCapacity(seq.underestimatedCount)
-    var runningResult = initial
-    for element in seq {
-    runningResult = combine(runningResult, element)
-    result.append(runningResult)
-    }
-    return result
-}
-
-/// Note: this function and associated scan function above are copied verbatim from
-/// swift standard lib where its private. 
-public func withArrayOfCStrings<R>(
-    _ args: [String], _ body: ([UnsafeMutablePointer<CChar>?]) -> R
-) -> R {
-    let argsCounts = Array(args.map { $0.utf8.count + 1 })
-    let argsOffsets = [ 0 ] + scan(argsCounts, 0, +)
-    let argsBufferSize = argsOffsets.last!
-    var argsBuffer: [UInt8] = []
-    argsBuffer.reserveCapacity(argsBufferSize)
-    for arg in args {
-        argsBuffer.append(contentsOf: arg.utf8)
-        argsBuffer.append(0)
-    }
-    return argsBuffer.withUnsafeMutableBufferPointer { (argsBuffer) in
-        let ptr = UnsafeMutableRawPointer(argsBuffer.baseAddress!).bindMemory(
-        to: CChar.self, capacity: argsBuffer.count)
-        var cStrings: [UnsafeMutablePointer<CChar>?] = argsOffsets.map { ptr + $0 }
-        cStrings[cStrings.count - 1] = nil
-        return body(cStrings)
+func withByteCursorArrayFromStringArray<R>(
+    _ arg: [String], _ body: (UnsafePointer<aws_byte_cursor>, Int) -> R) -> R {
+    let cursors = arg.map { aws_byte_cursor_from_c_str($0) }
+    let len = cursors.count
+    return cursors.withUnsafeBufferPointer { cursorsPtr in 
+        return body(cursorsPtr.baseAddress!, len)
     }
 }
--- a/Source/AwsCommonRuntimeKit/crt/Utilities.swift
+++ b/Source/AwsCommonRuntimeKit/crt/Utilities.swift
@@ -275,8 +275,14 @@ func withByteCursorFromStrings<Result>(
 
 func withByteCursorArrayFromStringArray<R>(
     _ arg: [String], _ body: (UnsafePointer<aws_byte_cursor>, Int) -> R) -> R {
+    let cStrings = arg.map { strdup($0) } + [nil]
     let cursors = arg.map { aws_byte_cursor_from_c_str($0) }
     let len = cursors.count
+
+    defer {
+        cStrings.forEach { free($0) }
+    }
+
     return cursors.withUnsafeBufferPointer { cursorsPtr in 
         return body(cursorsPtr.baseAddress!, len)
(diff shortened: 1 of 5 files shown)


Source/AwsCommonRuntimeKit/crt/Utilities.swift
```diff
@@ -272,3 +272,37 @@ func withByteCursorFromStrings<Result>(
...
 }
+
+public func scan<S: Sequence, U>(_ seq: S, _ initial: U, _ combine: (U, S.Element) -> U) -> [U] {
+    var result: [U] = []
+    result.reserveCapacity(seq.underestimatedCount)
+    var runningResult = initial
+    for element in seq {
+    runningResult = combine(runningResult, element)
+    result.append(runningResult)
+    }
```
waahm7, line 283: nit: lint

Source/AwsCommonRuntimeKit/crt/Utilities.swift
```diff
@@ -272,3 +272,37 @@ func withByteCursorFromStrings<Result>(
...
+    return result
+}
+
+/// Note: this function and associated scan function above are copied verbatim from
+/// swift standard lib where its private. 
+public func withArrayOfCStrings<R>(
```
waahm7, line 289: These functions should be internal to our module and not public. Also, if performance is not a concern, we should just implement the simpler version from https://oleb.net/blog/2016/10/swift-array-of-c-strings/.

Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointsRequestContext.swift
```diff
@@ -48,6 +48,34 @@ public class EndpointsRequestContext {
         }
     }
 
+    private func withByteCursorArrayFromStringArray<R>(
```
waahm7, line 51: Trivial: We should move this to Utilities.swift, as this will probably get reused in the future. We should also add a test case for just this function.

Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointsRequestContext.swift
```diff
@@ -48,6 +48,34 @@ public class EndpointsRequestContext {
...
+        _ arg: [String], _ body: (UnsafeMutablePointer<aws_byte_cursor>, Int) -> R) -> R {
+        withArrayOfCStrings(arg) { cStrArr in
+            var cursors = cStrArr.map { aws_byte_cursor_from_c_str($0) }
+            let len = cursors.count
+            return cursors.withUnsafeMutableBufferPointer { cursorsPtr in 
+                return body(cursorsPtr.baseAddress!, len)
+            }
```
waahm7, line 58: Do we need a mutable pointer to these, or should we update our C API to take a const ptr?

Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointsRequestContext.swift
```diff
@@ -48,6 +48,34 @@ public class EndpointsRequestContext {
...
+    }
+
+    /// Add a string array endpoint parameter to the request context
+    /// - Parameters:
+    ///   - name: The name of the parameter
+    ///   - value: The value of the parameter
+    public func add(name: String, value: [String]?) throws {
+        guard let value = value else {
+            return
+        }
```
waahm7, line 69: We should guard against empty array or have a test to make sure that it works as expected.

DmitriyMusatkin: let me add a test. empty array is a valid input

Source/AwsCommonRuntimeKit/crt/Utilities.swift
DmitriyMusatkin: ugh, i didnt really understand how swift works. updated to a much simpler version.

Source/AwsCommonRuntimeKit/crt/Utilities.swift
```diff
@@ -272,3 +272,12 @@ func withByteCursorFromStrings<Result>(
         }
     }
 }
+
+func withByteCursorArrayFromStringArray<R>(
+    _ arg: [String], _ body: (UnsafePointer<aws_byte_cursor>, Int) -> R) -> R {
+    let cursors = arg.map { aws_byte_cursor_from_c_str($0) }
```
waahm7, line 278: This is not memory-safe. $0 is a Swift string and is only valid for the scope of `aws_byte_cursor_from_c_str`. The behavior is undefined after this line because the cursor refers to a memory area not guaranteed by the Swift compiler. We will need to make a copy, and then release it using our allocator.

DmitriyMusatkin: yeah, you are right. i added the copy back. 
is it possible to run asan or some sort of mem checker in swift?

Source/AwsCommonRuntimeKit/crt/Utilities.swift
```diff
@@ -272,3 +272,18 @@ func withByteCursorFromStrings<Result>(
         }
     }
 }
+
+func withByteCursorArrayFromStringArray<R>(
+    _ arg: [String], _ body: (UnsafePointer<aws_byte_cursor>, Int) -> R) -> R {
+    let cStrings = arg.map { strdup($0) } + [nil]
+    let cursors = arg.map { aws_byte_cursor_from_c_str($0) }
```
waahm7, line 279:
```suggestion
    let cursors = cStrings.map { aws_byte_cursor_from_c_str($0) }
```

Source/AwsCommonRuntimeKit/crt/Utilities.swift
```diff
@@ -272,3 +272,18 @@ func withByteCursorFromStrings<Result>(
         }
     }
 }
+
+func withByteCursorArrayFromStringArray<R>(
+    _ arg: [String], _ body: (UnsafePointer<aws_byte_cursor>, Int) -> R) -> R {
```
waahm7, line 277: Trivial: Would be cleaner to do 
```suggestion
extension Array where Element == String {
  func withByteCursorArray<R>(
      _ arg: [String], _ body: (UnsafePointer<aws_byte_cursor>, Int) -> R) -> R {
``` 
and then you can do string.withByteCursorArray ...

waahm7 (approved):

Source/AwsCommonRuntimeKit/crt/Utilities.swift
waahm7: We do have our tracing allocator hooked up, but I never looked into ASAN. It seems easy enough to enable: https://www.swift.org/documentation/server/guides/llvm-sanitizers.html.

PR #288, opened 2024-10-01
Add TLS support

sbSteveK:
Bindings for TLS support


Test/AwsCommonRuntimeKitTests/mqtt/Mqtt5ClientTests.swift
```diff
@@ -4,6 +4,7 @@
 import XCTest
 import Foundation
 import AwsCMqtt
+import LibNative
```
xiazhvera, line 7: I think this import was from the old iOS app PR, and could be removed.

xiazhvera (approved):
The binding looks good to me. Just I'm a little confused about how the label works.  Does the label a requirement for set up the tls options? Or does the makeTLS function only "import" the certificates but does not set it ...?

sbSteveK:
> The binding looks good to me. Just I'm a little confused about how the label works. Does the label a requirement for set up the tls options? Or does the makeTLS function only "import" the certificates but does not set it ...?

The label is optional if the user wants to apply their own human-readable label to the certificate and private key being imported into the keychain. Default values are set in aws-c-io.

Issue #289, opened 2024-10-02
(short issue description)


jeffdav:
### Describe the bug

Building on Windows using `swift build` fails with errors.

### Expected Behavior

`swift build` should build the project on Windows.

### Current Behavior

```swift
PS C:\Users\jeffd\src\aws-crt-swift> git status
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
PS C:\Users\jeffd\src\aws-crt-swift> git pull
Already up to date.
PS C:\Users\jeffd\src\aws-crt-swift> swift build
warning: 'aws-crt-swift': Invalid Exclude 'C:\Users\jeffd\src\aws-crt-swift\aws-common-runtime\aws-checksums\source\intel\visualc': File not found.
Building for debugging...
...snip warnings about format specifiers...
C:\Users\jeffd\src\aws-crt-swift\aws-common-runtime\aws-c-io\source\windows\iocp\socket.c:293:27: error: field has incomplete type 'struct aws_overlapped'
  293 |     struct aws_overlapped signal;
      |                           ^
C:\Users\jeffd\src\aws-crt-swift\aws-common-runtime\aws-c-io\source\windows\iocp\socket.c:293:12: note: forward declaration of 'struct aws_overlapped'
  293 |     struct aws_overlapped signal;
      |            ^
C:\Users\jeffd\src\aws-crt-swift\aws-common-runtime\aws-c-io\source\windows\iocp\socket.c:844:57: error: incomplete definition of type 'struct aws_overlapped'
  844 |     struct socket_connect_args *socket_args = overlapped->user_data;
      |                                               ~~~~~~~~~~^
C:\Users\jeffd\src\aws-crt-swift\aws-common-runtime\aws-c-io\source\windows\iocp\socket.c:293:12: note: forward declaration of 'struct aws_overlapped'
  293 |     struct aws_overlapped signal;
      |            ^
C:\Users\jeffd\src\aws-crt-swift\aws-common-runtime\aws-c-io\source\windows\iocp\socket.c:974:5: error: call to undeclared function 'aws_overlapped_init'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  974 |     aws_overlapped_init(&socket_impl->read_io_data->signal, s_socket_connection_completion, connect_args);
      |     ^
[...]

PR #290, opened 2024-10-03
[Windows] Build and test on Windows.

jeffdav:
*Issue #, if available:* #289

*Description of changes:*
* Add posix and bsd paths to excludes for Windows.
* Define `AWS_OS_WINDOWS_DESKTOP` and `AWS_USE_IO_COMPLETION_PORTS` for Windows.
* Fix some type casts between `Int32` <--> `UInt32` that cause errors on Windows.
* Remove non-existent exclude path (silences a useless warning).
* Explicitly use `Foundation.UUID` where needed to avoid confusion with Windows SDK `UUID`.
* Add `setenv`/`unsetenv` shim for tests on Windows.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -198,6 +198,14 @@ let awsCEventStreamExcludes = [
     "CODE_OF_CONDUCT.md",
     "clang-tidy/run-clang-tidy.sh"] + excludesFromAll
 
+//////////////////////////////////////////////////////////////////////
+/// AwsCommonRuntimeKitTests
+//////////////////////////////////////////////////////////////////////
+var awsCommonRuntimeKitTestsExcludes: [String] = []
+#if !os(Windows)
+awsCommonRuntimeKitTestsExcludes.append("shims/windows")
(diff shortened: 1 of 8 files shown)


jeffdav:
Still has two build issues:
1. When building `AwsCommonRuntimeKit`, the `aws-c-cal` header `opensslcrypto_common.h` is picked up by the clang importer, even though it's excluded in `Package.swift`.  I fixed this locally by putting `#if _MSC_VER` around the whole header.  I think the proper fix is to provide a `module.modulemap` file in `aws-c-cal`.
2. When building `AwsCommonRuntimeKit`, the `aws-c-common` module fails to find `GetConsoleMode()`, despite including `windows.h`, which the documentation says should be enough.  I fixed it locally by including `consoleapi.h` as well.

Since these issues are in submodules, they may need to be fixed first, but I'd love feedback.

jeffdav:
I'm going to close this for now, since it's just a POC.

PR #293, opened 2024-10-11
Pick up `s2n_cleanup_thread`

TingDaoK:
*Issue #, if available:*
```
aws-c-auth         v0.7.30 -> v0.7.31
aws-c-cal          v0.7.4 ✓
aws-c-common       v0.9.28 ✓
aws-c-compression  v0.2.19 ✓
aws-c-event-stream v0.4.3 ✓
aws-c-http         v0.8.9 -> v0.8.10
aws-c-io           v0.14.18 -> v0.14.19
aws-c-sdkutils     v0.1.19 ✓
aws-checksums      v0.1.19 -> v0.1.20
s2n                v1.5.2 -> v1.5.5
```

With a symlink to point s2n/unstable/cleanup.h to api/unstable/cleanup.h, since swift is not using CMake to build, so that it cannot use any CMake trick to check if s2n was build before or to be built together. And we are building s2n with other deps together.


waahm7 (approved):
We should add a `aws-common-runtime/config/Readme.md` to explain this hack.

Issue #297, opened 2024-11-07
bindings for RPCv2 CBOR support


dayaffe:
### Describe the feature

Please provide all available functionality via Swift bindings.

### Use Case

AWS SDK for Swift would like to use CRT to provide support for RPCv2 CBOR protocol. 

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #300, opened 2024-11-21
CBOR Bindings

waahm7:
*Issue #, if available:*
https://github.com/awslabs/aws-crt-swift/issues/297

*Description of changes:*
- Adds bindings for CBOR v2.


Source/AwsCommonRuntimeKit/crt/CBOR.swift
```diff
@@ -0,0 +1,363 @@
...
+        case AWS_CBOR_TYPE_NEGINT:
+            do {
+                var out_value: UInt64 = 0
+                guard
+                    aws_cbor_decoder_pop_next_negative_int_val(self.rawValue, &out_value)
+                        == AWS_OP_SUCCESS
+                else {
+                    throw CommonRunTimeError.crtError(.makeFromLastError())
+                }
+                return .int(-(Int64(out_value + 1)))
```
waahm7, line 175: Please verify the math here.

Source/AwsCommonRuntimeKit/crt/CBOR.swift
```diff
@@ -0,0 +1,360 @@
...
+        case .null: aws_cbor_encoder_write_null(self.rawValue)
+        case .text(let value):
+            do {
+                value.withByteCursor { cursor in
+                    aws_cbor_encoder_write_text(self.rawValue, cursor)
+                }
+            }
+        case .date(let value):
+            do {
+                aws_cbor_encoder_write_tag(self.rawValue, UInt64(AWS_CBOR_TAG_EPOCH_TIME))
```
waahm7, line 103: Please validate.

Source/AwsCommonRuntimeKit/crt/CBOR.swift
```diff
@@ -0,0 +1,360 @@
...
+                    aws_cbor_decoder_consume_next_single_element(self.rawValue)
+                        == AWS_OP_SUCCESS
+                else {
+                    throw CommonRunTimeError.crtError(.makeFromLastError())
+                }
+                return .null
+            }
+        case AWS_CBOR_TYPE_TAG:
+            var out_value: UInt64 = 0
+            guard
```
waahm7, line 233: Please validate. I am asserting here that the tag will always be 1, and the next value can only be an int, double, or uint64, which is then converted to a time.

Source/AwsCommonRuntimeKit/crt/CBOR.swift
```diff
@@ -0,0 +1,360 @@
...
+    /// Double type. It might be encoded as an integer if possible without loss of precision. Half-precision floats are not supported.
+    case double(_ value: Double)
+    /// Bytes type for binary data
+    case bytes(_ value: Data)
+    /// Text type for utf-8 encoded strings
+    case text(_ value: String)
+    /// Array type
+    case array(_ value: [CBORType])
+    /// Map type
+    case map(_ value: [String: CBORType])
```
waahm7, line 21: Please validate, asserting that map keys will always be string. For indefinite map, it's up to the users.

Source/AwsCommonRuntimeKit/crt/CBOR.swift
dayaffe: I think this looks mathematically correct, but do we need to worry about integer overflow if out_value is the largest Int64 and then you try to add 1 to it?

waahm7: Thanks, that's a great point. I am looking into it and will add tests for this.

Source/AwsCommonRuntimeKit/crt/CBOR.swift
```diff
@@ -0,0 +1,360 @@
...
+import Foundation
+
+/// CBOR Types. These types don't map one-to-one to the CBOR RFC.
+public enum CBORType: Equatable {
+    /// UINT64 type for positive numbers.
+    case uint64(_ value: UInt64)
+    /// INT64 type for negative numbers. If the number is positive, it will be encoded as UINT64 type.
+    case int(_ value: Int64)
+    /// Double type. It might be encoded as an integer if possible without loss of precision. Half-precision floats are not supported.
+    case double(_ value: Double)
```
TingDaoK, line 13: worth to mention that the numbers will be encoded as "smallest possible".

TingDaoK: I don't get why we have `uint64` and `int`

Either we just do `int64` for both positive or negative, which means we only supports [-2^63; 2^63-1] range. (If SDKs are okay with that.)

Or we be clear about `unsgined 64` and `negative 64`, which both take unsigned int 64 to meet the range cbor supports, you can see [here](https://www.rfc-editor.org/rfc/rfc8949.html#name-major-types), cbor supports range from  [-2^64; 2^64-1] .

Or we do both, be clear about unsigned and negative, and then a helper to just take int. So that we can support the range cbor required, while provides an easier to use API.

Source/AwsCommonRuntimeKit/crt/CBOR.swift
```diff
@@ -0,0 +1,360 @@
...
+    /// the `indef_*_start` type first, followed by N elements, and the break type at the end.
+    case indef_break
+    /// Indefinite Array Type
+    case indef_array_start
+    /// Indefinite Map Type
+    case indef_map_start
+    /// Indefinite Bytes Type
+    case indef_bytes_start
+    /// Indefinite Text Type
+    case indef_text_start
```
TingDaoK, line 41: trivial:
use the order that matches C/above
```suggestion
    /// Indefinite Bytes Type
    case indef_bytes_start
    /// Indefinite Text Type
    case indef_text_start
    /// Indefinite Array Type
    case indef_array_start
    /// Indefinite Map Type
    case indef_map_start
```

Source/AwsCommonRuntimeKit/crt/CBOR.swift
```diff
@@ -0,0 +1,360 @@
...
+        case AWS_CBOR_TYPE_NEGINT:
+            do {
+                var out_value: UInt64 = 0
+                guard
+                    aws_cbor_decoder_pop_next_negative_int_val(self.rawValue, &out_value)
+                        == AWS_OP_SUCCESS
+                else {
+                    throw CommonRunTimeError.crtError(.makeFromLastError())
+                }
+                return .int(Int64(-Int64(out_value) - 1))
```
TingDaoK, line 175: The cbor negative int supports ` [-2^64; -1]`, it may overflow here. Check for overflow

Source/AwsCommonRuntimeKit/crt/CBOR.swift
```diff
@@ -0,0 +1,360 @@
...
+                throw CommonRunTimeError.crtError(CRTError(code: AWS_ERROR_CBOR_UNEXPECTED_TYPE.rawValue))
+            }
+            let timestamp = try popNext()
+
+            if case .double(let value) = timestamp {
+                return .date(Date.init(timeIntervalSince1970: value))
+            } else if case .uint64(let value) = timestamp {
+                return .date(Date.init(timeIntervalSince1970: Double(value)))
+            } else if case .int(let value) = timestamp {
+                return .date(Date.init(timeIntervalSince1970: Double(value)))
```
TingDaoK, line 251: in C, converting from unsigned 64 or int 64 to double may loss the precision. Not sure about swift double, maybe worth to check. But seems like the `timeIntervalSince1970` in swift only takes double, so, maybe just document it out that's a risk of loss of precision?

Source/AwsCommonRuntimeKit/crt/CBOR.swift
```diff
@@ -0,0 +1,360 @@
...
+    /// Bytes type for binary data
+    case bytes(_ value: Data)
+    /// Text type for utf-8 encoded strings
+    case text(_ value: String)
+    /// Array type
+    case array(_ value: [CBORType])
+    /// Map type
+    case map(_ value: [String: CBORType])
+    /// Date type. It will be encoded as epoch-based date/time.
+    case date(_ value: Date)
```
TingDaoK, line 23: I'd add another case like `undefined_tags` (naming to be decided). So that we still have a way to support tags other than epoch-based time instead of just error out.
Unless SDK really don't want it and want to error out for tags other than epoch time.

Source/AwsCommonRuntimeKit/crt/CBOR.swift
waahm7: Thanks, I didn't realize `-2^64` is a possibility. I think we should just keep it simple and expose Int64 and UInt64 since they are Swift types. Maybe we can use better names like `uint` and `int`?
 
> Either we just do int64 for both positive or negative, which means we only supports [-2^63; 2^63-1] range. (If SDKs are okay with that.)

UInt64.max is a valid value, so we should support encoding/decoding that. 

> Or we be clear about unsgined 64 and negative 64, which both take unsigned int 64 to meet the range cbor supports, you can see [here](https://www.rfc-editor.org/rfc/rfc8949.html#name-major-types), cbor supports range from [-2^64; 2^64-1] .

The problem with `-2^64` is that there is no way to represent that in Swift in a type-safe way. During encoding, I don't think there is any way for Swift to have a `-2^64` integer unless it does what CBOR is doing which is `UInt64 named NegInt`. 

During decoding, yeah, we might encounter `-2^64` since CBOR allows it, but we should just error out instead of passing that complexity to the users to deal with. This should never happen in practice because I don't think anyone would be using `-2^64` numbers. We can add the `UInt64 named NegInt` type in the future if needed. 

@dayaffe What do you think?

TingDaoK: Actually I thought of that again.

It seems to be a one-way door as it's an enum, that we cannot add the `UInt64 named NegInt` type in the future, since how will you deal with the value that's possible to fit into two different types??

TingDaoK: Like currently, looks like you have both int64 and uint64 for the positive number, but during decode, you only give the negative value to the int64, and all the positive value will be uin64. 

It's arguably working as expected? But, seems more reasonable to just have two different type for negative and positive?

Maybe something like
```
public func fromInt64(int) throws -> CBORType {
}
``` 
???

waahm7: Discussed offline, it's not a one-way door since we can add a negint type in the future. I have added a warning that this enum is non_exhaustive. It is not ideal because we can encode it as different types, but we already have this issue with bytes, text, and double types, etc., so it's not too bad.

> as a helper to take the int and convert it to the type???

Yeah, we can do that, but I am not too concerned about the encoding part. The problem is during decoding; we can't map -2^64 to a Swift type, and I think we should try to hide this complexity from our users until needed. The API I am trying to design is to just encode Swift types and decode to Swift types to keep it simple and expected.

Source/AwsCommonRuntimeKit/crt/CBOR.swift
waahm7: Thanks, I have added a comment.

Source/AwsCommonRuntimeKit/crt/CBOR.swift
waahm7: Thanks, I have added a `tag` case.

TingDaoK (approved):

PR #305, opened 2024-12-20
feat: Refactor CBOR bindings and add support for indefinite object decoding

dayaffe:
*Issue #, if available:*

*Description of changes:*
- Added support for indefinite object decoding
- Refactored during debugging

### DIFF
--- a/Source/AwsCommonRuntimeKit/crt/CBOR.swift
+++ b/Source/AwsCommonRuntimeKit/crt/CBOR.swift
@@ -61,36 +61,46 @@ public class CBOREncoder {
         self.rawValue = rawValue
     }
 
-    /// Encode a single type
-    /// - Parameters:
-    ///   - value: value to encode 
-    /// - Throws: CommonRuntimeError.crtError
     public func encode(_ value: CBORType) {
         switch value {
-        case .uint(let value): aws_cbor_encoder_write_uint(self.rawValue, value)
-        case .int(let value):
-            do {
-                if value >= 0 {
-                    aws_cbor_encoder_write_uint(self.rawValue, UInt64(value))
-                } else {
-                    aws_cbor_encoder_write_negint(self.rawValue, UInt64(-1 - value))
-                }
+        case .uint(let v):
+            aws_cbor_encoder_write_uint(self.rawValue, v)
+        case .int(let v):
+            if v >= 0 {
+                aws_cbor_encoder_write_uint(self.rawValue, UInt64(v))
+            } else {
+                aws_cbor_encoder_write_negint(self.rawValue, UInt64(-1 - v))
             }
-        case .double(let value): aws_cbor_encoder_write_float(self.rawValue, value)
-        case .array(let values):
-            do {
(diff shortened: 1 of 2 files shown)


Source/AwsCommonRuntimeKit/crt/CBOR.swift
```diff
@@ -63,34 +63,47 @@ public class CBOREncoder {
...
-        case .uint(let value): aws_cbor_encoder_write_uint(self.rawValue, value)
-        case .int(let value):
-            do {
-                if value >= 0 {
-                    aws_cbor_encoder_write_uint(self.rawValue, UInt64(value))
-                } else {
-                    aws_cbor_encoder_write_negint(self.rawValue, UInt64(-1 - value))
-                }
+        case .uint(let v):
+            aws_cbor_encoder_write_uint(self.rawValue, v)
```
waahm7, line 71: trivial: Personally, I don't like single-character variable names because they do not convey much meaning. For example, `d` could represent data or a date. Can we rename this to a more descriptive name instead?

Source/AwsCommonRuntimeKit/crt/CBOR.swift
```diff
@@ -151,218 +152,292 @@ public class CBORDecoder {
...
+    private func decodeTag() throws -> CBORType {
+        var out_value: UInt64 = 0
+        guard aws_cbor_decoder_pop_next_tag_val(self.rawValue, &out_value) == AWS_OP_SUCCESS else {
+            throw CommonRunTimeError.crtError(.makeFromLastError())
+        }
+
+        // If it's not the epoch-based time tag (tag 1), return as a generic tag.
+        if out_value != 1 {
+            return .tag(out_value)
+        }
```
waahm7, line 284: trivial: Existing code but just realized that guard might be better here.

Source/AwsCommonRuntimeKit/crt/CBOR.swift
```diff
@@ -151,218 +152,292 @@ public class CBORDecoder {
...
+                break
             }
-            return .array(array)
-        case AWS_CBOR_TYPE_MAP_START:
-            var out_value: UInt64 = 0
-            guard
-                aws_cbor_decoder_pop_next_map_start(self.rawValue, &out_value)
-                    == AWS_OP_SUCCESS
-            else {
+            array.append(try popNext())
```
waahm7, line 357: You can just do something like 
let cborType = try popNext(); 
if cborType == BREAK 
      break 
else 
      array.append(cborType)
      
      
 Same for other indef types.

waahm7 (commented):
Thanks for creating the PR. Just some trivial comments.

Source/AwsCommonRuntimeKit/crt/CBOR.swift
```diff
@@ -132,6 +132,7 @@ public class CBOREncoder {
     }
 }
 
+// swiftlint:disable type_body_length
```
waahm7, line 135: trivial: This is not needed.
```suggestion
```

waahm7 (approved):

PR #306, opened 2024-12-23
Cognito credential provider support

xiazhvera:
*Issue #, if available:*

*Description of changes:*
The PR added support for cognito credential provider. 
It also depends on latest crt-builder to enable the environment variables for xcodebuild test. https://github.com/awslabs/aws-crt-builder/pull/314


.builder/actions/crt-ci-prep-xcodebuild.py
```diff
@@ -0,0 +1,9 @@
+import Builder
+
+class CrtCiPrepXCodebuild(Builder.Action):
```
waahm7, line 3: I haven't looked at the PR but can we avoid adding this action in crt-swift? We don't need to test the Cognito credential provider functionality in bindings; we can just create/destroy it for simple binding tests. The MQTT-specific tests should be in the IoT SDK.

xiazhvera: This script change is used to allow passing environment variables into xcodebuild tests. Regardless of the Cognito credential provider changes, we still need it for the unit tests that requires environment variables.

Source/AwsCommonRuntimeKit/auth/credentials/CredentialsProvider.swift
```diff
@@ -76,7 +138,7 @@ extension CredentialsProvider {
 }
 
 extension CredentialsProvider.Source {
-
+
```
sbSteveK, line 141: There are a large number of weird white space additions onto empty lines. Is this an auto formatting issue?

sbSteveK (approved):

.builder/actions/crt-ci-prep-xcodebuild.py
waahm7, line 3: Can we merge this to Main instead of iot branch? Swift SDK might need this as well. (I haven't reviewed the PR)

.github/workflows/ci.yml
```diff
@@ -6,8 +6,8 @@ on:
       - 'main'
 
 env:
-  BUILDER_VERSION: v0.9.73
-  BUILDER_SOURCE: releases
+  BUILDER_VERSION: xcodebuild_setup
```
xiazhvera, line 9: This would need to be updated once crt-builder is merged. Currently the crt-builder has CI failure related to gcc-8.

Source/AwsCommonRuntimeKit/auth/credentials/CredentialsProvider.swift
```diff
@@ -10,6 +10,68 @@ public protocol CredentialsProviding {
...
+        return withByteCursorFromStrings(IdentityProviderName,
+                                         IdentityProviderToken) { identityProviderNameCursor, IdentityProviderTokenCursor in
+            token_pair.identity_provider_name = identityProviderNameCursor
+            token_pair.identity_provider_token = IdentityProviderTokenCursor
+            return body(token_pair)
+        }
+    }
+}
+
+extension Array where Element == CognitoLoginPair {
```
waahm7, line 37: A better reusable way to do this would be in the CStruct so that we can avoid copy pasting this logic whenever we need an array list. Something like 
```
extension Array where Element: CStruct {
    withAWSArrayList<Result>..... {
    }
}
```

Source/AwsCommonRuntimeKit/auth/credentials/CredentialsProvider.swift
```diff
@@ -10,6 +10,68 @@ public protocol CredentialsProviding {
...
+                    fatalError("Unable to add user property")
+                }
+            }
+        }
+        return try body(OpaquePointer(array_list.pointee.data))
+    }
+}
+
+/// Helper function to convert Swift [CognitoLoginPair]? into a native aws_cognito_identity_provider_token_pair pointer
+func withOptionalCognitoLoginPair<Result>(
```
waahm7, line 64: <del>Similarly, we can just define a helper function on CStruct to keep it reusable. </del>. We won't need this function if we just make the array empty instead of nil.

Source/AwsCommonRuntimeKit/auth/credentials/CredentialsProvider.swift
```diff
@@ -561,6 +623,66 @@ extension CredentialsProvider.Source {
...
+    ///    - customRoleArn: (Optional) ARN of the role to be assumed when multiple roles were received in the token from the identity provider.
+    ///    - proxyOptions: (Optional) Http proxy configuration for the http request that fetches credentials
+    ///   - shutdownCallback:  (Optional) shutdown callback
+    /// - Returns: `CredentialsProvider`
+    /// - Throws: CommonRuntimeError.crtError
+    public static func `cognito`(bootstrap: ClientBootstrap,
+                                 tlsContext: TLSContext,
+                                 endpoint: String,
+                                 identity: String,
+                                 logins: [CognitoLoginPair]? = nil,
```
waahm7, line 643: Is there a difference between empty vs nil array? Personally, I prefer empty arrays to have one way to pass value instead of 2 ways but same outcome. You can also handle empty array in withAWSArrayList function to return nil.

Test/AwsCommonRuntimeKitTests/auth/CredentialsProviderTests.swift
```diff
@@ -218,6 +218,50 @@ class CredentialsProviderTests: XCBaseTestCase {
...
+
+
+            let provider = try CredentialsProvider(source: .cognito(bootstrap: getClientBootstrap(), tlsContext: getTlsContext(), endpoint: cognitoEndpoint, identity: cognitoIdentity, shutdownCallback: getShutdownCallback()))
+            let credentials = try await provider.getCredentials()
+            XCTAssertNotNil(credentials)
+        } catch is XCTSkip{ // skip the test as the environment var is not set
+            shutdownWasCalled.fulfill()
+        }catch {
+            exceptionWasThrown.fulfill()
+        }
```
waahm7, line 236: bad formatting
```suggestion
        } catch is XCTSkip { 
            // skip the test as the environment var is not set
            shutdownWasCalled.fulfill()
        } catch {
            exceptionWasThrown.fulfill()
        }
```

Test/AwsCommonRuntimeKitTests/auth/CredentialsProviderTests.swift
```diff
@@ -218,6 +218,50 @@ class CredentialsProviderTests: XCBaseTestCase {
...
 
+
+    func testCreateDestroyCognitoCredsProviderWithoutHttpProxy() async throws {
+        let exceptionWasThrown = XCTestExpectation(description: "Exception was thrown")
+        do {
+            let cognitoEndpoint = try getEnvironmentVarOrSkipTest(environmentVarName: "AWS_TEST_MQTT5_COGNITO_ENDPOINT")
+            let cognitoIdentity = try getEnvironmentVarOrSkipTest(environmentVarName: "AWS_TEST_MQTT5_COGNITO_IDENTITY")
+
+
+            let provider = try CredentialsProvider(source: .cognito(bootstrap: getClientBootstrap(), tlsContext: getTlsContext(), endpoint: cognitoEndpoint, identity: cognitoIdentity, shutdownCallback: getShutdownCallback()))
```
waahm7, line 229: Seems like a very long line.

Test/AwsCommonRuntimeKitTests/auth/CredentialsProviderTests.swift
```diff
@@ -218,6 +218,50 @@ class CredentialsProviderTests: XCBaseTestCase {
...
+
+            let httpProxys = HTTPProxyOptions(hostName: httpproxyHost, port: UInt32(httpproxyPort)!, connectionType: .tunnel)
+
+            let provider = try CredentialsProvider(source: .cognito(bootstrap: getClientBootstrap(), tlsContext: getTlsContext(), endpoint: cognitoEndpoint, identity: cognitoIdentity, shutdownCallback: getShutdownCallback()))
+            let credentials = try await provider.getCredentials()
+            XCTAssertNotNil(credentials)
+        }
+        catch is XCTSkip{ // skip the test as the environment var is not set
+            shutdownWasCalled.fulfill()
+        }
```
waahm7, line 258:
```suggestion
        catch is XCTSkip { 
            // skip the test as the environment var is not set
            shutdownWasCalled.fulfill()
        }
```

Source/AwsCommonRuntimeKit/auth/credentials/CredentialsProvider.swift
```diff
@@ -10,6 +10,60 @@ public protocol CredentialsProviding {
...
+        return withByteCursorFromStrings(IdentityProviderName,
+                                         IdentityProviderToken) { identityProviderNameCursor, IdentityProviderTokenCursor in
+            token_pair.identity_provider_name = identityProviderNameCursor
+            token_pair.identity_provider_token = IdentityProviderTokenCursor
+            return body(token_pair)
+        }
+    }
+}
+
+extension Array where Element: CStruct {
```
waahm7, line 37: Please move this to CStruct class. We can also replace the https://github.com/awslabs/aws-crt-swift/blob/iot/Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift#L48 with this helper.

Source/AwsCommonRuntimeKit/auth/credentials/CredentialsProvider.swift
```diff
@@ -10,6 +10,60 @@ public protocol CredentialsProviding {
...
+            aws_array_list_clean_up(array_list)
+            allocator.release(array_list)
+        }
+        guard aws_array_list_init_dynamic(
+            array_list,
+            allocator.rawValue,
+            count,
+            MemoryLayout<Element.RawType>.size) == AWS_OP_SUCCESS else {
+            fatalError("Unable to initialize array of user properties")
+        }
```
waahm7, line 54: We don't handle OOM errors. The C code already has a fatal assert if OOM.

Source/AwsCommonRuntimeKit/auth/credentials/CredentialsProvider.swift
```diff
@@ -10,6 +10,60 @@ public protocol CredentialsProviding {
...
+            return body(token_pair)
+        }
+    }
+}
+
+extension Array where Element: CStruct {
+    func withAWSArrayList<Result>(_ body: (OpaquePointer?) throws -> Result) rethrows -> Result {
+        if capacity == 0 {
+            return try body(nil)
+        }
```
waahm7, line 41: trivial/debatable: We can use guard here.

Source/AwsCommonRuntimeKit/auth/credentials/CredentialsProvider.swift
```diff
@@ -10,6 +10,60 @@ public protocol CredentialsProviding {
...
+        
+        let array_list: UnsafeMutablePointer<aws_array_list> = allocator.allocate(capacity: 1)
+        defer {
+            aws_array_list_clean_up(array_list)
+            allocator.release(array_list)
+        }
```
waahm7, line 47: Add a comment at the function level that CStruct mustn't contain any dynamically allocated memory, otherwise, it will leak.

waahm7 (approved):
Fix n ship.

Source/AwsCommonRuntimeKit/auth/credentials/CredentialsProvider.swift
xiazhvera: Besides the OOM, `aws_array_list_init_dynamic` also return error if the multiply result overflowed, which is not asserted. We probably would like to verify it succeed here.

PR #310, opened 2025-01-07
Fix Concurrency Warnings in Swift 6

### LINKED ISSUE #185: Use of GCD in asynchronous context
### Describe the bug

GCD semaphores can cause unpredictable behavior when used from within a Swift asynchronous task.

The following was seen in the build logs when building aws-crt-swift:
```
/Users/runner/work/aws-sdk-swift/aws-sdk-swift/target/build/deps/aws-crt-swift/Source/Elasticurl/Elasticurl.swift:320:23: warning: instance method 'wait' is unavailable from asynchronous contexts; Await a Task handle instead; this is an error in Swift 6
            semaphore.wait()
                      ^
Dispatch.DispatchSemaphore:4:17: note: 'wait()' declared here
    public func wait()
```

### Expected Behavior

aws-crt-swift does not use GCD semaphores within asynchronous code, and this warning is not generated

### Current Behavior

The warning above (soon to be raised to an error) is generated at build time

### Reproduction Steps

Build aws-crt-swift, observe build logs

### Possible Solution

_No response_

### Additional Information/Context

_No response_

### AWS CRT SWIFT version used

0.9.0

### Compiler and Version used

Xcode 14.3

### Operating System and version

macOS 13.3.1

waahm7:
Fixes #185 
*Description of changes:*
This PR fixes a bunch of Swift 6 errors (warnings for pre-Swift 6). 
- Marks a lot of our classes as unchecked Sendable; Swift cannot verify the Send-ability of these classes due to C pointers. 
- Refactors our tests to address some concurrency issues. 
- Marks the global allocator as unsafe, since Swift complains that global mutable things are not safe.


Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointProperty.swift
```diff
@@ -84,17 +84,17 @@ extension EndpointProperty: Decodable {
...
-        for (key, value) in self {
-            dict[key] = value.toAnyHashable()
-        }
-        return dict
-    }
-}
+//extension Dictionary where Key == String, Value == EndpointProperty {
+//    /// Converts EndpointProperty to a dictionary of `String`: `AnyHashable`
+//    /// - Returns: Dictionary of `String`: `AnyHashable`
+//    func toStringHashableDictionary() -> [String: AnyHashable] {
```
xiazhvera, line 90: The function seems not used. Is it safe to remove it?

Test/AwsCommonRuntimeKitTests/sdkutils/EndpointsRuleEngineTests.swift
```diff
@@ -220,22 +220,30 @@ class EndpointsRuleEngineTests: XCBaseTestCase {
...
-        ]
+        let expectedProperties: [String: EndpointProperty] = ["authSchemes": .array([
+            .dictionary([
+                "name": .string("sigv4"),
+                "signingName": .string("serviceName"),
+                "signingRegion": .string("us-west-2")
+            ])
+        ])]
         XCTAssertEqual(expectedProperties, properties)
+//        let expectedProperties = [
```
xiazhvera, line 237: Need clean up?

xiazhvera (approved):

Source/AwsCommonRuntimeKit/sdkutils/endpoint/EndpointProperty.swift
waahm7: Thanks, removed.

Test/AwsCommonRuntimeKitTests/sdkutils/EndpointsRuleEngineTests.swift
waahm7: Thanks, removed.

.github/workflows/ci.yml
```diff
@@ -71,6 +71,34 @@ jobs:
...
+        aws-region: ${{ env.AWS_DEFAULT_REGION }} 
+    - name: Checkout repository
+      uses: actions/checkout@v4
+      with:
+        submodules: true
+        fetch-depth: 0
+    - name: Set Swift version to 6
+      run: |
+        sed -i '' 's/swift-tools-version:5\.7/swift-tools-version:6\.0/' Package.swift
+        cat Package.swift | head -3
```
TingDaoK, line 96: if `Package.swift` doesn't have the exact string `swift-tools-version:5.7`, this step will just fail silently, which will make it hard to debug and probably leave the CI doesn't do the expected thing for us.


GenAI suggested something like:
```
Solution 1: Combined sed command with conditional exit
Bash
 Copy code
sed -i '' '/swift-tools-version:5\.7/{s/swift-tools-version:5\.7/swift-tools-version:6\.0/;b};q1' Package.swift
This will:
Search for the pattern
If found, make the substitution and branch to end of script (with b)
If not found, exit with code 1 (with q1)
```

Also, using the exact version seems less future proven. I am not sure if we will ever change the version in the package.swift, but in general, maybe just match all version instead.

waahm7: Thanks, I have made it more generic so that it works with any 5.xx version and also fails it can't replace it.

TingDaoK (approved):

PR #328, opened 2025-04-14
Swift6 Support for Mqtt

xiazhvera:
*Issue #, if available:*

*Description of changes:*


Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -209,8 +209,10 @@ public class Mqtt5Client {
 
 // MARK: - Internal/Private
 
+// IMPORTANT: You are responsible for concurrency correctness of Mqtt5ClientCore.
+// The rawValue is mutable cross threads and protected by the rwlock.
 /// Mqtt5 Client Core, internal class to handle Mqtt5 Client operations
-public class Mqtt5ClientCore: @unchecked Sendable {
+internal class Mqtt5ClientCore: @unchecked Sendable {
```
waahm7, line 215: The default visibility is `internal`

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -267,8 +273,10 @@ public class PubackPacket {
     }
 }
 
+// We can't mutate this class after initialization. Swift can not verify the sendability due to direct use of c pointer,
+// so mark it unchecked Sendable
```
waahm7, line 277: This makes me think that class being non-final is intentional. I think we mark it final but the actual reason is that it contains C Pointers/Structs. Same comment for all of them.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
```diff
@@ -237,8 +241,10 @@ public class PublishResult: @unchecked Sendable {
     }
 }
 
+// We can't mutate this class after initialization. Swift can not verify the sendability due to the class is non-final,
+// so mark it unchecked Sendable
 /// "Data model of an `MQTT5 PUBACK <https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901121>`_ packet
-public class PubackPacket {
+public class PubackPacket: @unchecked Sendable {
```
waahm7, line 247: Why is this a class? Can we just make it a struct with Sendable instead of Unchecked Sendable? We should verify all the classes.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -125,8 +125,8 @@ public typealias OnWebSocketHandshakeInterceptComplete = (HTTPRequestBase, Int32
 public typealias OnWebSocketHandshakeIntercept = (HTTPRequest, @escaping OnWebSocketHandshakeInterceptComplete) async -> Void
 
 // MARK: - Mqtt5 Client
-public class Mqtt5Client {
-    internal var clientCore: Mqtt5ClientCore
+public final class Mqtt5Client: Sendable {
+    internal let clientCore: Mqtt5ClientCore
```
waahm7, line 129: default visibility is internal https://docs.swift.org/swift-book/documentation/the-swift-programming-language/accesscontrol/#Default-Access-Levels. We should remove internal everywhere.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
waahm7: Normally, we have classes because C Pointers require a deinit function which structs can't have it.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
waahm7: Sorry, commented at the wrong comment. I was talking about some comments which say that we can't mark it Sendable because it's non final.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
xiazhvera: It's a little wired to say so, but the fundamental issue was I didnt found a clean way to handle an array of objects without storing raw data inside the class, as a result a `deinit` is required to properly deallocate that data. 
For this reason, any CStruct would potentially need a `deinit` if we would like to convert an Array of the object into c. I'm currently trying to find a good way to handle it the array conversion...

waahm7: > I didnt found a clean way to handle an array of objects without storing raw data inside the c
This class doesn't have any raw data

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift
xiazhvera: If you look at the classes referenced in those comments, you'll notice they don’t contain any C pointers or structs. If we mark those classes as `final`, we could declare them as `Sendable` directly instead of using `@unchecked Sendable`.

I didn’t initially mark them as final because I thought we might want to extend them in the future. For example, if IoT Core were to add support for QoS2, we might need to extend `PublishResult`. Of cause, it seems unlikely to happen at this point.

I'm fine either way (mark the classes as `final` or leave them as they are.)

waahm7: > I didn’t initially mark them as final because I thought we might want to extend them in the future. For example, if IoT Core were to add support for QoS2, we might need to extend PublishResult. Of cause, it seems unlikely to happen at this point.

That should be a non-breaking change. I believe marking a class final is a breaking change because other people might be extending it, but removing the final should be fine in the future.

xiazhvera:
Merged as this is required by Swift SDK , and I will update the comments in a follow up PR

PR #331, opened 2025-04-18
Fix IStreamable with End Of Stream Status

waahm7:
*Description of changes:*
Previous implementation always does one extra read to figure out that the stream has ended; this can result in sub-optimal performance where we send an extra empty chunk for the final empty read. Add the new isEndOfStream function to fix this issue.


TingDaoK (approved):
lgtm

only that the length implementation needs to seek twice and the value is not cached anywhere seem awkward, but genAI didn't provide a better way

waahm7:
> lgtm
> 
> only that the length implementation needs to seek twice and the value is not cached anywhere seem awkward, but genAI didn't provide a better way

Thanks, I have removed the FileHandle extension for now. I believe it was not used anywhere and we can add it when we need it.

PR #332, opened 2025-04-21
Request-Response Client Operations

xiazhvera:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
+++ b/Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
@@ -74,34 +74,60 @@ public struct IncomingPublishEvent: Sendable {
 }
 
 /// Function signature of a SubscriptionStatusEvent event handler
-public typealias SubscriptionStatusEventHandler = @Sendable (SubscriptionStatusEvent) async -> Void
+public typealias SubscriptionStatusEventHandler = (SubscriptionStatusEvent) -> Void
 
 /// Function signature of an IncomingPublishEvent event handler
-public typealias IncomingPublishEventHandler = @Sendable (IncomingPublishEvent) async -> Void
+public typealias IncomingPublishEventHandler = (IncomingPublishEvent) -> Void
 
 /// Encapsulates a response to an AWS IoT Core MQTT-based service request
 public struct MqttRequestResponseResponse {
     let topic: String
     let payload: Data
-    let error: CRTError?
     
     public init(topic: String, payload: Data, error: CRTError? = nil) {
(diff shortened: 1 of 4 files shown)


Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
```diff
@@ -149,11 +153,31 @@ public struct RequestResponseOperationOptions: CStructWithUserData, Sendable {
 
     typealias RawType = aws_mqtt_request_operation_options
     func withCStruct<Result>(userData: UnsafeMutableRawPointer?, _ body: (RawType) -> Result) -> Result {
-        // TODO: convert into aws_mqtt_request_operation_options
-        let options = aws_mqtt_request_operation_options()
-        return body(options)
+        var raw_options = aws_mqtt_request_operation_options()
+        return self.subscriptionTopicFilters.withMutableByteCursorArray { topicsCursor, len in
+            raw_options.subscription_topic_filters = topicsCursor
```
sbSteveK, line 158: This creation of the CStruct is pretty different from what we're doing for unsubscribe packets. Specifically the use of this func: https://github.com/awslabs/aws-crt-swift/blob/011a801f3c01aea854df5f01308222644c3778bb/Source/AwsCommonRuntimeKit/mqtt/Mqtt5Packets.swift#L465 to store the topic filter strings. Does this mean we're overcomplicating things for the unsubscribe packet? We should stick to a single pattern if possible. Right now I'm not entirely sure why we decided to make a copy of the [String] for use as an aws_byte_cursor but I feel like we probably had a reason. Wish we left better comments >,<

xiazhvera: Nice catch. We should be able to removed the `convertTopicFilters()` function and use the same pattern for unsubscribe packet.

sbSteveK (approved):

Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
```diff
@@ -337,7 +401,7 @@ public class MqttRequestResponseClient {
   /// - Parameters:
   ///     - operationOptions: configuration options for request response operation
   /// - Returns: MqttRequestResponse
-  /// - Throws: CommonRuntimeError.crtError if submit failed
+  /// - Throws:CommonRuntimeError.crtError if submit failed
```
sbSteveK, line 404: Trivial: Shouldn't there be a space between Throws: and `CommonRuntimeError.crtError` ?

PR #334, opened 2025-04-29
Fix ByteBuffer End of Stream Bug 

waahm7:
*Issue #, if available:*

*Description of changes:*
- Data.startIndex doesn't start with 0 if the Data is a slice in a larger portion. Use endIndex instead. 


TingDaoK (approved):
can we add some regression tests for this.

PR #335, opened 2025-04-29
[WIP]RR Streaming Operation

xiazhvera:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
+++ b/Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
@@ -202,7 +202,7 @@ internal func MqttRRStreamingOperationSubscriptionStatusCallback(_ eventType: aw
     let operationCore = Unmanaged<StreamingOperationCore>.fromOpaque(userData).takeUnretainedValue()
     operationCore.rwlock.read {
         // Only invoke the callback if the streaming operation is not closed.
-        if let rawValue = operationCore.rawValue, let callback = operationCore.options.subscriptionStatusEventHandler {
+        if let _ = operationCore.rawValue, let callback = operationCore.options.subscriptionStatusEventHandler {
             let subStatusEvent = SubscriptionStatusEvent(event: SubscriptionStatusEventType(eventType),
                                                          error: CRTError(code: Int32(errorCode)))
             callback(subStatusEvent)
--- a/Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
+++ b/Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
@@ -195,7 +195,19 @@ internal func MqttRRStreamingOperationIncomingPublishCallback(_ publishEvent: Un
 internal func MqttRRStreamingOperationSubscriptionStatusCallback(_ eventType: aws_rr_streaming_subscription_event_type,
                                                                  _ errorCode: Int32,
                                                                  _ userData: UnsafeMutableRawPointer?) {
-    // TODO: finish subscription status update handler callback
+    guard let userData else {
+        // No userData, directly return
+        return
+    }
+    let operationCore = Unmanaged<StreamingOperationCore>.fromOpaque(userData).takeUnretainedValue()
+    operationCore.rwlock.read {
+        // Only invoke the callback if the streaming operation is not closed.
+        if let rawValue = operationCore.rawValue, let callback = operationCore.options.subscriptionStatusEventHandler {
+            let subStatusEvent = SubscriptionStatusEvent(event: SubscriptionStatusEventType(eventType),
+                                                         error: CRTError(code: Int32(errorCode)))
+            callback(subStatusEvent)
+        }
+    }
 }
 
 /// Configuration options for streaming operations
@@ -226,7 +238,9 @@ public struct StreamingOperationOptions: CStructWithUserData, Sendable {
     }
 }
 
-internal class StreamingOperationCore {
+// IMPORTANT: You are responsible for concurrency correctness of StreamingOperationCore.
+// The rawValue is mutable cross threads and protected by the rwlock.
+internal class StreamingOperationCore: @unchecked Sendable {
     fileprivate var rawValue: OpaquePointer? // <aws_mqtt_rr_client_operation>?
     fileprivate let rwlock = ReadWriteLock()
     fileprivate let options: StreamingOperationOptions
@@ -246,7 +260,7 @@ internal class StreamingOperationCore {
     internal func open() {
         rwlock.read {
             if let rawValue = self.rawValue {
-                aws_mqtt_rr_client_operation_activate(self.rawValue)
+                aws_mqtt_rr_client_operation_activate(rawValue)
             }
         }
     }
--- a/Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
+++ b/Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
@@ -228,8 +228,11 @@ public struct StreamingOperationOptions: CStructWithUserData, Sendable {
 
 internal class StreamingOperationCore {
     fileprivate var rawValue: OpaquePointer? // <aws_mqtt_rr_client_operation>?
-
+    fileprivate let rwlock = ReadWriteLock()
+    fileprivate let options: StreamingOperationOptions
+    
     internal init (streamOptions: StreamingOperationOptions, client: MqttRequestResponseClientCore) throws {
+        self.options = streamOptions
         let rawValue = streamOptions.withCPointer(userData: Unmanaged<StreamingOperationCore>.passRetained(self).toOpaque()) { optionsPointer in
             return aws_mqtt_request_response_client_create_streaming_operation(client.rawValue, optionsPointer)
         }
@@ -241,12 +244,19 @@ internal class StreamingOperationCore {
     
     /// Opens a streaming operation by making the appropriate MQTT subscription with the broker.
     internal func open() {
-        // TODO: open the stream
+        rwlock.read {
+            if let rawValue = self.rawValue {
+                aws_mqtt_rr_client_operation_activate(self.rawValue)
+            }
+        }
     }
     
     /// Closes the operation
     internal func close() {
-        aws_mqtt_rr_client_operation_release(self.rawValue)
+        rwlock.write {
+            aws_mqtt_rr_client_operation_release(self.rawValue)
+            self.rawValue = nil
+        }
     }
 }
 
@@ -254,7 +264,7 @@ internal class StreamingOperationCore {
 public class StreamingOperation {
     fileprivate var operationCore: StreamingOperationCore
     
-    /// The end user should create the operation through MqttRequestResponseClient->createStream()
+    /// The end user should init the operation through MqttRequestResponseClient->createStream()
     fileprivate init(operationCore: StreamingOperationCore) {
         self.operationCore = operationCore
     }
--- a/Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
+++ b/Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
@@ -70,6 +70,15 @@ public struct IncomingPublishEvent: Sendable {
     /// The topic associated with this PUBLISH packet.
     let topic: String
     
+    /// The content type of the payload
+    let contentType: Data
+
+    /// User Property
+    let userProperties: [UserProperty]
+    
+    ///
+    let messageExpiryInterval: TimeInterval
+    
     // TODO: More options for IoT Command changes
 }
 
--- a/Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
+++ b/Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
@@ -180,45 +180,92 @@ public struct RequestResponseOperationOptions: CStructWithUserData, Sendable {
     }
 }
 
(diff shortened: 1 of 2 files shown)


Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
```diff
@@ -60,17 +51,36 @@ public struct SubscriptionStatusEvent: Sendable {
...
+    /// (Optional) The content type of the payload
+    public let contentType: String?
 
-    /// The topic associated with this PUBLISH packet.
-    let topic: String
+    /// (Optional) User Properties, if there is no user property, the array is empty
+    public let userProperties: [UserProperty]
     
-    // TODO: More options for IoT Command changes
+    /// (Optional) Some service use this field to specify client-side timeouts.
```
sbSteveK, line 69: Trivial: "services"

Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
```diff
@@ -60,17 +51,36 @@ public struct SubscriptionStatusEvent: Sendable {
...
     /// The payload of the publish message in a byte buffer format
-    let payload: Data
+    public let payload: Data
+    
+    /// (Optional) The content type of the payload
+    public let contentType: String?
 
-    /// The topic associated with this PUBLISH packet.
-    let topic: String
+    /// (Optional) User Properties, if there is no user property, the array is empty
```
sbSteveK, line 66: Trivial: Should we allow this to be nil by making it optional ( [UserProperty]? ) ? Other optional properties are treated as optional so it feels a little off for this "Optional" to not be.

Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
```diff
@@ -181,60 +220,154 @@ public struct RequestResponseOperationOptions: CStructWithUserData, Sendable {
...
+
 /// Configuration options for streaming operations
-public struct StreamingOperationOptions: CStruct, Sendable {
-    let subscriptionStatusEventHandler: SubscriptionStatusEventHandler
-    let incomingPublishEventHandler: IncomingPublishEventHandler
-    let topicFilter: String
+public struct StreamingOperationOptions: CStructWithUserData, Sendable {
+    /// The MQTT topic that the streaming is "listen" to.
+    public let topicFilter: String
+    /// (Optional) The handler function a streaming operation will use for subscription status events.
```
sbSteveK, line 268: Trivial: Our service clients will ALWAYS pass down these callbacks. They should never be nil and if they are, something is broken with the generated service client. We can optionally change this to a non optional property and handle it as such.

Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
```diff
@@ -327,58 +461,58 @@ internal class MqttRequestResponseClientCore: @unchecked Sendable {
         }
     }
     
-    /// create a stream operation, throws CRTError if the creation failed. You would need call open() on the operation to start the stream
-    public func createStream(streamOptions: StreamingOperationOptions) throws -> StreamingOperation {
-        // TODO: create streamming operation
-        return StreamingOperation()
+    /// Creates a stream operation, throws CRTError if the creation failed. You would need call open() on the operation to start the stream
```
sbSteveK, line 464: Trivial: "You would need call open() on the operation" -> "You will need to call open() on the operation"

Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
```diff
@@ -327,58 +461,58 @@ internal class MqttRequestResponseClientCore: @unchecked Sendable {
...
+    fileprivate func createStream(streamOptions: StreamingOperationOptions) throws -> StreamingOperation {
+        return try StreamingOperation(streamOptions: streamOptions, client: self)
     }
     
-    /// release the request response client. You must not use the client after call `close()`.
-    public func close() {
+    /// Releases the request response client. You must not use the client after call `close()`.
```
sbSteveK, line 469: Trivial: "You must not use the client after call `close()`." "You must not use the client after calling `close()`."

Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
```diff
@@ -327,58 +461,58 @@ internal class MqttRequestResponseClientCore: @unchecked Sendable {
...
+    fileprivate func close() {
         aws_mqtt_request_response_client_release(self.rawValue)
         self.rawValue = nil
     }
     
 }
 
+/// The Mqtt
```
sbSteveK, line 477: Is this missing something?

Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
```diff
@@ -181,60 +220,154 @@ public struct RequestResponseOperationOptions: CStructWithUserData, Sendable {
...
+    fileprivate var operationCore: StreamingOperationCore
+    
+    /// Constructor. The end user should only create the operation through MqttRequestResponseClient->createStream()
+    fileprivate init (streamOptions: StreamingOperationOptions, client: MqttRequestResponseClientCore) throws {
+        self.operationCore = try StreamingOperationCore(streamOptions: streamOptions, client: client)
+    }
+    
+    /// Opens a streaming operation by making the appropriate MQTT subscription with the broker.
+    /// - Throws: CommonRunTimeError
+    public func open() throws {
```
sbSteveK, line 345: Are we missing a close() function? opened streams need to be able to be closed to free up streaming subscriptions.

Test/AwsCommonRuntimeKitTests/mqtt/MqttRRClientTests.swift
```diff
@@ -72,11 +88,35 @@ class Mqtt5RRClientTests: XCBaseTestCase {
...
+                        self.subscriptionStatusErrorExpectation.fulfill()
+                    }
+                }
+            }
+            
         }
         
-        // release the context before exit the test case
-        func cleanup(){
+        // make sure cleanup the resources before exit the test caseto
```
sbSteveK, line 116: Trivial: "make sure to cleanup" or "make sure we cleanup" and there's a trailing 'to' after the end of 'case'

Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
xiazhvera: Close() is automatically called on destructor. We don't allow end user manually call close() here.

Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
```diff
@@ -327,58 +461,58 @@ internal class MqttRequestResponseClientCore: @unchecked Sendable {
...
+        return try MqttRequestResponseClient(mqttClient: mqtt5Client, options: options ?? MqttRequestResponseClientOptions())
     }
     
-    internal init(mqttClient: Mqtt5Client, options: MqttRequestResponseClientOptions) throws {
+    init(mqttClient: Mqtt5Client, options: MqttRequestResponseClientOptions) throws {
         clientCore = try MqttRequestResponseClientCore(mqttClient: mqttClient, options: options)
     }
     
-    /// Submit a request responds operation, throws CRTError if the operation failed
+    /// Submits a generic request to the request-response client, throws CRTError if the operation failed
```
sfod, line 499: Trivial: "Submits a generic request to the request-response client" -> "Initiates a request-response operation"

bretambrose (commented):
Please no "ResponseResponse". If I did it in the C API it was a mistake there too, but let's not propagate it.

Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
```diff
@@ -60,17 +51,38 @@ public struct SubscriptionStatusEvent: Sendable {
...
   /// The payload of the publish message in a byte buffer format
-  let payload: Data
+  public let payload: Data
 
-  /// The topic associated with this PUBLISH packet.
-  let topic: String
+  /// (Optional) The content type of the payload
+  public let contentType: String?
+
+  /// (Optional) User Properties, if there is no user property, the array will be empty
```
sbSteveK, line 66: Trivial: "be an empty array" ? for clarity

Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
```diff
@@ -60,17 +51,38 @@ public struct SubscriptionStatusEvent: Sendable {
...
+  public let userProperties: [UserProperty]
+
+  /// (Optional) Some services use this field to specify client-side timeouts.
+  public let messageExpiryInterval: TimeInterval?
 
-  // TODO: More options for IoT Command changes
+  /// Internal constructor that setup IncomingPublishEvent from native aws_mqtt_rr_incoming_publish_event
```
sbSteveK, line 72: Trivial: "Internal constructor that initializes an `IncomingPublishEvent` from a native `aws_mqtt_rr_incoming_publish_event`"

Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
```diff
@@ -79,11 +91,14 @@ public typealias SubscriptionStatusEventHandler = @Sendable (SubscriptionStatusE
...
 
-/// Encapsulates a response to an AWS IoT Core MQTT-based service request
+/// A response to an AWS IoT Core MQTT-based service request
 public struct MqttRequestResponse: Sendable {
+  /// The MQTT Topic that the response was received on.
   public let topic: String
+  /// Payload of the response that correlates to a submitted request.
   public let payload: Data
 
+  /// Internal constructor that setup MqttRequestResponse from native aws_mqtt_rr_incoming_publish_event
```
sbSteveK, line 101: Trivial: Same as above. Or you can use "sets up" instead of "setup". It just shouldn't be "that setup <Thing>"

Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
```diff
@@ -96,10 +111,17 @@ public struct MqttRequestResponse: Sendable {
 /// A response path is a pair of values - MQTT topic and a JSON path - that describe where a response to
 /// an MQTT-based request may arrive.  For a given request type, there may be multiple response paths and each
 /// one is associated with a separate JSON schema for the response body.
-public class ResponsePath: CStruct, @unchecked Sendable {
+final public class ResponsePath: CStruct, @unchecked Sendable {
```
sbSteveK, line 114: Trivial: Other areas you've used @unchecked explains why it's unchecked. Should it also be explained here?

Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
```diff
@@ -96,10 +111,17 @@ public struct MqttRequestResponse: Sendable {
...
+  /// MQTT topic that a response may arrive on.
   let topic: String
+  /// JSON path for finding correlation tokens within payloads that arrive on this path's topic.
```
sbSteveK, line 117: This feels misleading because it's labeled here and elsewhere as "JsonPath" but it's actually just the unique string used to correlate. Unsure what we should do from a naming the member perspective but I don't think we should say it's a "JSON path" in the description we provide.

Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
```diff
@@ -96,10 +111,17 @@ public struct MqttRequestResponse: Sendable {
...
   let correlationTokenJsonPath: String
 
+  /// Create a response path
+  ///
+  /// - Parameters:
+  ///     - topic: MQTT topic that a response may arrive on.
+  ///     - correlationTokenJsonPath: JSON path for finding correlation tokens within payloads that arrive on this path's topic.
```
sbSteveK, line 124: Same as above

Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
```diff
@@ -358,30 +509,32 @@ internal class MqttRequestResponseClientCore: @unchecked Sendable {
...
+/// The MqttRequestResponseClient is a client that allows you to send requests and receive responses over MQTT.
 public class MqttRequestResponseClient {
   fileprivate var clientCore: MqttRequestResponseClientCore
 
   /// Creates a new request-response client using an MQTT5 client for protocol transport
   ///
   /// - Parameters:
-  ///     - mqtt5Client: protocolClient MQTT client to use for transport
-  ///     - options: request-response client configuration options
+  ///     - mqtt5Client: the MQTT5 client that use for protocol transport
```
sbSteveK, line 534: Trivial: "MQTT5 client that the request-response client should use as transport"

sbSteveK (approved):
Review trivial suggestions then ship

Source/AwsCommonRuntimeKit/mqtt/MqttRequestResponseClient.swift
xiazhvera: Calling it a "path" still feels accurate to me. Currently, the correlation token is just a top-level string, so the "path" is simply "clientToken" like in this example:
```
Response{
  "clientToken": "<token>"
}
```
But our current design leaves room for future responses where the token could be nested. For example, in the following case, the path might be something like "level1.level2.clientToken".
```
Response{
  "level1": {
    "level2" : {
        "clientToken": "<token>"
    }
  }
}
```

PR #339, opened 2025-05-02
Remove Swift Lint and Use Swift-Format

waahm7:
*Issue #, if available:*
#74 

*Description of changes:*
- Remove SwiftLint Formatter, it had the following problems
   - No good way to automatically format everything
   - VSCode auto formatting had different formatting
   - Warning and Error levels for formatting issues. The warning results in just getting ignored indefinitely because sometimes they were from Code that you didn't touch. 
   
- Use Swift-Format (Not Swift Format which is a different popular 3p tool). It has the following advantages
    - Officially by Apple and automatically included in the Swift toolchain.
    - Only one customization required. We should try to avoid customizing the formatter and just use defaults.
    - One command to automatically format all swift files.
    - No warnings, only errors
    - Defaults seem sensible enough like line length etc so that we don't need to customize it a lot. 

From Rust Fmt experience, I believe that an auto formatter is crucial so that we can ignore personal formatting preferences and just focus on coding.


sbSteveK (changes requested):
Let's pull out the non-formatting related changes into a separate PR. Formatting shows a huge diff and we don't want to lose the point we made non-formatting changes to ci.yml.

waahm7:
> Let's pull out the non-formatting related changes into a separate PR. Formatting shows a huge diff and we don't want to lose the point we made non-formatting changes to ci.yml.

Thanks, updated.

.github/workflows/ci.yml
```diff
@@ -34,11 +34,13 @@ jobs:
...
+    runs-on: macos-15 # latest
     steps:
-      - uses: actions/checkout@v4
-      - name: GitHub Action for SwiftLint
-        uses: norio-nomura/action-swiftlint@3.2.1
+      - uses: actions/checkout@v3
+      - name: Run Swift-Format
+        run: |
+          brew install swift-format
+          swift-format lint --recursive --strict .
```
xiazhvera, line 43: Can we specify a swift-format version so that we could use the same version across the team.

waahm7: Great question but unfortunately for swift-format, I don't think there is an easy way to specify a version. The only option I can think of is building from source, which will be a pain. We can do something like brew install swift-format@version, but as new versions will be released, brew won't keep the old versions.

Using the latest from brew should work fine as we can just do brew upgrade whenever new swift-format versions are released with any breaking changes.

waahm7: The problem with this is that we will need to do a reformatting PR as soon as new versions of Swift-format are released with incompatible changes but I'm not sure how often that would be.

waahm7: @xiazhvera Any concerns with this approach or are we good to ship it?

xiazhvera: No, good to go.

PR #344, opened 2025-06-02
Android support

gevartosky:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-swift/issues/4000

*Description of changes:*
These changes seek to add Android support to `amplify-swift` and its dependencies.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,8 @@ jobs:
     steps:
       - name: "Checkout"
         uses: actions/checkout@v4
+        with:
+          submodules: recursive
       - name: "Build for Android"
         uses: skiptools/swift-android-action@v2
         with:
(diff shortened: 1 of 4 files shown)


gevartosky:
I understand there are android repos but I'm specifically looking for Android support in this Swift repository as Swift runs stably on Android, has an Android Working Group which has filed the request to make the android swift toolchain official last year.

The usage of swift on Android has been increasing and it really simplify development if we don't need to keep bridging code between `amplify-swift` and `amplify-android`

sbSteveK:
Hi @gevartosky, Thank you for your interest in aws-crt-swift and adding Android as a platform. This is something we would consider a feature request and the work involved in properly supporting Android via aws-crt-swift would take a material amount of work. Please create an Issue for this here: https://github.com/awslabs/aws-crt-swift/issues which will allow other users to comment and voice their desire for this feature.

In the meantime, if you need this specific work done, I suggest you fork the repo with the changes you've made. Thanks!

PR #347, opened 2025-08-08
Mqtt5 Canary Application

xiazhvera:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Source/Canary/Mqtt5Canary/Mqtt5Canary.swift
+++ b/Source/Canary/Mqtt5Canary/Mqtt5Canary.swift
@@ -1,0 +1,535 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+
+import ArgumentParser
+import AwsCommonRuntimeKit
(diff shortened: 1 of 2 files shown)


Source/Canary/Mqtt5Canary/Mqtt5Canary.swift
```diff
@@ -0,0 +1,559 @@
...
+    }
+  }
+}
+
+func Mqtt5CanaryOperationDistributionSetup(_ distributionDataSet: inout [Mqtt5CanaryOperation]) {
+  let operationDistribution = [
+    (Mqtt5CanaryOperation.STOP, 1),
+    (Mqtt5CanaryOperation.SUBSCRIBE, 200),
+    (Mqtt5CanaryOperation.UNSUBSCRIBE, 200),
+    (Mqtt5CanaryOperation.UNSUBSCRIBE_BAD, 100),
```
sbSteveK, line 385: Trivial: Since we may hit the bad unsub due to sub and unsub being equal, should we lower this further to prevent an excess of unsub bad? I don't think it really matters...

PR #349, opened 2025-08-28
Mqtt5 Canary Codebuild Setup

xiazhvera:
*Issue #, if available:*

*Description of changes:*
1. Update Canary Application. Use actor to make sure the test context is synced properly. 
2. Add canary script for dispatch queue and kqueue


sbSteveK:
Let's move Mqtt5Canary out of the Source folder and into the Test folder if possible. I think we can move the files and point to it in Package.swift. If it's not possible, that's fine but ideally we don't want something customers shouldn't be looking at in the main Source folder.

Source/Canary/Mqtt5Canary/Mqtt5Canary.swift
```diff
@@ -124,65 +130,85 @@ class Mqtt5CanaryTestContext: @unchecked Sendable {
...
-    if !canaryClient.is_connected {
+    if await canaryClient.getConnected() == false {
       await statistic.incrementTotalOperation()
-      try canaryClient.client!.start()
+      if let _client = await canaryClient.client {
+        // Client already exists, just start it
+        try _client.start()
+      } else {
+        // Client does not exist, create a new one
+        throw CanaryTestError.InvalidArgument
```
sbSteveK, line 140: We currently only have one type of `CanaryTestError` of `InvalidArgument`. It might be useful to expand the enums so we throw an error that points to what exactly happened. e.g. Client index not found, InvalidOperation, various operational failures.

Source/Canary/Mqtt5Canary/Mqtt5Canary.swift
```diff
@@ -269,9 +302,13 @@ class Mqtt5CanaryTestContext: @unchecked Sendable {
...
       let canaryClient = try getCanaryClient(clientIndex)
-      try canaryClient.client!.stop()
+      if let _client = await canaryClient.client {
+        try _client.stop()
+      } else {
+        throw CanaryTestError.InvalidArgument
+      }
       await statistic.incrementTotalOperation()
-      canaryClient.client = nil
+      await canaryClient.resetClient()
```
sbSteveK, line 311: following this we call `mqtt5CanaryClients.remove(at: mqtt5CanaryClients.index(mqtt5CanaryClients.startIndex, offsetBy: clientIndex))`. Would it be safer to get the `canaryClient.clientId` after we get a client using the clientIndex above and then remove the client using the key instead of an index that could potentially shift around with `mqtt5CanaryClients.removeValue(forKey: clientId)`? In general, since mqtt5CanaryClients is established as `[String: Mqtt5CanaryClient]` I feel a little uneasy about using an index instead of the actual key everywhere. Maybe it's fine and there are no race conditions to be worried about here since these operations SHOULD be happening synchronously but using clientIds instead of index would make sure we're not doing things unintentionally

codebuild/canary/CanaryWrapper_Classes.py
```diff
@@ -0,0 +1,1159 @@
...
+                        self.restart_monitoring()
+                    else:
+                        self.print_message(
+                            "[ApplicationMonitor] Monitor application has stopped and monitor is not supposed to restart... Finishing...")
+                        self.error_has_occurred = True
+                        self.error_reason = "Canary Application Finished"
+                        self.error_code = 0
+            else:
+                self.print_message(
+                    "[ApplicationMonitor] Monitor application is still running...")
```
sbSteveK, line 1037: Let's remove this. It prints a wall of this in logs before we see the useful data snapshots in between.

codebuild/canary/mqtt5-swift-canary-kqueue-test.yml
```diff
@@ -0,0 +1,31 @@
...
+      - python3 -m pip install psutil
+      - python3 -m pip install boto3
+  build:
+    commands:
+      - echo Build started on `date`
+      - export GIT_HASH=$(git rev-parse HEAD)
+      # Get the endpoint, certificate, and key for connecting to IoT Core
+      - export ENDPOINT=$(aws secretsmanager get-secret-value --secret-id ${CANARY_SERVER_ARN} --query "SecretString" | cut -f2 -d":" | sed -e 's/[\\\"\}]//g')
+      # Run the Canary
+      - python3 codebuild/canary/CanaryWrapper.py --canary_executable swift --canary_arguments "run Mqtt5Canary --endpoint ${ENDPOINT} --seconds ${CANARY_DURATION} --tps ${CANARY_TPS} --threads ${CANARY_THREADS} -c ${CANARY_CLIENT_COUNT} --verbose ${CANARY_LOG_LEVEL}" --git_hash ${GIT_HASH} --git_repo_name $PACKAGE_NAME --ticket_item "${CODEBUILD_TICKET_ITEM}""
```
sbSteveK, line 28: Looks like an extra " at the end of this line.

codebuild/canary/CanaryWrapper_MetricFunctions.py
```diff
@@ -0,0 +1,52 @@
...
+            print("ERROR - No psutil.process passed! Cannot gather metric!", flush=True)
+            return None
+        # We always need to skip the first CPU poll
+        if (cache_cpu_psutil_process != psutil_process):
+            psutil.cpu_percent(interval=None)
+            cache_cpu_psutil_process = psutil_process
+            return None
+        return psutil.cpu_percent(interval=None)
+    except Exception as e:
+        print("ERROR - exception occurred gathering metrics!")
```
sbSteveK, line 24: There are three separate instances of this pattern of exception handling. Let's call out which part of the metrics gathering the metrics exception was encountered in. e.g. "exception occurred gathering memory usage value metrics"

sbSteveK (approved):
Some minor change requests but overall looks good. I'd say if we don't add an alarm related to CPU usage we should probably think about not tracking it at all.

PR #378, opened 2026-03-09
Manual Puback Control

sbSteveK:
Bind manual puback
Opaque manual puback control handle

Tests for Manual Puback


Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -533,6 +592,24 @@ internal func MqttClientHandleLifecycleEvent(
   }
 }
 
+/// A box that holds a `PublishAcknowledgementHandle` and allows it to be taken exactly
+/// once. Used to implement the "take once" semantics for manual publish acknowledgement control.
+/// The first call to `take()` returns the handle and clears the box. subsequent calls return `nil`.
+private final class PublishAcknowledgementHandleBox: @unchecked Sendable {
```
xiazhvera, line 599: Sendable objects should be thread-safe. Swift will validate the classes that are marked as Sendable, and the @uncheck the flag will disable the validation. If we use "unchecked", we should justify why we explicitly mark the class as "unchecked" Sendable in the comments. 

Most of our data class are immutable, which justify that it is thread-safe.  (example: LifecycleDisconnectData:  https://github.com/awslabs/aws-crt-swift/blob/9c6a3c641dca1570c8e2f2a0a1648dcdb80fd453/Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift#L86 ) 

For mutable classes as Mqtt5Client, we had locks to explicitly protect the class:  https://github.com/awslabs/aws-crt-swift/blob/9c6a3c641dca1570c8e2f2a0a1648dcdb80fd453/Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift#L223-L225) 

We should have locks or actor to protect the handle to make sure it is thread-safe.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -542,13 +619,51 @@ internal func MqttClientHandlePublishRecieved(
...
+    // For QoS 1 messages, we eagerly acquire the publish acknowledgement control ID from native
+    // BEFORE invoking the user callback. The control ID is acquired up front, then the user is
+    // given a closure (acquirePublishAcknowledgement) to "take" the handle within the callback.
+    // After the callback returns, if the user did not take control, the publish acknowledgement
+    // is sent automatically.
+    var publishAcknowledgementId: UInt64 = 0
+    var handleBox: PublishAcknowledgementHandleBox? = nil
+    var acquirePublishAcknowledgement: (@Sendable () -> PublishAcknowledgementHandle?)? = nil
+
+    if publishPacket.qos == .atLeastOnce, let rawValue = clientCore.rawValue {
```
xiazhvera: We already have rawValue validation at line 622.

Test/AwsCommonRuntimeKitTests/mqtt/Mqtt5ClientTests.swift
```diff
@@ -2028,6 +2028,536 @@ class Mqtt5ClientTests: XCBaseTestCase, @unchecked Sendable {
...
+      secondNil,
+      "Second call to acquirePublishAcknowledgement() on the same message should return nil")
+
+    try await stopClient(client: client, testContext: testContext)
+  }
+
+  /*
+  * [ManualPuback-UC4] Calling acquirePublishAcknowledgement() after the callback returns also returns nil
+  */
+  func testMqtt5ManualPubackAcquirePostCallbackReturnsNil() async throws {
```
xiazhvera, line 2426: Could we have a test that call acquirePublishAcknowledgement in a task in the publish received callback?

acquirePublishAcknowledgement() should return nil in this case as it is called outside the callback. However, if the task is fired before the callback return, it could still acquire the publish.

xiazhvera (changes requested):
The PublishAcknowledgementHandleBox doesn't seem to be thread-safe.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
sbSteveK: removed

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
sbSteveK: Adding a lock

Test/AwsCommonRuntimeKitTests/mqtt/Mqtt5ClientTests.swift
sbSteveK: confirmed offline that this test was to check pre-locks.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
```diff
@@ -542,13 +623,51 @@ internal func MqttClientHandlePublishRecieved(
...
+
+    clientCore.onPublishReceivedCallback(publishReceivedData)
+
+    // After the callback returns, check whether the user took control of the publish
+    // acknowledgement. If the handle is still in the box, auto-invoke the
+    // publish acknowledgement.
+    if publishAcknowledgementId != 0, let box = handleBox, box.take() != nil,
+      let rawValue = clientCore.rawValue
+    {
+      aws_mqtt5_client_invoke_publish_acknowledgement(rawValue, publishAcknowledgementId, nil)
```
xiazhvera, line 669: Should we do error handling for `aws_mqtt5_client_invoke_publish_acknowledgement`? `aws_mqtt5_client_invoke_publish_acknowledgement` seems always return true. In this case it seems unnecessary for error handling in invokePublishAcknowledgement neither.

xiazhvera (approved):
LGTM. Fix & ship.

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Client.swift
sbSteveK: thought about it and I don't know that we should do anything. Any failure would be logged as an error in native and it didn't make sense to me to throw or do anything that the user needs to handle from the perspective of a publish received callback.

PR #383, opened 2026-05-18
IoT Metrics Implementation

xiazhvera:
*Issue #, if available:*

*Description of changes:*
The review will focus on high level API changes and structure. It include the following changes 
1. Add  `IoTDeviceSDKMetrics` structure. The struct will hold metrics information. 
`AwsIoTSDKMetrics` (private) -> `IoTDeviceSDKMetrics` (public). 
2. MqttClientOptions 
    - `enableMetrics` -> `disableMetrics` : change name convention to `disable` to focus on the "opt-out" pattern and reduce cognitive load.
    - Add metrics option, so user can now set their own metrics
3. Add IoTSDKMetricsEncoder (Internal struct) to help build the encoded metrics string  

### DIFF
--- a/Source/AwsCommonRuntimeKit/mqtt/IoTSDKMetrics.swift
+++ b/Source/AwsCommonRuntimeKit/mqtt/IoTSDKMetrics.swift
@@ -1,0 +1,355 @@
+///  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+///  SPDX-License-Identifier: Apache-2.0.
+
+import AwsCMqtt
+import Foundation
+
+// MARK: - IoT Device SDK Metrics
+
+/// Configuration for the AWS IoT Device SDK Metrics.
+/// This structure is used to pass metrics configuration from the SDK layer through the CRT to aws-c-mqtt.
+///
+/// The metrics will be appended to the MQTT CONNECT packet's username field.
+public class IoTDeviceSDKMetrics: CStruct {
+  /// The library name identifier for the SDK (e.g., "IoTDeviceSDK/Swift")
+  /// This maps to the SDK attribute in the username field.
+  public let libraryName: String
+
+  /// Metadata dictionary for key-value pairs.
+  public var metadata: [String: String]
+
+  /// Default library name for Swift SDK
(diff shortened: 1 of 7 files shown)


rakshil14-2 (approved):
LGTM

Source/AwsCommonRuntimeKit/mqtt/Mqtt5Options.swift
```diff
@@ -294,10 +279,11 @@ public class MqttClientOptions: CStructWithUserData {
   public let onLifecycleEventDisconnectionFn: OnLifecycleEventDisconnection?
 
   /// Enable AWS IoT Metrics, including SDK name, version, and platform. Default to True.
-  public let enableMetrics: Bool
+  public let disableMetrics: Bool
```
sfod: nit: Comment needs to be updated.

Source/AwsCommonRuntimeKit/mqtt/IoTSDKMetrics.swift
```diff
@@ -0,0 +1,401 @@
...
+    var userFeatureString: String = ""
+
+    if let userMetadata = options.metrics?.metadata {
+
+      if let userFeatureVersion = userMetadata["IoTSDKMetricsVersion"],
+        Int(userFeatureVersion) == ioTSDKMetricsFeatureVersion,
+        let userFeature = userMetadata["IoTSDKFeature"]
+      {
+        userFeatureString = userFeature
+      }
```
sfod, line 273: I can't understand why `IoTSDKMetricsVersion` is needed. Is this field supposed to be set on iot sdk layer by mqtt builder?

xiazhvera: The metrics version needs to be defined in both CRT and IoT SDKs. If the version input from the IoT SDK doesn’t match, we’ll ignore the metrics input from it to avoid inconsistencies

sfod (approved):
LGTM. The only note is it'd be nice to get socket implementation details from aws-c-io where logic is implemented, to avoid potential data discrepancy when we change it. But we can implement it later if needed.

PR #384, opened 2026-05-18
workflow to set crt version string

xiazhvera:
*Issue #, if available:*

*Description of changes:*
Added a workflow to setup CRT version string. Whenever the team cut it will automatically grab the release tag and update the `CRTVersion` version string.

HOW IT WORKS:
   1. Triggers when a release is published via the GitHub UI
   2. Extracts the semantic version from the tag (e.g., v0.61.0 -> 0.61.0)
   3. Updates CommonRuntimeKit.swift with the new version
   4. Commits the change to main branch
   5. Moves the release tag to point to the new commit (with updated version) 

### DIFF
--- a/.github/workflows/update-version.yml
+++ b/.github/workflows/update-version.yml
@@ -3,6 +3,7 @@ name: Update Version on Release
 # This workflow automatically updates the CRTVersion in CommonRuntimeKit.swift when a new release is created.
 # It extracts the version from the tag (e.g., v0.61.0 -> 0.61.0), updates the source file,
 # and moves the tag to point to the new commit with the updated version.
+# The original release title and body are preserved when recreating the release.
 
 on:
   release:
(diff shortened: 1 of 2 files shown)


.github/workflows/update-version.yml
```diff
@@ -0,0 +1,90 @@
...
+          # Push the commit to main
+          git push origin main
+          
+          # Delete the old tag (both locally and remotely)
+          git tag -d "$TAG_NAME"
+          git push origin --delete "$TAG_NAME"
+          
+          # Create a new tag pointing to the new commit with the updated version
+          git tag -a "$TAG_NAME" -m "Release $TAG_NAME"
+          git push origin "$TAG_NAME"
```
TingDaoK, line 110: looks like we are removing the release message here. Comparing to the one in aws-crt-cpp. https://github.com/awslabs/aws-crt-cpp/blob/main/update-version.sh#L35-L45

TingDaoK (approved):

PR #386, opened 2026-05-21
Setup deploy key

xiazhvera:
*Issue #, if available:*

*Description of changes:*
- Use deploy key to grant the admin permission for automation


TingDaoK (approved):
worth to move this to some shared place? eg: maybe an action in builder? So that we only need one place to update if any new rules come out

PR #387, opened 2026-05-21
Support s2n on macOS

sfod:
Add support for s2n-tls as a TLS backend on macOS.
Apple Secure Transport remains the default for FIPS compliance.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -126,7 +126,7 @@ var awsCCalPlatformExcludes =
   packageTargets.append(
     .target(
       name: "S2N_TLS",
-      dependencies: ["LibCrypto"],
+      dependencies: [.target(name: "LibCrypto", condition: .when(platforms: [.macOS, .linux]))],
       path: "aws-common-runtime/s2n",
       exclude: s2nExcludes,
       publicHeadersPath: "api",
--- a/Package.swift
+++ b/Package.swift
@@ -81,7 +81,7 @@ var calDependencies: [Target.Dependency] = ["AwsCCommon"]
       name: "LibCrypto",
       pkgConfig: "libcrypto",
       providers: [
-        .brew(["openssl"]),
+        .brew(["openssl"])
       ]
     ))
   calDependencies.append(.target(name: "LibCrypto", condition: .when(platforms: [.macOS])))
--- a/Package.swift
+++ b/Package.swift
@@ -117,13 +117,14 @@ var awsCCalPlatformExcludes =
   packageTargets.append(
     .target(
       name: "S2N_TLS",
-      dependencies: ["LibCrypto"],
+      dependencies: [.target(name: "LibCrypto", condition: .when(platforms: [.macOS, .linux]))],
       path: "aws-common-runtime/s2n",
       exclude: s2nExcludes,
       publicHeadersPath: "api",
       cSettings: [
         .headerSearchPath("./"),
         .define("S2N_NO_PQ"),
+        .define("OPENSSL_SUPPRESS_DEPRECATED"),
         // This is a hack to get around the fact that S2N uses the compiler option `-include`
         // to include `s2n_prelude.h` in all .c files. Since SwiftPM doesn't support compiler flags,
         // we manually define the macros from `s2n_prelude.h`. When SwiftPM supports compiler flags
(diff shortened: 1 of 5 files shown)


Package.swift
```diff
@@ -116,13 +126,14 @@ var awsCCalPlatformExcludes =
   packageTargets.append(
     .target(
       name: "S2N_TLS",
-      dependencies: ["LibCrypto"],
+      dependencies: [.target(name: "LibCrypto", condition: .when(platforms: [.macOS, .linux]))],
```
xiazhvera: Would we need the platform condition? As this is already under `#if os(Linux) || os(macOS)`.

Package.swift
```diff
@@ -116,13 +126,14 @@ var awsCCalPlatformExcludes =
...
       path: "aws-common-runtime/s2n",
       exclude: s2nExcludes,
       publicHeadersPath: "api",
       cSettings: [
         .headerSearchPath("./"),
         .define("S2N_NO_PQ"),
+        .define("OPENSSL_SUPPRESS_DEPRECATED"),
```
xiazhvera, line 141: Why do we introduce `OPENSSL_SUPPRESS_DEPRECATED` here, but not in other PRs?

aws-common-runtime/aws-c-io
xiazhvera, line 1: It looks like aws-c-io is not point to the latest commit yet.

Test/AwsCommonRuntimeKitTests/XCBaseTestCase.swift
```diff
@@ -14,7 +14,7 @@ class XCBaseTestCase: XCTestCase {
     // XCode currently lacks a way to enable logs exclusively for failed tests only.
     // To prevent log spamming, we use `error` log level to only print error message.
     // We should update this once a more efficient log processing method becomes available.
-    try? Logger.initialize(target: .standardOutput, level: .error)
```
xiazhvera, line 17: Probably wanna keep use error. Using INFO would generate excessive log output.

builder.json
```diff
@@ -54,6 +54,12 @@
   "variants": {
     "localhost": {
       "!test_steps": ["localhost-test"]
+    },
+    "macos-s2n": {
```
xiazhvera, line 58: Xcode only pass in the environment variables with prefix `TEST_RUNNER_` for the test framework. For example, if we would like to use environment variable `FOO` in the unit test, we would need setup `TEST_RUNNER_FOO`. Then Xcode will pick up `TEST_RUNNER_FOO`, strip `TEST_RUNNER_`, and use `FOO` in it's unit tests. 
This does not affect applications, but if we are going have a README about this environment variable... this might be something worth a note.

xiazhvera (commented):
Got couple questions regards to the PR.

Package.swift
sfod: Cmake config for s2n-tls completely [disables deprecation warnings](https://github.com/aws/s2n-tls/blob/v1.7.3/CMakeLists.txt#L156-L157). But since crt-swift doesn't use cmake to build dependencies, deprecated functions cause compilation errors here. According to the comment right below this line, "SwiftPM doesn't support compiler flags", so we can't pass `-Wno-deprecated-declarations`.
I added a comment with explanation.

Package.swift
sfod: There is a difference between `#if os(macoS)` and `[.target(condition: .when(platforms: [.macOS]))]`. When you cross-compile on macOS for other platforms, the first is true and the second is false. I caught this issue on the iOS CI job where we build on macOS for iOS simulator.

sfod: Though, thinking more about it - libcrypto is always a dependency of s2n, so this condition looks out-of-place here. Reverted this change and added more comments where s2n and libcrypto are added as dependencies of CRT libs.

xiazhvera (approved):
LGTM, quick reminder that we need update aws-c-io before merge.

xiazhvera (changes requested):
The tls1.3 test seems to be skipped. What is plan for tls1.3 support then? Is it possible to adjust the Package.swift to make s2n build properly?

sfod:
> The tls1.3 test seems to be skipped. What is plan for tls1.3 support then? Is it possible to adjust the Package.swift to make s2n build properly?

It's possible but our previous attempts to use CMake for building dependencies failed. I suggest adding support for s2n-tls without exposing it to users. It'll allow other CRT repos to support s2n on macOS.

xiazhvera (approved):
Sounds good, let's have s2n in first. Let's remove the related doc though.

Issue #352, opened 2025-10-06
Potential race condition in `CommonRuntimeKit.initialize()`


jbelkins:
### Describe the bug

A customer recently reported this crash on AWS SDK for Swift, which occurred when they ran tests on Ubuntu / Swift 6.2 which utilize SDK components:
https://github.com/awslabs/aws-sdk-swift/issues/1984

When investigating with the customer, it was discovered that the cause of the crash reported by the customer was this error originating in aws-crt-swift:
```
AwsCommonRuntimeKit.CRTError(code: 1173, message: "Default TLS trust store not found on this system. 
Trusted CA certificates must be installed, or \"override default trust store\" must be used while creating the TLS 
context.", name: "AWS_IO_TLS_ERROR_DEFAULT_TRUST_STORE_NOT_FOUND")
```

This error is thrown when the SDK attempts to create a TLS context here:
https://github.com/smithy-lang/smithy-swift/blob/main/Sources/ClientRuntime/Networking/Http/CRT/SDKDefaultIO.swift#L74

We were able to eliminate the crash by calling `CommonRuntimeKit.initialize()` early in the test setup for the test that was crashing.

However, we already call `CommonRuntimeKit.initialize()` immediately above where the crash happens:
https://github.com/smithy-lang/smithy-swift/blob/main/Sources/ClientRuntime/Networking/Http/CRT/SDKDefaultIO.swift#L27

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

I would expect that once `CommonRuntimeKit.initialize()` has been called and has returned to the caller, that CRT and its components would be completely set up & ready for immediate use without any further delay or asynchronous completion.

### Current Behavior

The fact that calling `CommonRuntimeKit.initialize()` long before this crash eliminates it, but calling the same method immediately before it allows the crash, suggests that there is a race condition or some other problem with CRT initialization that results in returning to the caller before CRT setup is actually completed.

### Reproduction Steps

See https://github.com/awslabs/aws-sdk-swift/issues/1984 for reproduction of the customer's original issue

### Possible Solution

Not known

### aws-crt-swift version used

0.54.0

### Compiler and Version used

`swift:6.2-jammy` official Swift docker container

### Operating System and version

Ubuntu Jammy

sebsto:
Problem persist, please re-open
New report : https://github.com/awslabs/aws-sdk-swift/issues/2138

xiazhvera:
[...]

PR #389, opened 2026-05-26
Make CRT initialize thread safe

azkrishpy:
*Issue #, if available:*
#352 

*Description of changes:*
https://github.com/awslabs/aws-sdk-swift/issues/1984 & https://github.com/awslabs/aws-sdk-swift/issues/2138 talk about an issue that was filed with us but could not be reliably reproduced.

Creating a test that would reliably force a crash seems hard as this is a race, however running initialize from multiple threads and enabling ThreadSanitizer while testing allows us to see the race condition warning even if the test does not fail. Not adding the test since it might not be a necessary CI improvement. 

### DIFF
--- a/Source/AwsCommonRuntimeKit/CommonRuntimeKit.swift
+++ b/Source/AwsCommonRuntimeKit/CommonRuntimeKit.swift
@@ -19,7 +19,7 @@ public struct CommonRuntimeKit {
   // Tracks whether the C libraries have been initialized, so a second call to
   // `initialize()` is a no-op instead of re-running aws_*_library_init. Only
   // read/written while holding `lock`.
-  private static var isInitialized = false
+  private nonisolated(unsafe) static var isInitialized = false
 
   /// Initializes the library.
   /// Must be called before using any other functionality.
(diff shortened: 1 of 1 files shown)


azkrishpy:
The change follows the pattern observed in logger initialization:   https://github.com/awslabs/aws-crt-swift/blob/8290645b1b318c23c962b889a6d8b3d4523413f6/Source/AwsCommonRuntimeKit/crt/Logger.swift#L22-L57

Source/AwsCommonRuntimeKit/CommonRuntimeKit.swift
```diff
@@ -2,17 +2,22 @@ import AwsCAuth
...
 import LibNative
+import Foundation
 
 /// Initializes the library.
 /// `CommonRuntimeKit.initialize` must be called before using any other functionality.
 public struct CommonRuntimeKit {
 
   /// The current version of the AWS Common Runtime Kit.
   public static let CRTVersion = "0.0.0"
+  private static let lock = NSLock()
```
rakshil14-2, line 18: Could we have a short comment explaining why the lock exists?

azkrishpy: added a comment

rakshil14-2 (approved):
LGTM

xiazhvera (changes requested):
We should also prevent  multi-init/multi-cleanup of aws-c-* along side with the lock.

xiazhvera (approved):

xiazhvera:
Swift 6 is strict on concurrency safety, while the `isInitialized` was actually protected by the lock. Refer to https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/commonproblems/ Fix&ship.

jbelkins (approved):

PR #395, opened 2026-06-12
Add certificateSource for TLS Context

xiazhvera:
*Issue #, if available:*

*Description of changes:*
- Add certificateSource for TLS Context

### DIFF
--- a/Source/AwsCommonRuntimeKit/io/TLSContextOptions.swift
+++ b/Source/AwsCommonRuntimeKit/io/TLSContextOptions.swift
@@ -11,6 +11,9 @@ public class TLSContextOptions: CStruct {
   /// track minimum tls version set by user
   internal var minimumTLSVersion: TLSVersion?
 
+  /// The certificate source automatically determined from the creation method. Used for metrics tracking.
+  internal var certificateSource: CertificateSource?
+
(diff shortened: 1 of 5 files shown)


Source/AwsCommonRuntimeKit/io/TLSContextOptions.swift
```diff
@@ -219,3 +226,21 @@ public enum TLSVersion: UInt32 {
...
+/// The source of the TLS certificate used for authentication.
+/// This value is tracked in IoT SDK metrics as feature ID "I".
+/// This type is package-private and not accessible to external libraries.
+enum CertificateSource {
+  /// Certificate and private key loaded from PEM files
+  case certificateFiles
+  /// Certificate managed via PKCS#11 hardware security module
+  case pkcs11
+  /// Certificate stored in the Windows certificate store
+  case windowsCertStore
```
rakshil14-2: **Nit**:  Here  `.pkcs11`, `.windowsCertStore`, and `.javaKeystore` are defined, but no code path in this Swift library ever sets them. Should these be added later when they're actually reachable, rather than now?

rakshil14-2 (approved):

PR #403, opened 2026-07-28
Add socket option tcp nodelay

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Source/AwsCommonRuntimeKit/io/TCPNoDelay.swift
+++ b/Source/AwsCommonRuntimeKit/io/TCPNoDelay.swift
@@ -1,0 +1,25 @@
+//  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//  SPDX-License-Identifier: Apache-2.0.
+
+import AwsCIo
+
+/// Controls the TCP_NODELAY socket option (whether Nagle's algorithm is disabled).
+/// TCP only. Ignored for UDP and AWS_SOCKET_LOCAL sockets.
+public enum TCPNoDelay {
+  /// Leave the OS default in place (Nagle's algorithm enabled on most platforms).
+  case osDefault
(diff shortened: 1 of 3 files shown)


Source/AwsCommonRuntimeKit/io/TCPNoDelay.swift
azkrishpy, line 1: Would we be better off keeping socket options related things within socket options itself? I see that SocketType does something similar so I am not feeling strongly about this. 
We might want to add new options and each extension getting a new file seems overkill.

Test/AwsCommonRuntimeKitTests/io/SocketOptionsTest.swift
```diff
@@ -37,5 +37,8 @@ class SocketOptionsTests: XCBaseTestCase {
 
     socketOptions.socketType = SocketType.stream
     XCTAssertEqual(socketOptions.socketType, SocketType.stream)
+
+    socketOptions.tcpNoDelay = TCPNoDelay.off
```
azkrishpy, line 41: Unrelated, but this test does not seem particularly useful.. `var` itself is a mutable variable in swift so it would get modified.

azkrishpy (approved):
some minor questions

Source/AwsCommonRuntimeKit/io/TCPNoDelay.swift
DmitriyMusatkin: both SocketType and SocketDomain used separate file style, so my vote is to continue doing that until we have strong reason to refactor all of those cases

Test/AwsCommonRuntimeKitTests/io/SocketOptionsTest.swift
DmitriyMusatkin: yeah, this test is truthfully not testing much. i considered dropping it, but dont think there is currently a strong need for that

